PR #2, opened 2015-12-29
Adding hg support

MatthewTurk:
This adds the ability to prefix a URL with `hg+` and use `python-hglib` to interact with it.


scopatz:
Woohoo! Thanks. This was definitely needed.

PR #6, opened 2016-03-17
Add sorting by first letter of last name

MatthewTurk:
I *think* this is a not-likely-to-fail way of sorting by the first letter of the last name in addition to `position`


scopatz:
For sure, like totally.

PR #8, opened 2017-09-19
Rename README.rst to README.md

CJ-Wright:


scopatz:
Thanks for the sentiment here, @CJ-Wright - but it is probably worth changing the document format to be in rST.  For this PR to be valid, the ``setup.py`` file would need to change to convert md to rsST, which no one wants :)

PR #43, opened 2017-11-10
Former

CJ-Wright:
Add space/key for former members (former and `active: false`, respectively)

### DIFF
--- a/regolith/templates/former.html
+++ b/regolith/templates/former.html
@@ -3,7 +3,7 @@
 <h1>Former Members</h1>
 {% endblock header %}
 {% block content %}
-{% for p in people if p.get('active', False) %}
+{% for p in people if not p.get('active', True) %}
   <a href="{{root}}/people/{{p['_id']}}.html" class="image-link">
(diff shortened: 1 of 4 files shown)


regolith/templates/former.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base.html" %}
+{% block header %}
+<h1>Former Members</h1>
+{% endblock header %}
+{% block content %}
+{% for p in people if p.get('active', False) %}
```
scopatz: Shouldn't this be `if not p.get('active', True)`?  That former members are those whose active setting is False, and the default active setting is True?

scopatz:
Thanks @CJ-Wright! I think the people page also needs the active filter so that only current members are selected.

scopatz:
Scratch that last comment, just saw the other PR

CJ-Wright:
@scopatz is that level of docs ok or do you want more?

scopatz:
Yeah that is great!

PR #46, opened 2017-11-15
ENH: add conda forge rever activity

CJ-Wright:
@scopatz ready for review


rever.xsh
```diff
@@ -1,10 +1,14 @@
+$PROJECT = 'regolith'
 $REVER_DIR = 'rever-tmp'
```
scopatz: I think that this line can / should be removed.  we can use the default rever dir here

scopatz:
Thanks for putting this in @CJ-Wright!  Also, can you please add the rever dir (which should be `rever/`) to the gitignore file? Thanks!

scopatz:
Thanks @CJ-Wright!

PR #66, opened 2017-12-12
WIP: schemas

CJ-Wright:


CJ-Wright:
Try this instead http://docs.python-cerberus.org/en/stable/index.html

regolith/tests/test_validators.py
```diff
@@ -0,0 +1,11 @@
+from regolith.validators import validate_schema
+from regolith.schemas import schemas
+from regolith.exemplars import exemplars
+import pytest
```
scopatz: PEP8 Third party imports go before package imports

requirements/run.txt
```diff
@@ -1 +1,3 @@
-ruamel.yaml
\ No newline at end of file
+python
```
scopatz: Is this really a valid requirement?

CJ-Wright: I took inspiration for this from conda forge recipes.

scopatz: That is crazy if it is.

regolith/validators.py
```diff
@@ -109,4 +113,49 @@ def ensure_email(email):
...
+    ----------
+    record : dict
+        The database record to be tracked
+    schema : dict
+        The schema to validate the record against
+    keys : tuple
+        The sequential keys for the data being accessed, used for
+        printing/debuging errors
+    """
+    if isinstance(record, dict):
```
scopatz: check `collections.abc.Mapping`, not dict.  Same goes for below.

requirements/run.txt
CJ-Wright: We usually include python as a run requirement in conda recipes, right?

regolith/validators.py
```diff
@@ -109,4 +113,49 @@ def ensure_email(email):
...
+        total_keys.update(set(record.keys()))
+
+        for k in total_keys:
+            if k not in schema:
+                pass
+            if k not in record and schema[k].get('required', False):
+                raise ValueError('{} is required in {}'.format(k, keys))
+            elif k in record and k in schema:
+                validate_schema(record[k], schema[k], keys + (k, ))
+    elif isinstance(record, collections.Iterable) and not isinstance(record,
```
scopatz: Does this really include set and generators? Or is this supposed to be `collections.abc.Sequence`?

regolith/validators.py
```diff
@@ -109,4 +113,49 @@ def ensure_email(email):
...
+                                                                     str):
+        if isinstance(record[0], dict):
+            for r in record:
+                validate_schema(r, schema['type'], keys)
+        else:
+            for r in record:
+                validate_schema(r, schema, keys)
+    else:
+        if not isinstance(record, schema['type']):
```
scopatz: This should probably be an elif, rather than an if inside of an else.

requirements/run.txt
scopatz: yes, that is true. I wonder about whether this affects pip, though maybe we don't care :man_shrugging:

regolith/tests/test_validators.py
```diff
@@ -0,0 +1,11 @@
+from regolith.validators import validate_schema
+from regolith.schemas import schemas
+from regolith.exemplars import exemplars
```
scopatz: If the exemplars are only used in the tests, then they should be in the test suite, not the package.

regolith/schemas.py
```diff
@@ -0,0 +1,400 @@
+"""Database schemas"""
+schemas = {
```
scopatz: PEP8 global constants should be in `UPPERCASE`

regolith/validators.py
```diff
@@ -109,4 +113,49 @@ def ensure_email(email):
...
+def validate_schema(record, schema, keys=()):
+    """Validate a database record against a schema
+
+    Parameters
+    ----------
+    record : dict
+        The database record to be tracked
+    schema : dict
+        The schema to validate the record against
+    keys : tuple
```
scopatz: Should be listed as `, optional`

regolith/validators.py
```diff
@@ -109,4 +113,49 @@ def ensure_email(email):
...
+        total_keys = set(schema.keys())
+        remove_keys = ['required', 'type', 'description']
+        for k in remove_keys:
+            if k in total_keys:
+                total_keys.remove(k)
+        total_keys.update(set(record.keys()))
+
+        for k in total_keys:
+            if k not in schema:
+                pass
```
scopatz: Isn't it an error if we get a key that we don't expect?

regolith/validators.py
```diff
@@ -109,4 +113,49 @@ def ensure_email(email):
...
+
+    Parameters
+    ----------
+    record : dict
+        The database record to be tracked
+    schema : dict
+        The schema to validate the record against
+    keys : tuple
+        The sequential keys for the data being accessed, used for
+        printing/debuging errors
```
scopatz: Broadly speaking, I guess I'd prefer if this returned (bool, str message) rather than raising an error.  That way the call site can decide whether or not to raise the error. Also, it allows the message to be appended to so that all of the ways that the record doesn't adhere to the schema can be known with one call to this function.

scopatz:
This is a really great first cut @CJ-Wright! Other than the parts listed inline, I think the thing that makes sense (so that info isn't duplicated) is to have the docs in conf.py generate the schema section of the collections pages from the schema itself. I think that this is needed so there aren't two, possibly diverging sources of truth.

scopatz:
Also, cerebus is interesting, but it seems to not be on conda-forge.

CJ-Wright:
https://github.com/conda-forge/cerberus-feedstock

scopatz:
Ahh, mistyped it!

CJ-Wright:
Sorry, I just pushed up a cerberus driven version

regolith/validators.py
CJ-Wright: I think this is done with cerberus. (validator.errors)

regolith/validators.py
```diff
@@ -1,10 +1,15 @@
-"""Validators and convertors for regolith input."""
+"""Validators and converters for regolith input."""
+import collections
 import os
 import re
+
+from cerberus import Validator
 from getpass import getpass
```
scopatz: PEP8 cerberus is a 3rd party library and so should go below getpass, which is from the standard library and should be grouped with those above.

regolith/validators.py
```diff
@@ -109,4 +116,35 @@ def ensure_email(email):
...
-    }
+}
+
+
+def pop_description(schema):
+    """Pop description off the schema since Cerberus doesn't support it"""
+    if isinstance(schema, dict):
+        schema.pop('description', None)
+        for v in schema.values():
+            pop_description(v)
```
scopatz: You shouldn't need to do this. Instead subclass validtor and make it skip custom keys in the schema we want to add.

scopatz: See http://docs.python-cerberus.org/en/stable/customize.html#custom-validators

regolith/tests/test_validators.py
```diff
@@ -0,0 +1,11 @@
+from regolith.validators import validate_schema
+from regolith.schemas import SCHEMAS
+from regolith.exemplars import EXEMPLARS
+import pytest
```
scopatz: PEP8 import order

regolith/validators.py
CJ-Wright: I spent some time trying to do that without much headway, but I will try again.

regolith/validators.py
```diff
@@ -109,4 +113,35 @@ def ensure_email(email):
...
+    Returns
+    -------
+    bool:
+        True is valid
+
+    """
+    from regolith.schemas import SCHEMAS
+    schema = SCHEMAS[db]
+    pop_description(schema)
+    v = Validator(schema, allow_unknown=True)
```
scopatz: I don't think we want this, in general. Are you doing this only for the citations entry?  Maybe there is a way to add allow unknow for only certain schema.

regolith/validators.py
```diff
@@ -109,4 +113,35 @@ def ensure_email(email):
...
+    Parameters
+    ----------
+    db : str
+        The name of the db in question
+    record : dict
+        The record to be validated
+
+    Returns
+    -------
+    bool:
```
scopatz: I think this should be listed as `rtn : bool`

regolith/validators.py
CJ-Wright: We can put it into each schema level as needed. http://docs.python-cerberus.org/en/stable/usage.html#allowing-the-unknown

regolith/validators.py
```diff
@@ -109,4 +113,35 @@ def ensure_email(email):
...
+        True is valid
+
+    """
+    from regolith.schemas import SCHEMAS
```
scopatz: There is no circular importing here. Please move this import to the top of the file.

scopatz: Or maybe, better yet, move this function to schemas.py or SCHEMAs to this module.  Modules with only one object don't deserve to exist :)

regolith/validators.py
scopatz: There has to be a way.  Popping stuff out of a global, mutable object is wrong.

scopatz: In this case at least. The alternative is to have a second dict which has the same structure and only contains the descriptions, but that feels wrong too.

regolith/validators.py
scopatz: Great! Do that then :)

regolith/validators.py
CJ-Wright: Btw there is an open issue for `description` https://github.com/pyeve/cerberus/issues/254

CJ-Wright: And now a SO question https://stackoverflow.com/questions/47851546/cerberus-additional-schema-keys

docs/conf.py
```diff
@@ -253,3 +260,85 @@
...
+    'courses': 'This is a collection that describes a course, when it happened, and\nwho is taking it. This is likely private.\n\n',
+    'grades': 'The grade for a student on an assignment. This information should be private.\n\n',
+    'grants': 'This collection represents grants that have been awarded to the group.\n\n',
+    'jobs': 'This collection describes the research group jobs. This is normally public data.\n\n\n',
+    'news': 'This collection describes the research group news. This is normally public data.\n\n',
+    'people': 'This collection describes the members of the research group.  This is normally public\ndata.\n\n',
+    'projects': 'This collection describes the research group projects. This is normally public data.\n\n',
+    'proposals': 'This collection represents proposals that have been submitted by the group.\n\n',
+    'students': 'This is a collection of student names and metadata. This should probably be private.\n\n'}
```
scopatz: These seem like these could / should also go into the SCHEMAS data structure.

docs/conf.py
```diff
@@ -253,3 +260,85 @@
 
 # How to display URL addresses: 'footnote', 'no', or 'inline'.
 # texinfo_show_urls = 'footnote'
+
+schema_top_docs = {
+    'abstracts': 'Abstracts for a conference or workshop. This is generally public information\n\n',
+    'assignments': 'Information about assignments for classes.\n\n',
+    'blank': '\n',
```
scopatz: not needed since it isn't a real schema and it shouldn't be generated.

scopatz: you actually can probably just remove the blank.rst file since it is no longer needed.

docs/conf.py
```diff
@@ -253,3 +260,85 @@
...
+    'grants': 'This collection represents grants that have been awarded to the group.\n\n',
+    'jobs': 'This collection describes the research group jobs. This is normally public data.\n\n\n',
+    'news': 'This collection describes the research group news. This is normally public data.\n\n',
+    'people': 'This collection describes the members of the research group.  This is normally public\ndata.\n\n',
+    'projects': 'This collection describes the research group projects. This is normally public data.\n\n',
+    'proposals': 'This collection represents proposals that have been submitted by the group.\n\n',
+    'students': 'This is a collection of student names and metadata. This should probably be private.\n\n'}
+
+from pprint import pprint
+import string
```
scopatz: PEP8 imports should go at the top of the file.

docs/conf.py
```diff
@@ -253,3 +260,85 @@
...
+    'grades': 'The grade for a student on an assignment. This information should be private.\n\n',
+    'grants': 'This collection represents grants that have been awarded to the group.\n\n',
+    'jobs': 'This collection describes the research group jobs. This is normally public data.\n\n\n',
+    'news': 'This collection describes the research group news. This is normally public data.\n\n',
+    'people': 'This collection describes the members of the research group.  This is normally public\ndata.\n\n',
+    'projects': 'This collection describes the research group projects. This is normally public data.\n\n',
+    'proposals': 'This collection represents proposals that have been submitted by the group.\n\n',
+    'students': 'This is a collection of student names and metadata. This should probably be private.\n\n'}
+
+from pprint import pprint
```
scopatz: also it might be nice if you did, 

```python
try:
    from pprintpp import pprint
except ImportError:
    from pprint import pprint
```

and then added pprintpp as a dep, since it's output is so much nicer.

regolith/schemas.py
```diff
@@ -0,0 +1,762 @@
...
+                              'year': {'required': True, 'type': 'integer'}}
+               },
+               'title': {'description': 'for example, Dr., etc.',
+                         'required': False,
+                         'type': 'string'}
+               }
+}
+
+
+class MyValidator(Validator):
```
scopatz: MyValidator is not a good class name :)

regolith/schemas.py
```diff
@@ -0,0 +1,762 @@
...
+    def _validate_description(self, description, field, value):
+        if False:
+            self._error(field, "Shouldn't be here")
```
scopatz: Can't this function just be a `pass`?

regolith/schemas.py
```diff
@@ -0,0 +1,762 @@
...
+    errors: dict
+        The errors encountered (if any)
+
+    """
+    schema = SCHEMAS[db]
+    v = MyValidator(schema)
+    return v.validate(record), v.errors
+
+
+EXEMPLARS = {'abstracts': {'_id': 'Mouginot.Model',
```
scopatz: PEP8 Constants should be at the top of the file.

docs/conf.py
CJ-Wright: Maybe, but that might be rough, since we don't have any special keys at that level.

scopatz: Maybe add an `_description` or `_description_` or `__description__` key that is also skipped?

regolith/fsclient.py
```diff
@@ -48,10 +49,14 @@ def dump_yaml(filename, docs, inst=None):
     """Dumps a dict of documents into a file."""
     inst = YAML() if inst is None else inst
     inst.indent(mapping=2, sequence=4, offset=2)
-    for doc in docs.values():
+    my_sorted_dict = ruamel.yaml.comments.CommentedMap()
```
scopatz: please remove the `my_` from this variable.

docs/collections/index.rst
```diff
@@ -8,16 +8,16 @@ The following contain the regolith schemas and examples in both YAML and JSON/Mo
...
     jobs
     news
-    people
-    projects
-    proposals
-    students
+    auto/people
+    auto/projects
+    auto/proposals
+    auto/students
```
scopatz: I'd really prefer these not go into an auto folder. It doesn't seems necessary.

docs/collections/index.rst
```diff
@@ -8,16 +8,16 @@ The following contain the regolith schemas and examples in both YAML and JSON/Mo
...
+    auto/assignments
+    auto/blog
     citations
     courses
-    grades
-    grants
+    auto/grades
+    auto/grants
     jobs
     news
```
scopatz: Why are these not generated?

regolith/schemas.py
```diff
@@ -0,0 +1,662 @@
...
+                               'year': 1998},
+                           'status': 'submitted',
+                           'title': 'A very fine proposal indeed',
+                           'year': 1999},
+             'students': {'_id': 'Human A. Person',
+                          'aka': ['H. A. Person'],
+                          'email': 'haperson@uni.edu',
+                          'university_id': 'HAP42'}}
+
+SCHEMAS = {'abstracts': {'_description': {'description': 'Abstracts for a conference or workshop. This is generally public information\n\n'},
```
scopatz: many lines too long here, also probably shouldn't have `\n\n` at the end of them. That can be added in the doc generator

docs/collections/index.rst
CJ-Wright: I haven't made schemas for them yet

scopatz: haha OK - should that be a requirement for this PR? Or do you want to delay it?

CJ-Wright: I was hoping to delay it. I'm expecting a couple more takes at this as I start using more of the dbs so I might come around to it.

scopatz: Sounds good to me!

regolith/app.py
```diff
@@ -6,6 +6,8 @@
 
 from flask import Flask, abort, request, render_template, redirect, url_for
 
+from regolith.schemas import validate
+
```
scopatz: PEP8 two newlines required after import lines

regolith/app.py
```diff
@@ -59,6 +61,17 @@ def collection_page(dbname, collname):
                     f.write(form['body'])
                 traceback.print_exc()
                 raise
+            tv, errors = validate(dbname, body)
+            if not tv:
+                td = tempfile.TemporaryDirectory()
+                n = os.path.join(td.name, 'regolith.txt')
+                print(errors)
+                print('Writing text file to {}. '
+                      'Please try again.'.format(n))
```
scopatz: I think these prints should go into the body of the error message below. At the very least, they should be printed to stderr....

regolith/app.py
```diff
@@ -59,6 +61,17 @@ def collection_page(dbname, collname):
...
+                with open(n) as f:
```
scopatz: This file isn't opened in write mode. Do you have any tests for the invalid case?

regolith/app.py
```diff
@@ -75,6 +88,16 @@ def collection_page(dbname, collname):
...
+                    f.write(form['body'])
```
scopatz: Same as above

regolith/tests/test_validators.py
```diff
@@ -0,0 +1,8 @@
...
```
scopatz: Can you please move the tests files to a top-level tests/ dir, rather than embedding them in the project dir? Thanks!

scopatz:
Thanks! Some more minor comments.

regolith/app.py
CJ-Wright: I don't think we have any tests for the app.

scopatz: Fair point!

CJ-Wright: Made an issue https://github.com/regro/regolith/issues/69

CJ-Wright:
@scopatz ready for another round!

scopatz:
Looks good to go! Can you please provide a news entry?

news/schema.rst
```diff
@@ -0,0 +1,16 @@
+**Added:**
+
+* Schemas for the tables
```
scopatz: This is not particularly informative.  Can you please expand on it. Also, aren't we calling them collections, not tables?

news/schema.rst
```diff
@@ -0,0 +1,22 @@
+**Added:**
+
+* Schemas and exemplars for the collections.
+Database entries are checked against the schema, making sure that all the
+required fields are filled and the values are the same type(s) listed in the
+schema. The schema also includes descriptions of the data to be included.
+The exemplars are examples which have all the specified fields and are
+used to check the validation.
```
scopatz: This is not valid rst, which requires the paragraph to be indented after the bullet point.

scopatz: I'll merge this and fix it though.

Issue #56, opened 2017-11-22
ENH: Match public db entries with private


CJ-Wright:
When running interactively (or with builders which use both the public and private dbs) it may be nice to be able to combine similarly named dbs with the same `_id`. Eg a public project entry may have some information but the private entry keeps track of accounting, ongoing work, etc. but when we go to build an internal report we want both dbs to be accessed together.

scopatz:
Good idea!

PR #70, opened 2018-01-03
WIP: mega_db the agglomeration of all the databases in the rc

CJ-Wright:


CJ-Wright:
@scopatz do you want me to change `all_documents` to pull from the mega_db?

regolith/fsclient.py
```diff
@@ -93,23 +95,42 @@ def load_json(self, db, dbpath):
...
+                        self.mega_db[base][k].update(v)
+                    else:
+                        self.mega_db[base][k] = v
+            else:
+                self.mega_db[base] = coll
 
     def load_database(self, db):
         """Loads a database."""
         dbpath = dbpathname(db, self.rc)
+        print(dbpath)
```
scopatz: I don't think this is needed.

regolith/fsclient.py
```diff
@@ -93,23 +95,42 @@ def load_json(self, db, dbpath):
...
+                    else:
+                        self.mega_db[base][k] = v
+            else:
+                self.mega_db[base] = coll
 
     def load_yaml(self, db, dbpath):
         """Loads the YAML part of a database."""
         dbs = self.dbs
-        for f in iglob(os.path.join(dbpath, '*.y?ml')):
+        for f in iglob(os.path.join(dbpath, '*.y*ml')):
```
scopatz: Shouldn't this be a `?`. The only valid option is an `a`, which may or may not be present.  We shouldn't be matching thins like yajsjhsml

regolith/fsclient.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, rc):
         self.rc = rc
         self.closed = True
         self.dbs = None
+        self.mega_db = None
```
scopatz: Mongo client would also need to be changed...

scopatz:
Hi @CJ-Wright - thanks for putting this in! First there is no news entry. Second, I am not sure I really understand the abstraction here.  What exactly are you going for and is there a more illuminating name than "mega_db"?

regolith/fsclient.py
CJ-Wright: The `?` failed on my machine with `.yml` files
"A '?' (not between brackets) matches any single character." So it won't match an empty character.

CJ-Wright:
The idea is to close #56 by creating a group of collections which merge all the documents across the databases. Eg the publicly facing people DB may contain some of the people information and the private has the rest (their funding, private notest, etc.) but if you have addresses/access to both it may be nice to have them both loaded into the same dict.

scopatz:
I think that you want to use ChainMap for this: https://docs.python.org/3/library/collections.html#collections.ChainMap

regolith/fsclient.py
```diff
@@ -93,19 +95,39 @@ def load_json(self, db, dbpath):
...
             self._collfiletypes[base] = 'json'
             print('loading ' + f + '...', file=sys.stderr)
-            dbs[db['name']][base] = load_json(f)
+            coll = load_json(f)
+            dbs[db['name']][base] = coll
+            if base in self.chained_db:
+                for k, v in coll.items():
+                    if k in self.chained_db[base]:
+                        self.chained_db[base][k] = (self.chained_db[base][k].
+                            new_child(v))
```
scopatz: Is this correct? I think that the maps should be added in the same order that the databases are added.  As such, it is probably not best to add them in the loader function for a single database, but rather construct the chained_db after *all* databases are loaded.  Also, here, it seems that new_child() prepends the map: https://docs.python.org/3/library/collections.html#collections.ChainMap.new_child, we probably just want to append to the maps list.

regolith/mongoclient.py
```diff
@@ -5,6 +5,10 @@
 import shutil
 import subprocess
 from glob import iglob
+from warnings import warn
+
+warn("Mongo support will be deprecated in the near future please use "
+     "fsclient.FileSystemClient")
```
scopatz: I think the warning should only be presented if the client is instantiated, not on import...

regolith/fsclient.py
```diff
@@ -98,18 +100,20 @@ def load_json(self, db, dbpath):
             base, ext = os.path.splitext(collfilename)
             self._collfiletypes[base] = 'json'
             print('loading ' + f + '...', file=sys.stderr)
-            dbs[db['name']][base] = load_json(f)
+            coll = load_json(f)
+            dbs[db['name']][base] = coll
```
scopatz: no reason to break this up onto two lines, non-blocking.

regolith/fsclient.py
```diff
@@ -3,7 +3,7 @@
 import sys
 import json
 from glob import iglob
-from collections import defaultdict
+from collections import defaultdict, ChainMap
```
scopatz: ChainMap is no longer used here.

regolith/database.py
```diff
@@ -5,6 +5,7 @@
 import subprocess
 from glob import iglob
 from warnings import warn
+from collections import ChainMap, defaultdict
```
scopatz: defaultdict isn't used here.

regolith/database.py
```diff
@@ -119,8 +120,19 @@ def dump_database(db, client, rc):
...
     for db in rc.databases:
         load_database(db, client, rc)
+        for base, coll in client.dbs[db['name']].items():
+            if base not in chained_db:
+                chained_db[base] = {}
+            for k, v in coll.items():
+                if k in chained_db[base]:
+                    chained_db[base][k].maps.append(v)
+                else:
+                    chained_db[base][k] = ChainMap(v)
```
scopatz: This is a good first cut, but it is not fully clear what happens on insertions. That is, 

1. how are insertions of new collections and documents in a database reflected in the chained_db?
2. how are insertions of new collections and documents into the chained db inserted into a concrete db?

CJ-Wright:
Ran into an interesting issue in: 
`{% set docdump = json.dumps(doc, sort_keys=True, indent=1) %}`
`TypeError: Object of type 'ChainMap' is not JSON serializable`

CJ-Wright:
I may need to revert the changes to `all_documents`.

scopatz:
Well you probably just need to add a simple `default()`  and/or `object_hook()` function that is able to de/persist  ChainMaps.  See `fixie.jsonutils` for an example for some other data types: https://github.com/ergs/fixie/blob/master/fixie/jsonutils.py

regolith/chained_db.py
```diff
@@ -0,0 +1,17 @@
+"""Base class for chaining DBs"""
+
+from collections import ChainMap
+
+
+class ChainDB(ChainMap):
+    """ A ChainMap who's ``_getitem__`` returns an instance of itself"""
```
scopatz: Since getitem does not return `self` (or a copy of self) it is not fair to say that it returns itstance of itself.  Rather, it returns an instance of ChainDB.  Also should it return an instance of ChainMap?

Also, should there be a setitem method so that incoming mappings are converted to chainmaps?

CJ-Wright:
@scopatz ready for more review!

regolith/chained_db.py
```diff
@@ -0,0 +1,27 @@
...
+
+class ChainDB(ChainMap):
+    """ A ChainMap who's ``_getitem__`` returns either a ChainDB or
+    the result"""
+
+    def __getitem__(self, key):
+        res = None
+        results = []
+        # Try to get all the data from all the mappings
+        for i, mapping in enumerate(self.maps):
```
CJ-Wright: remove enum.

regolith/chained_db.py
```diff
@@ -0,0 +1,27 @@
...
+            results.append(mapping.get(key, None))
+        # if all the results are mapping create a ChainDB
+        if all([isinstance(result, MutableMapping) for result in results]):
+            for result in results:
+                if res is None:
```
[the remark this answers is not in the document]
scopatz: Oh never mind, it is

regolith/templates/collection.html
```diff
@@ -27,7 +27,7 @@ <h1>Collection {{dbname}}/{{collname}}</h1>
   <button name="shutdown" value="shutdown">Shutdown Server</button>
 </form>
 
-{% for doc in rc.client.all_documents(dbname, collname) %}
+{% for doc in rc.client.all_documents( collname) %}
```
scopatz: weird space before collname

tests/test_chain_db.py
```diff
@@ -0,0 +1,51 @@
+from regolith.chained_db import ChainDB
+
+
+def test_dddi():
+    a = {'a': {'a': {'a': 1}}}
+    b = ChainDB(a)
+    print(b['a'])
+    print(b['a']['a'])
+    print(b['a']['a']['a'])
```
scopatz: It would be better to have asserts here. Even assert isinstance()

tests/test_chain_db.py
```diff
@@ -0,0 +1,126 @@
...
+    m2 = {'a': {'m': e}}
+    z = ChainDB(m1)
+    z.maps.append(m2)
+    g = [-1, -2]
+    z['a']['mm'] = g
+    assert isinstance(z['a'], ChainDB)
+    assert isinstance(z['a']['m'], list)
+    assert isinstance(z['a'].maps[0], dict)
+    assert g is z['a'].maps[0]['mm']
+    # We sent this to the first map not the last
```
scopatz: Wait, why does access default to the last, but setting defaults to the first?

CJ-Wright: This is from chainmap, if the key doesn't exist it then defaults into the first mapping.

scopatz: This comment still doesn't make sense to me.  `z['a'].maps[0]['mm']` should be `g` and  `z['a']['m']` should be a chainmap that represents g, right? its isn't an order in the maps list test.

CJ-Wright: Ah sorry I may have copied two lines too many

regolith/templates/grader.html
```diff
@@ -10,15 +10,15 @@
...
+  {% for course in sorted(rc.client.all_documents( 'courses'), key=by_id) %}
   {% if course['active'] %}
     {% set course_id = course['_id'] %}
     <h4>Course: {{course_id}}</h4>
     <div class="course_{{course_id}}"></div>
     <script type="text/javascript">
     var courseColumns = [
         {name: "student", type: "string"},
-        {%- for assign in sorted(rc.client.all_documents(dbname, 'assignments'), key=by_id) -%}
+        {%- for assign in sorted(rc.client.all_documents( 'assignments'), key=by_id) -%}
```
scopatz: PEP8 odd space before str

regolith/templates/grader.html
```diff
@@ -32,7 +32,7 @@ <h4>Course: {{course_id}}</h4>
     var courseData = [
         {% for student in course['students'] %}
         {student: "{{student}}",
-          {%- for grade in rc.client.all_documents(dbname, 'grades') -%}
+          {%- for grade in rc.client.all_documents( 'grades') -%}
```
scopatz: PEP8 odd space before str

requirements/run.txt
```diff
@@ -1,4 +1,5 @@
 python
 ruamel.yaml
+docutils
```
scopatz: Wait, really? Is this truly a run requirement?

tests/test_chain_db.py
```diff
@@ -0,0 +1,141 @@
...
+    z = ChainDB(m1)
+    z.maps.append(m2)
+    g = 'c'
+    z['a']['m'] = g
+    assert isinstance(z['a'], ChainDB)
+    assert isinstance(z['a']['m'], str)
+    assert isinstance(z['a'].maps[0], dict)
+    assert g is z['a'].maps[1]['m']
+    # We sent this to the first map not the last
+    assert z['a']['m'] is g
```
scopatz: This is not a great comparison for a sort of subtle reason.  Namely, CPython will intern (cache) string. So two string objects (normally literals) that are independently created may, in fact, be the same object. Here for example, 

```python
>>> 'c' is 'c'
True
```

It is better to do this kind of test with some kind of container type.  Even tuples work, because

```python
>>> ('c',) is ('c',)
False
```

tests/test_chain_db.py
```diff
@@ -0,0 +1,141 @@
...
+    z = ChainDB(m1)
+    z.maps.append(m2)
+    g = [-1, -2]
+    z['a']['mm'] = g
+    assert isinstance(z['a'], ChainDB)
+    assert isinstance(z['a']['m'], list)
+    assert isinstance(z['a'].maps[0], dict)
+    assert g is z['a'].maps[1]['mm']
+    # We sent this to the first map not the last
+    assert z['a']['m'] is not g
```
scopatz: Was this supposed to be `z['a']['mm']`? I am not sure that I understand the comment. 

Also did g go to m2 because m2 already had an `'mm'` key or because m2 is the last entry in the maps list?

scopatz: from the following test it looks like it is because it m2 had 'mm'

tests/test_chain_db.py
```diff
@@ -0,0 +1,141 @@
...
+    c = [{'m': 1}, {'n': 2}]
+    d = [{'o': 3}, {'p': 4}]
+    t = c + d
+    m1 = {'a': {'b': c}}
+    m2 = {'a': {'b': d}}
+    z = ChainDB(m1)
+    z.maps.append(m2)
+    assert isinstance(z['a'], ChainDB)
+    assert isinstance(z['a']['b'], list)
+    assert z['a']['b'] is not t
```
scopatz: Shouldn't this be an equality test ('==').  `t = c + d` will always be a new list instance, and so `is not` is implied.  It would be nice to have tests that check list mutability with append() and/or extend()

requirements/run.txt
CJ-Wright: https://github.com/regro/regolith/blob/cb94f4a8c5a5401aeaec655f5125354ada46c786/regolith/emailer.py

tests/test_chain_db.py
CJ-Wright: Correct, we first look through all the elements in the `ChainedDB` to see if any of the keys match. If there is a match we overwrite else we send it to the first map.

CJ-Wright:
@scopatz ready for more review!

requirements/run.txt
scopatz: Fair :)

regolith/templates/grader.html
```diff
@@ -10,15 +10,15 @@
 <h1>Welcome to the Regolith Grader!</h1>
 {% for dbname in rc.client.keys() if dbname != 'local' %}
   <h3>{{ dbname }}{% if status %} [{{status}}]{% endif %}</h3>
-  {% for course in sorted(rc.client.all_documents(dbname, 'courses'), key=by_id) %}
+  {% for course in sorted(rc.client.all_documents( 'courses'), key=by_id) %}
```
scopatz: PEP8 no space before courses

CJ-Wright: Damn it I thought I caught all those

scopatz:
Can you add an autodoc hook in `docs/api` for regolith/chained_db.py, please?

scopatz:
After that I think it is gtg!

CJ-Wright:
There seems to be an error in the fsclient line 44 when building the docs.

CJ-Wright:
```
/home/christopher/dev/regolith/docs/api/interactive.rst:16: WARNING: autodoc: failed to import module 'regolith.interact'; the following exception was raised:                                         
Traceback (most recent call last):
  File "/home/christopher/mc/envs/regro/lib/python3.6/site-packages/sphinx/ext/autodoc.py", line 658, in import_object
    __import__(self.modname)
  File "/home/christopher/dev/regolith/regolith/interact.py", line 11, in <module>
    with connect(rc) as rc.client:
  File "/home/christopher/mc/envs/regro/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/home/christopher/dev/regolith/regolith/database.py", line 123, in connect
    load_database(db, client, rc)
  File "/home/christopher/dev/regolith/regolith/database.py", line 61, in load_database
    load_git_database(db, client, rc)
  File "/home/christopher/dev/regolith/regolith/database.py", line 38, in load_git_database
    client.load_database(db)
  File "/home/christopher/dev/regolith/regolith/fsclient.py", line 122, in load_database

looking for now-outdated files... none found
pickling environment...     self.load_yaml(db, dbpath)
  File "/home/christopher/dev/regolith/regolith/fsclient.py", line 114, in load_yaml
    coll, inst = load_yaml(f, return_inst=True)
  File "/home/christopher/dev/regolith/regolith/fsclient.py", line 44, in load_yaml
    doc['_id'] = _id
TypeError: 'str' object does not support item assignment
```

CJ-Wright:
This seems to work properly when testing it on my local, so I'm not sure what that error is all about. @scopatz thoughts?

regolith/fsclient.py
```diff
@@ -103,13 +107,16 @@ def load_json(self, db, dbpath):
     def load_yaml(self, db, dbpath):
         """Loads the YAML part of a database."""
         dbs = self.dbs
-        for f in iglob(os.path.join(dbpath, '*.y?ml')):
+        for f in iglob(os.path.join(dbpath, '*.y*ml')):
+            if f not in db['blacklist']:
```
scopatz: Shouldn't you continue if f is in the blacklist?

regolith/fsclient.py
```diff
@@ -88,12 +89,15 @@ def is_alive(self):
...
     def open(self):
         self.dbs = defaultdict(lambda: defaultdict(dict))
+        self.chained_db = {}
         self.closed = False
 
     def load_json(self, db, dbpath):
         """Loads the JSON part of a database."""
         dbs = self.dbs
         for f in iglob(os.path.join(dbpath, '*.json')):
+            if f not in db['blacklist']:
```
scopatz: Same here

scopatz:
Thanks a million @CJ-Wright!

scopatz:
It is awesome to have this!

PR #94, opened 2018-02-18
ENH: allow db loading from local filesystems outside of vc

### LINKED ISSUE #91: Support local dbs
Currently regolith *always* pulls from a remote source to get the dbs when running any connected command. This creates a bunch of problems (eg "I wanted to validate something before committing it, but regolith only looks at the dbs already on GH"). We need a way in the RC to specify to pull from the local.

CJ-Wright:
Closes #91

### DIFF
--- a/news/build_local.rst
+++ b/news/build_local.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Regolith commands can run using a local db rather than a remote
+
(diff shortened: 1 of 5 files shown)


CJ-Wright:
Note that this hasn't been tested yet. I'll try to test with my local soon.

PR #100, opened 2018-02-19
ENH: make a latex builder base class

CJ-Wright:


scopatz:
Thanks for putting this in @CJ-Wright! I think that there is also the publists and gradesbuilders that also should be based on the new latexbuilder

regolith/tools.py
```diff
@@ -225,10 +226,20 @@ def awards_grants_honors(p):
         }
         aghs.append(d)
     for x in p.get('service', []) + p.get('honors', []):
-        d = {'description': latex_safe(x['name']),
-             'year': x['year'],
-             '_key': date_to_float(x['year'], x.get('month', 0)),
-             }
+        print(x)
```
scopatz: probably don't need this print.

regolith/tools.py
```diff
@@ -216,6 +216,7 @@ def awards_grants_honors(p):
         The person entry
     """
     aghs = []
+    print(p['name'])
```
scopatz: probably don't need this sprint

CJ-Wright:
Should we have a folder for all the builders?

CJ-Wright:
@scopatz ready for another round.

scopatz:
Eh I am a flat-is-better person in general, but if you think it is time to have a builders subfolder, that is fine too

scopatz:
This looks good to go to me barring the point you made

CJ-Wright:
I'll open an issue for that (no need to block this)

PR #110, opened 2018-04-19
MTN: moved builders to builders folder

### LINKED ISSUE #105: Builders folder
We have a few builders now, should they live in a common folder?
@scopatz prefers flat
From https://github.com/regro/regolith/pull/100#issuecomment-369632575

sbillinge:
closes #105


CJ-Wright:
This is failing due to upstream. I'll put in a fix today so that this can be rebased onto master.

PR #112, opened 2018-04-19
FIX: pin cerberus to 1.1

CJ-Wright:

PR #114, opened 2018-04-19
Revert "FIX: pin cerberus to 1.1"

scopatz:
Reverts regro/regolith#112


scopatz:
maybe don't merge this quite yet, since it does seem like it is a bug with Cerberus.  But I hope there is a way to route around it, since it sounds like we are using a deprecated API maybe?

CJ-Wright:
Maybe? A depreciation warning wouldn't have gone amiss here though :smiling_imp:

PR #125, opened 2018-05-02
Build a viable test rig based on the exemplars

### LINKED ISSUE #79: Integration tests
It might be nice to have integration tests.
These tests might include having a fake research group which we run tests against. This way we can make sure that we don't break anything major when PRing.

### LINKED ISSUE #95: Build test rig
It would be nice to have at least one test db which was not being used in production but could be used to test regolith system, commands, and more.

CJ-Wright:
This is my attempt to build a viable test rig and close #79 and close #95 
Attn: @scopatz


CJ-Wright:
@scopatz and @sbillinge ready for review!

LICENSE
```diff
@@ -118,4 +118,36 @@ For more information, please see
...
-licences.
\ No newline at end of file
+licences.
+
+For conftest.py
+
+BSD 3-Clause License
+
+Copyright (c) 2017, Anthony Scopatz
+All rights reserved.
```
scopatz: Simply add this statement to the conftest.py, you don't need to modify the license file.

regolith/builders/basebuilder.py
```diff
@@ -30,7 +30,8 @@ def __init__(self, rc):
         self.bldir = os.path.join(rc.builddir, self.btype)
         self.env = Environment(loader=FileSystemLoader([
             'templates',
-            os.path.join(os.path.dirname(__file__), 'templates'),
+            os.path.join(os.path.split(os.path.dirname(__file__))[0],
```
scopatz: Same as in other PR, this should likley be dirname

tests/conftest.py
```diff
@@ -0,0 +1,67 @@
+"""Copyright (c) 2017, Anthony Scopatz
+All rights reserved."""
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+
+import pytest
+import json
```
scopatz: json is part of the standard library and should be grouped with the first set of imports

tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    dirname : str
+        Directory to be removed
+    """
+    try:
+        shutil.rmtree(dirname)
+    except PermissionError:
+        if sys.platform == 'win32':
+            subprocess.check_call(['del', '/F/S/Q', dirname], shell=True)
+        else:
+            raise
```
scopatz: single newline and end of file please

scopatz:
Why is the PDF checked into the repo? I don't think that is a safe / stable test

scopatz:
Thanks for putting this in  @CJ-Wright! My major comment is that I think that testing against PDFs is the wrong method here, because regolith isn't itself producing the PDFs. So the contents of PDF files is beyond our control and influence and could change in many small, insignificant ways (i.e. different fonts are installed on the machine). Rather, I think we should simply test against the objects that regolith itself generates (i.e. the tex files). 

In that vein, I think it makes sense to add an option on the LaTeX builder and to the command line to not actually create the pdf, but only create the tex file.  This will make testing faster and allow the CI systems to not have to have LaTeX installed (which is like 2 GB). This is probably the right PR to add such capability.

CJ-Wright:
@scopatz thank you for the review! We don't actually test the PDF (it is black listed in the test), but I did put it there so we humans could look at it to make sure it looked ok, since we need to bootstrap the tests.

I can add the things for not producing the PDFs. One could consider the UC where the Tex documents are edited prior to the PDF production do I think it makes sense outside of just a testing situation.

scopatz:
Ok given that the PDFs are just for humans, I think they should be removed

LICENSE
CJ-Wright: I just need those two lines right?

tests/conftest.py
```diff
@@ -0,0 +1,67 @@
+"""Copyright (c) 2017, Anthony Scopatz
+All rights reserved."""
```
scopatz: yep that's right

CJ-Wright:
@scopatz ready for more review!

regolith/main.py
```diff
@@ -116,6 +116,11 @@ def create_parser():
     bldp.add_argument('build_targets', nargs='+',
                       help='targets to build. Currently valid targets are: \n{}'.
                       format([k for k in BUILDERS]))
+    bldp.add_argument('--no-pdf', dest='no_pdf',
```
scopatz: ie here the dest can be pdf

tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    for coll, example in EXEMPLARS.items():
+        dump_yaml('db/{}.yaml'.format(coll), {example['_id']: example})
+    subprocess.run(['git', 'add', '.'])
+    subprocess.run(['git', 'commit', '-am', 'Initial readme'])
+    yield repo
+    os.chdir(cwd)
+    rmtree(repo)
+
+
+def rmtree(dirname):
```
scopatz: maybe this should go into tools.py?

CJ-Wright: Yea, I vended this from rever.

scopatz: OK no worries, let's call this non-blocking then

CJ-Wright: At some point we might a) have rever as a req, b) move this to xonsh.

scopatz: > a) have rever as a req,

really?

CJ-Wright: I mean I prefer to move this to xonsh or maybe have a xonsh-utils package with stuff like this? It seems like it would be useful.

scopatz: Sure.  Feel free to add it to xonsh, if something like this isn't there already

CJ-Wright:
@scopatz is this ready to go?

scopatz:
Is it still WIP?

regolith/builders/gradebuilder.py
```diff
@@ -119,7 +119,9 @@ def latex(self):
                             max_wavg=max_wavg, curve=curve,
                             show_letter_plot=show_letter_plot,
                             **students_kwargs[student_id])
-                self.pdf(base)
+                # TODO: this seems like something for the base class to handle
+                if not self.rc['no_pdf']:
```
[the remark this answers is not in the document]
CJ-Wright: I thought I had gotten all of them drat.

CJ-Wright:
@scopatz one more time?

scopatz:
LGTM! Thanks @CJ-Wright!

CJ-Wright:
@scopatz thank you very much for all the review!

scopatz:
No problem! Thanks for addressing this long overdue definciency!!

PR #126, opened 2018-05-02
FIX: BuilderBase

CJ-Wright:


regolith/builders/basebuilder.py
```diff
@@ -30,7 +30,8 @@ def __init__(self, rc):
         self.bldir = os.path.join(rc.builddir, self.btype)
         self.env = Environment(loader=FileSystemLoader([
             'templates',
-            os.path.join(os.path.dirname(__file__), 'templates'),
+            os.path.join(os.path.split(os.path.dirname(__file__))[0],
```
scopatz: Why not just call dirname() twice, rather than split()[0]

CJ-Wright: Fair.

CJ-Wright:
Closing for #125

PR #127, opened 2018-05-04
Enum pos

### LINKED ISSUE #117: enum positions for people
This way we can have one source of truth.

CJ-Wright:
closes #117

### DIFF
--- a/docs/api/basebuilder.rst
+++ b/docs/api/basebuilder.rst
@@ -1,10 +1,10 @@
 .. _regolith_basebuilder:
 
 ******************************************************
-Builder Base Class (``regolith.basebuilder``)
+Builder Base Class (``regolith.builders.basebuilder``)
 ******************************************************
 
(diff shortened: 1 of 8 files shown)


CJ-Wright:
@scopatz ready for review!

scopatz:
Is this backwards compatible?

CJ-Wright:
I think it should be, what did we do if a position was not in the sorter dict?

CJ-Wright:
Ah I see the problem we defaulted the lookup to `''`. Do you want to run a dep cycle?

CJ-Wright:
I guess one could make an argument that they are already broken and we're just being silent about it. It would be nice if we could enter this into the schema but not have it enforced so it would show up in the docs but not be validated against.

CJ-Wright:
@scopatz I think this is ready for another look!

scopatz:
Thanks!

PR #130, opened 2018-05-09
Total weight zero bug fix

scopatz:


CJ-Wright:
At some point should we have exemplars for grades so we can run grade builder tests?

scopatz:
Yes, probably

CJ-Wright:
LGTM

PR #140, opened 2018-05-22
FIX: give validate real teeth (previously didn't print errors)

CJ-Wright:
Oops :sheep:


CJ-Wright:
Attn: @sbillinge this is why it looked like all the DBs were clean.

CJ-Wright:
@scopatz I think this is ready for review

scopatz:
Thanks!

PR #147, opened 2018-05-24
FIX: actually use ChainedDB

CJ-Wright:


CJ-Wright:
@scopatz ready for review!

regolith/chained_db.py
```diff
@@ -15,7 +15,7 @@ def __getitem__(self, key):
         results = []
         # Try to get all the data from all the mappings
         for mapping in self.maps:
-            results.append(mapping.get(key, None))
+            results.append(mapping.get(key, "--bad-request--"))
```
scopatz: I think you should define a different sentinal, something that can't possibly be a valid value,and preferably as singleton.

Check out xonsh's DefaultNotGiven for an example: https://github.com/xonsh/xonsh/blob/master/xonsh/tools.py#L232

regolith/chained_db.py
```diff
@@ -27,8 +27,9 @@ def __getitem__(self, key):
             return list(itertools.chain(*results))
         else:
             for result in reversed(results):
-                if result is not None:
+                if result is not "--bad-request--":
```
scopatz: Then this could be `result is not Singleton`

scopatz:
Just the one bit about the singleton, but otherwise looks great!

CJ-Wright:
@scopatz ready for review!

scopatz:
Thanks!

PR #155, opened 2018-06-05
WIP: initial stab at the experiments schema

CJ-Wright:
It would be great to have a builder to go with this so we could understand what is going on.

### DIFF
--- a/regolith/schemas.py
+++ b/regolith/schemas.py
@@ -577,6 +577,35 @@ SCHEMAS = {
         'year': {'description': 'Publication year',
                  'required': True,
                  'type': 'integer'}},
+    'experiments': {
+        '_description': {
+            'description': 'This collection represents experiments (physical'
+                           'or otherwise) run by members of the group.'},
+        '_id': {
+            'description': 'short representation, such as this-is-my-title',
+            'required': True,
+            'type': 'string'},
+        'lead_experimenter': {'description': 'The lead experimenter for the '
+                                             'experiment',
+                              'required': True,
+                              'type': 'string'},
+        'pi': {'description': 'The pi for the experiment',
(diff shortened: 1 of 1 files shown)


CJ-Wright:
Comments greatly appreciated @sbillinge @scopatz

regolith/schemas.py
```diff
@@ -577,6 +577,35 @@
...
+            'required': True,
+            'type': 'string'},
+        'lead_experimenter': {'description': 'The lead experimenter for the '
+                                             'experiment',
+                              'required': True,
+                              'type': 'string'},
+        'pi': {'description': 'The pi for the experiment',
+               'required': True,
+               'type': 'string'},
+        'identifiers': {'description': 'List of identifiers for the experiment'
```
sbillinge: shall we call this `descriptors`?  For our purposes things like beamline, facility etc. will go in here.  In that case it might be nicer for this to be a dict?  e.g., I could imagine wanting to retrieve all experiments where exp['descriptors']['beamlines'] == xpd` in some date-range or sthg

the list of descriptors will be different for different kinds of experiments so it is nice to make this very user-configurable (i.e., have the 'descriptors' level)

CJ-Wright: Yes, I was thinking about making it a dict. My initial plan was to put the SAF information there as an example. Maybe call this `tags`?. Or maybe `tags` is separate? One could imagine listing all the SLURM ids associated with a computational campaign.

sbillinge:
@CJ-Wright I guess we will close this when I/we return to the project/experiment schema.  Songsheng has actually made a beamtime builder too which maybe captures some of this.

CJ-Wright:
great!

PR #168, opened 2018-06-10
Groups in people schema

### LINKED ISSUE #167: group in people schema
This should go under education and under employment and be optional.

The idea is to find group members by linking people to groups in the group yml rather than having to maintain lists of current and former people in a ``team`` list somewhere.

sbillinge:
closes #167 

people can be linked back to a group in the groups.yml

Also added department to education schema.

I made "location" not required because if you specify institution and it can be found in institutions collection you don't need it.  Not sure if this is the best idea because people can then not have a location and cvs won't build correctly.


news/groups_in_people_schema.rst
```diff
@@ -0,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+ENH: added group item in people schema
```
CJ-Wright: This needs `* ` before the text, also you don't need the `ENH:` for this (although it is ok if you want).

regolith/schemas.py
```diff
@@ -838,7 +847,10 @@
                                'type': 'integer'},
                            'end_month': {'required': False, 'type': 'string'},
                            'end_year': {'required': False, 'type': 'integer'},
-                           'location': {'required': True, 'type': 'string'},
+                           'group': {'required': False, 'type': 'string',
```
CJ-Wright: Should we have a higher level of validation where we check that `fuzzy_finder` finds the linkages between collections? (eg someone puts in a bad group name).

regolith/schemas.py
```diff
@@ -811,13 +814,19 @@
                                'type': 'string'},
                            'begin_year': {'required': True, 'type': 'integer'},
                            'degree': {'required': True, 'type': 'string'},
```
CJ-Wright: On a side note, does this field coupled with begin/end date help to track the highest degree sought?

sbillinge: I did think of that.  The challenge is that this field is a free-write string.  But if we could parse it for a degree from a list of degrees (BA, BSc, BS, PhD etc.) I guess it could work.    Most entries look like "PhD in Materials Science" or sthg, so it shouldn't be too hard.

regolith/schemas.py
sbillinge: We will need that for this infrastructure to all work.  Because the "condition" is not just that ``group`` contains a ``string``, but that the ``string`` in ``group`` is in ``groups.yml``

This is all getting very complicated, and at some point not worth it maybe, but if we can do it, it will take a lot of pressure off the data-entry and bug chasing of data entry later, so will make the students much more happy!

news/groups_in_people_schema.rst
sbillinge: ok, fixed in the next commit

regolith/schemas.py
CJ-Wright: I think this can be handled by a separate validation system. It would not be hard to:
1. add something to the schema marking this as a cross-linked field and what it is cross linked to
1. Have a separate validation system which checks that the `fuzzy_finder` finds something for this, if not say something.

regolith/schemas.py
```diff
@@ -248,7 +248,9 @@
                               'Nuclear and Radiation Engineering '
                               'Program',
                     'end_year': 2011,
-                    'institution': 'The University of Texas at Austin',
+                    'group': 'ergs',
+                    'institution': 'columbiau',
```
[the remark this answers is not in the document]
sbillinge: yes, its breaking that test because I blew away the name being matched in the test.

I can put it back as it was so it passes the test, and we can put in an issue to use our nice ``institutions`` colletion wherever possible in the cv and resume builders.

sbillinge:
@CJ-Wright @scopatz  I think this is ready for review/merge

scopatz:
Thanks @sbillinge!

PR #172, opened 2018-06-11
Preslist for all

### LINKED ISSUE #171: preslist builder builds preslist for all group members
1 find all group members, past and present
2 find all presentations that contain that group member in the authors
3 make the list and write it to a file with that person's name in the filename

sbillinge:
closes #171 

getting started on preslist for all.

First push is just a method for getting all group members from people if the group ``'bg'`` is tagged in their ``employment`` or ``education``.  This seems to be working.

@CJ-Wright  @scopatz I wonder if the ``list_group_members`` method should be a function in tools.py rather than here though?


regolith/builders/preslistbuilder.py
```diff
@@ -31,12 +31,31 @@ def construct_global_ctx(self):
         gtx['str'] = str
         gtx['zip'] = zip
 
+    def list_group_members(self, grp):
+        # get all group members
```
scopatz: Can you please add a docstring here

regolith/builders/preslistbuilder.py
```diff
@@ -31,12 +31,31 @@ def construct_global_ctx(self):
...
+        grpmembers = []
+        print(self.gtx['people'][0]['_id'])
```
scopatz: I don't think this print needs to present

regolith/builders/preslistbuilder.py
```diff
@@ -31,12 +31,31 @@ def construct_global_ctx(self):
...
+
+
     def latex(self):
         """Render latex template"""
         for group in self.gtx['groups']:
             pi = fuzzy_retrieval(self.gtx['people'], ['aka', 'name', '_id'],
                                  group['pi_name'])
-
+        listgrpmembers = self.list_group_members('bg')
+        print(listgrpmembers)
```
scopatz: please remove this print.

regolith/builders/preslistbuilder.py
```diff
@@ -31,12 +31,31 @@ def construct_global_ctx(self):
...
+        return grpmembers
+
+
     def latex(self):
         """Render latex template"""
         for group in self.gtx['groups']:
             pi = fuzzy_retrieval(self.gtx['people'], ['aka', 'name', '_id'],
                                  group['pi_name'])
-
+        listgrpmembers = self.list_group_members('bg')
```
scopatz: We generally don't put type names into varaible names.  I suggest renaming to just `grpmembers`

scopatz:
Thanks for putting this in @sbillinge!

regolith/builders/preslistbuilder.py
CJ-Wright: I would pull this from the `regolithrc` since there is a `groupname` field. This can be accessed by `rc.groupname`.

CJ-Wright: Would it be possible to build preslists for every member of the group? Maybe name them `presentations_<id>.tex`?

sbillinge: OK, I ran into a problem with this.  If we want to iterate over more than one group (e.g., columbia and BNL groups) I would need to run the program twice with a manual edit of the rc file in between.

Also, this bifurcates the definitive place where groups are specified.  We have a ``groups.yml`` which is the groups we are tracking, and an rc.groupname which is some name/tag that is specified in the rc.  These two sources of truth don't talk to each other.   For example, the group specified in ``rc.groupname`` may not even exist in groups.yml.  I would rather be able to specify a group id when I run the program, so I could do things like ``[groupnames.append(list_group_names(grp)) for grp in ['bg','bnl']]`` (or somesuch that keeps the list items unique.)
Is there a reason to not do this and to use the rc instead?

sbillinge:
Not quite ready to be merged yet....I can't figure out how to build the list of groups....

Also, I can't test if the outer loop is working in the preslist-for-all because my version crashes with a subprocess error the first time through the loop. @CJ-Wright any progress on this by any chance?

CJ-Wright:
I was thinking about that today, @scopatz would moving to xonsh for all subprocess calls fix the windows subprocess problems?

scopatz:
It couldn't hurt :)

sbillinge:
@scopatz @CJ-Wright this needs tests, for the new method, but it now builds presentation lists for all group members of all groups.  I will need help making the test.

I can get it to run on my windows by commenting out the pdf line, which is where the subprocess call is.

It does not filter for accepted, which is probably the topic of another PR.

regolith/builders/preslistbuilder.py
```diff
@@ -31,30 +31,81 @@ def construct_global_ctx(self):
...
+        with a name and an id.
+        - People are in a group during an educational or employment period.
+        - To assign a person to a tracked group during one such period, add
+        a "group" key to that education/employment item with a value
+        that is the group name.
+        - This function takes the group id that is passed and searches
+        the people collection for all people that have been
+        assigned to that group in some period of time and returns a list of
+        """
+        grpmembers = []
```
CJ-Wright: You might consider using `set` here, since it makes sure that all elements are unique (allowing you to remove the `if person...` syntax)

regolith/builders/preslistbuilder.py
```diff
@@ -31,30 +31,81 @@ def construct_global_ctx(self):
...
+        for person in self.gtx['people']:
+            for position in person.get('education', {}):
```
CJ-Wright: Would it be more concise to for loop through education and group the keys?
eg
```python
grpmembers = {}
for person in self.gtx['people']:
    for k in ['education', 'employment']:
        for position in person.get(k, {}):
            grpmembers.add(person['_id'])
```

regolith/builders/preslistbuilder.py
```diff
@@ -31,30 +31,81 @@ def construct_global_ctx(self):
...
-                                                      pres['institution'])
-                if 'department' in pres:
-                    pres['department'] = pres['institution']['departments'][
-                        pres['department']]
-        self.render('preslist.tex', 'presentations.tex', pi=pi,
-                    presentations=presentationsdict)
-        self.pdf('presentations')
+            grp = group['_id']
+            grpmember_ids = self.group_member_ids(grp)
+            for member in grpmember_ids:
```
CJ-Wright: I think this is going over the IDs but the authors in the presentations will be their names. Does this clash with line 83?

CJ-Wright: Would this be easier as:
```python
for p in self.gtx['people']:
    if p['_id'] not in grpmember_ids:
        continue
```
(This would skip the rest of the loop for any person not in the group)

This would give you direct access to the person in question.
This might also let you use [`filter_publications`](https://github.com/regro/regolith/blob/master/regolith/tools.py#L105) to filter the presentations by the person.

(I'm drawing inspiration from the [cvbuilder ](https://github.com/regro/regolith/blob/master/regolith/builders/cvbuilder.py))

sbillinge: ideally we need a fuzzy match here.  I have been entering "group standard" name-ids as author names but that is just me.  This will break when people start putting loosy-goosy names.

I should probably use ``fuzzy_retrieval`` and then get the ``id`` from the returned document here.

sbillinge: I tried really hard to make something like this work, but the skipped entries were still in the collection, just in the wrong format, so it was breaking the jinja2 when it was passed.  I can go look at ``cvbuilder`` though, as this same problem was presumably solved over there.

regolith/builders/preslistbuilder.py
sbillinge: Thanks Chris, that is cleaner.

regolith/builders/preslistbuilder.py
```diff
@@ -1,5 +1,24 @@
...
+"""Builder for Lists of Presentations.
+
+This builder will build a presentation list for each group-member in each group
+listed in groups.yml.  Group members are indicated in the employment and
+education sections of the individual in people.yml.
+
+There are a number of filtering options, i.e., for presentation-type (invited,
+colloquium, seminar, poster, contributed_oral) and for whether the invitation
+was accepted or declined.  As of now, these filters can only be updated by
+editing this file but may appear as command-line options later.  It will also
```
CJ-Wright: Do you want me to put up a PR adding these options?

regolith/builders/preslistbuilder.py
```diff
@@ -1,5 +1,24 @@
...
+get institution and department information from institutions.yml if they are
+there.
+
+The author list is built from information in people.yml where possible.  The
+does a fuzzy search for the person in people.yml but if the person is absent
+from people, it will still build but using the string name given in
```
CJ-Wright: :heart:

regolith/builders/preslistbuilder.py
sbillinge: @CJ-Wright do you mean command-line options?

Definitely things are heading in that direction for me anyway.  I sometimes need all my invites (for the Dean) but sometimes only the talks I gave (for my cv or my grant report) and the Dean wants it broken out by seminar vs invited vs plenary and the grant just wants invited talks, and so on and so on and so on.  I don't think it is a good idea to build a new builder for every option, so capturing the most common ones with filters and allowing to specify which filters at runtime seems like a good way to go.

CJ-Wright: Yes, we could add some optional kwargs to the CLI which could specify these different situations. That would prevent us from having to build a ton of builders (which would be painful).

CJ-Wright:
@sbillinge do you want this merged or do you want to wait for the new schema?

regolith/schemas.py
```diff
@@ -183,23 +183,26 @@
...
-                         'chemistry': 'Department of Chemistry',
-                         'apam': 'Department of Applied Physics'
-                                 'and Applied Mathematics',
-                         'cheme': 'Department of Chemical'
-                                  'Engineering',
-                         'cs': 'Computer Science Department',
-                         'eaee': 'Department of Earth and'
-                                 'Environmental Engineering'
-                     },
+                     'departments': {{'_id': 'physics',
```
CJ-Wright: I think you have an extra curly brace here. That denotes a set syntax which is causing problems.

sbillinge:
Thanks Chris....

CJ-Wright:
Whelp, there seems to be a bug in cerberus for this.
I think the expectation here is that we can tell cerberus that we don't know what the keys of the dict will be but we do know what the values for those unknown keys are.

I have posted an issue to stack overflow: https://stackoverflow.com/questions/50994398/allow-unknown-keys-but-validated-values-in-schema we'll see what we can get back as a response.

CJ-Wright:
Yea, they had an all caps telling me to open a SO first. I will open an issue though, it seems that `allow_unknown` destroys all sub schemas which might not be a great idea.

CJ-Wright:
xref: https://github.com/pyeve/cerberus/issues/416

sbillinge:
Thanks Chris.  OK, I am not going mad then....

sbillinge:
@CJ-Wright @scopatz I think this is ready for final review.  Would love to pull it in. 

I will make a separate PR for humanName and logic for printing days in the list if they are present. I am sick of looking at all those commits....

CJ-Wright:
I think this needs a news entry

sbillinge:
@scopatz @CJ-Wright  please look at the last commit and comment.

In the end I decided to do the string formatting in python because it was too hard for me to figure out how to build the style file for latex.....
It is probably reproducing something out there already, but I couldn't find it, and this gets the job done I think.

Any chance we can pull this in soon?  I want to keep working on improvements to this list but would rather do it on new PR's but they will all build on what is in this PR.

regolith/stylers.py
```diff
@@ -0,0 +1,26 @@
...
+
+import re
+
+
+def sentencecase(string):
+    """returns a string in sentencecase but with text in braces preserved
+
+    parameters
+    ----------
+    string
```
CJ-Wright: I think this needs to be `string ;`

news/stylers.rst
```diff
@@ -0,0 +1,17 @@
+**Added:**
+
+* a stylers.py module
+* a function that puts strings into sentence case but preserving capitalization
+of text in braces
```
CJ-Wright: this needs two spaces

regolith/builders/preslistbuilder.py
```diff
@@ -31,30 +50,125 @@ def construct_global_ctx(self):
...
+                    if pres['status'] in statuses or 'all' in statuses:
+                        secondclean.append(pres)
+                # only list the presentation if it is invited
+                for pres in secondclean:
+                    if pres['type'] in types or 'all' in types:
+                        presclean.append(pres)
+
+                # format the filtered collection
+                # make title sentencecase but preserve capitalization of things in braces
+                pres['title'] = sentencecase(pres['title'])
```
CJ-Wright: I would pass `sentencecase` into the jinja2 render and use it in the template.

regolith/stylers.py
```diff
@@ -2,6 +2,12 @@
 
 import re
 
+month_fullnames = {1: 'January', 2: 'February', 3: 'March', 4: 'April', 5: 'May',
```
CJ-Wright: Can these go with the other month translation maps in https://github.com/regro/regolith/blob/master/regolith/dates.py ?

CJ-Wright:
I think this needs a test I'll put one in.

CJ-Wright:
Thank you very much @sbillinge!

sbillinge:
my pleasure.  That was fun.

PR #175, opened 2018-06-16
schools and departments in institutions are now dicts of dicts

sbillinge:
updated schemas for the new structure for departments and schools as dicts in institutions

### DIFF
--- a/regolith/schemas.py
+++ b/regolith/schemas.py
@@ -762,11 +762,9 @@ SCHEMAS = {
         'departments': {'description': 'all the departments and centers and'
                                        'various units in the institution',
                         'required': False,
+                        'type': 'dict',
                         'schema': {'type': 'dict',
                                    'schema': {
-                                        '_id': {
-                                            'required': True,
-                                            'type': 'string'},
                                         'name': {
                                             'description': 'The canonical name',
                                             'required': True,
@@ -785,11 +783,9 @@ SCHEMAS = {
                                    'be used for larger divisions in big '
                                    'organizations',
                     'required': False,
+                    'type': 'dict',
                     'schema': {'type': 'dict',
                                'schema': {
-                                   '_id': {
-                                       'required': True,
-                                       'type': 'string'},
                                    'name': {
                                        'description': 'The canonical name',
                                        'required': True,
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -85,6 +85,8 @@ def make_bad_db():
         else:
             d = {example['_id']: example}
         d.update({'bad': {'_id': 'bad', 'bad': True}})
+        if coll == 'presentations':
+            d.update({'institution': 'noinstitution', 'department': 'nodept'})
```
CJ-Wright: I think this still needs an `_id` and a top level key. eg
```python
{'bad2': {'_id': 'bad2', ...}}
```

regolith/schemas.py
```diff
@@ -184,21 +184,22 @@
...
-                         'physics': 'Department of Physics',
-                         'chemistry': 'Department of Chemistry',
-                         'apam': 'Department of Applied Physics'
-                                 'and Applied Mathematics',
-                         'cheme': 'Department of Chemical'
-                                  'Engineering',
-                         'cs': 'Computer Science Department',
-                         'eaee': 'Department of Earth and'
-                                 'Environmental Engineering'
+                         'physics': {'name': 'Department of Physics',
```
CJ-Wright: Due to this change I think you need to change the builders which rely on this, since they are expecting `dict['physics']` to be a string, but now it is a dict (I think you need `dict['physics']['name']`).

sbillinge: yup, this will break the builder, but I can fix it again when the new schema is fixed and pulled in.

CJ-Wright: It might be nice if we could not break the builder. If you don't want to handle it in this PR, do you want to merge this PR with the other preslist PR?

CJ-Wright: You could pull this branch into your other branch, you could rebase one onto the other, you could issue a PR into yourself, you could copy paste this code into the other branch (although I don't suggest that so much), there are a couple options. Whichever one works the best for you.

sbillinge: you mean there is no "merge PRs" button on GitHub?  :(

CJ-Wright: Not to my knowledge @scopatz thoughts?

sbillinge: just kidding.....I will rebase and close the other PR

sbillinge:
This was rebased onto PR #172 so I am closing this PR

PR #187, opened 2018-07-03
Figstore

CJ-Wright:


CJ-Wright:
Attn: @sbillinge filestore for figures. Also build tex documents based off of data stored in the database.

regolith/builders/figurebuilder.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class FigureBuilder(LatexBuilderBase):
+    """Build resume from database entries"""
+    btype = 'figure'
+    
+    def __init__(self, rc):
+        super().__init__(rc)
+        self.env = Environment(
+            loader=FileSystemLoader(
```
CJ-Wright: Maybe do '.' and then run render on all the .tex files in the dir?

CJ-Wright:
@scopatz this needs some serious docs, but I think the code is ready for review.

scopatz:
Code looks good, but needs docs everywhere

scopatz:
Thanks for putting this in @CJ-Wright!

CJ-Wright:
@scopatz ready for docs review

regolith/broker.py
```diff
@@ -0,0 +1,88 @@
...
+        Parameters
+        ----------
+        document : dict
+            The document to add the file to
+        name : str
+            Name of the reference to the file
+        filepath : str
+            Location of the file on local disk
+        """
+        output_path = self.store.copydoc(filepath)
```
CJ-Wright: This may need a little change so that the filename is unique. Otherwise we may run into collisions (not all files can be called `fig1.pdf`).

CJ-Wright: I'm going to let this be for now, we currently error on same names without force and that might be ok for now.

CJ-Wright:
Might need to default RC to false for force

scopatz:
Looks like there are some conflicts with master

CJ-Wright:
@scopatz I think the conflicts have been fixed

PR #190, opened 2018-07-09
case insensitive option added to fuzzy_retrieval

sbillinge:
I wanted to try an option where we don't have to worry about capitalization in ``fuzzy_retrieval()``.  This will make it easier building aka lists, but could lead to more false-positives in the retrieval.  Anyway, I added it as an option with current behavior (case sensitive) as default so it should be safe.   I also mistakenly committed changes to ``preslistbuilder`` which are not strictly part of this PR, but....
@scopatz @CJ-Wright I think this is ready for review

### DIFF
--- a/news/case_insensitive_fuzzy.rst
+++ b/news/case_insensitive_fuzzy.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* option for fuzzy_retrieval to be case insensitive
+
+**Changed:** None
+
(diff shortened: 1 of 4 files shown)


scopatz:
Thanks @sbillinge!

sbillinge:
wow, my last PR had 10^4 commits and is still not merged, and I managed to do everything in one commit this time....I am improving..... :)

PR #191, opened 2018-07-12
Little bit of schema cleaning in Institutions

sbillinge:
WIP please wait to review

### DIFF
--- a/news/clean_institutions_schema.rst
+++ b/news/clean_institutions_schema.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* institutions schema to add street and make conditionals work better
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/news/clean_institutions_schema.rst
+++ b/news/clean_institutions_schema.rst
@@ -1,4 +1,6 @@
-**Added:** None
+**Added:**
+
+* added banner to groups schema, which is an image for website banner
 
 **Changed:**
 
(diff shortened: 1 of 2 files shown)


scopatz:
Just let us know when it is ready for review!

sbillinge:
@CJ-Wright one thing I would like is a conditional validation on 'state' such that it is required when 'country' is 'USA' but not required otherwise.  Right now it is failing validation when it finds a 'state' field when 'country' is not 'USA'

CJ-Wright:
I can try to get that to work, I think there was another case like that too (with seminar type and department).

sbillinge:
@CJ-Wright it already has a conditional validation, but it seems to be not quite working as desired.  We somehow want the condition on whether or not the field is required, rather than whether or not it is present.  I have no idea how one would do it, but I think this will be a more common situation so would be super-userful to figure out.....

sbillinge:
it looks as if something like this could work:
```
import cerberus

v = cerberus.Validator()
schema1 = {"type": {"required": True, "allowed": ["folder"]}, "layer": {"required": True}}
schema2 = {"type": {"required": True, "allowed": ["category"]}}

schema = {"oneof": [schema1, schema2]}
doc = {"type": "folder", "layer": "a"}
v = Validator(schema)
v.validate(doc)
print v.errors
```
I will try it.

sbillinge:
@CJ-Wright @scopatz this is ready for review.  I either need help with the ``!= USA`` within a dependency, or we could pull this in, since it is working, but create a new issue for the better not equal logic that I want.

sbillinge:
@scopatz this should be ready to merge now.  I want to start working on the group web-page, which won't build without this merge, and another one to follow with small schema tweaks in people.  Thanks so much.

scopatz:
thanks again!

PR #194, opened 2018-07-17
added orcid id to people schema

sbillinge:
it is a brave new world and we all have orcid ids which differentiate our researcher selves.


sbillinge:
whoops, rogue files in there from another branch.....I will blow up this PR and redo it.

PR #205, opened 2018-08-15
fix open

CJ-Wright:


CJ-Wright:
As soon as the CI is done I'll merge. Thank you @scopatz for the review!

CJ-Wright:
I guess ruamel got updated, yey!

CJ-Wright:
Can we/is there a good way to turn the loaded yaml back into base types?

scopatz:
`p.get("aka", [])` probably needs to be `list(p.get("aka", []))`

CJ-Wright:
yes but then I get `AttributeError: 'CommentedSeq' object has no attribute 'sort'`

scopatz:
what about `[x for x in p.get("aka", [])]`

CJ-Wright:
Sorry I should have been more explicit, that works for that line, but then we get further issues from yaml where the types aren't right. My sort comment was for education. I'll patch things as I can but if I need to put in more than ~5 I might look at some sort of recursive re-typing on yaml load.

CJ-Wright:
@scopatz ready for another round!

scopatz:
Thanks! Given this change, we probably want a hotfix release

CJ-Wright:
Thank you! Will do

PR #216, opened 2018-08-31
no try empty commit

### LINKED ISSUE #215: check for dirty git before commit
This bypasses a common error which xonsh is catching properly currently.

CJ-Wright:
Closes #215


CJ-Wright:
This might not be needed since xonsh does the right thing now.

PR #220, opened 2018-09-21
better fix for the suffix spacing

sbillinge:
dang, that previous fix didn't fix it, but now with @CJ-Wright 's help we have a fix.  Jinja2 is a sensitive and delicate, and slightly irrational, beast


scopatz:
Thanks @sbillinge!

PR #231, opened 2018-10-30
fixes URLs in CV builder to be latex-safe

scopatz:
and some black changes.  CC @CJ-Wright


scopatz:
That didn't seem to work

tests/outputs/cv/scopatz.tex
```diff
@@ -327,29 +327,29 @@
...
-    { 
+    {
         \resitem{This course covers conventional reactors.}
-      
-      
+
+
         \resitem{
-            \href{https://docs.google.com/document/d/1uMAx_KFZK9ugYyF6wWtLLWgITVhaTBkAf8-PxiboYdM/edit?usp=sharing}{Syllabus}.
+            \href{https://docs.google.com/document/d/1uMAx\_KFZK9ugYyF6wWtLLWgITVhaTBkAf8-PxiboYdM/edit?usp=sharing}{Syllabus}.
```
CJ-Wright: Did you mean to put in a `\` before the `_`?

scopatz: Yep, absolutely.  `_` on their own aren't renderable.

scopatz:
It seems like the trailing whitespace is being kept in for some reason.

tests/outputs/cv/scopatz.tex
CJ-Wright: Yes but `\href` doesn't render the URL, right? It should give a link named `Syllabus` and the link is the URL.

scopatz: Maybe it is just `#` that fail to render without and escape then.

scopatz:
Alright @CJ-Wright - I believe that this is all fixed up.

PR #234, opened 2019-01-10
Clean reimb builder

sbillinge:
fixes bug that builder couldn't build more than one reimbursement without dying (thanks for the help @CJ-Wright ). 

I had big plans for this PR (to handle grants differently/properly in the reimbursements and to allow for more than one grant) but it was a bigger job than I originally thought and some people need reimbursements so this is just a bug-fix PR now.  I would like it merged if possible and make a new one for that refactor

### DIFF
--- a/regolith/builders/reimbursementbuilder.py
+++ b/regolith/builders/reimbursementbuilder.py
@@ -53,12 +53,13 @@ class ReimbursementBuilder(BuilderBase):
             key=position_key,
             reverse=True,
         )
-        for n in ["expenses", "projects", "grants"]:
+        for n in ["expenses", "grants"]:
             gtx[n] = all_docs_from_collection(rc.client, n)
         gtx["all_docs_from_collection"] = all_docs_from_collection
 
     def excel(self):
         gtx = self.gtx
+        #TODO: don't build all expense reports from all time
         for ex in gtx["expenses"]:
             # open the template
             wb = openpyxl.load_workbook(self.template)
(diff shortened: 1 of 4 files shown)


regolith/builders/reimbursementbuilder.py
```diff
@@ -2,6 +2,7 @@
 
 import datetime
 import os
+from pprint import pprint
```
CJ-Wright: Can you remove this unused import?
```suggestion
```

regolith/builders/reimbursementbuilder.py
```diff
@@ -70,9 +71,12 @@ def excel(self):
             project = fuzzy_retrieval(
                 gtx["projects"], ["name", "_id"], ex["project"]
             )
-            grant = fuzzy_retrieval(
-                gtx["grants"], ["name", "_id"], project["grant"]
-            )
+            #            grant = fuzzy_retrieval(
```
CJ-Wright: Can you remove these comments?

regolith/tools.py
```diff
@@ -406,24 +406,17 @@ def fuzzy_retrieval(documents, sources, value, case_sensitive=True):
...
-            if isinstance(v, list):
-                fs.extend(v)
-            else:
-                fs.append(v)
         returns = []
         for k in sources:
             ret = doc.get(k, [])
             if isinstance(ret, str):
                 returns.append(ret)
+            elif isinstance(ret, int):
```
CJ-Wright: this most likely should be combined with the above so if it is not a list then it makes it into a list and extends returns by it.

sbillinge: how about:
```
if not isinstance(ret, list):
    ret = [ret]
elif ret:
    returns.extend(ret)
```

CJ-Wright: I think we can remove the `elif ret:` and de-indent the `returns.extend(ret)` since we want to extend the returns with ret.

sbillinge: yes, it is not needed because we set up get to return [] if it can't find k.

OK, look for the next push.

regolith/builders/reimbursementbuilder.py
CJ-Wright: Do you still need the comments then?

CJ-Wright:
thank you very much @sbillinge

PR #236, opened 2019-02-17
post-doc ad builder

sbillinge:
Hi @scopatz @CJ-Wright 
I don't want this pulled in because it is rather specific to MY needs, but I am making a PR to initiate some discussion about this issue....

I want to build builders with templates that are specific to my needs.  Perhaps others could adapt them for themselves, or not, I don't really care.  What I think this calls for is something like a local template database for these things, and perhaps local builders, I am not sure.

For now I just put this into a branch so I can build it locally, but to avoid it getting too complicated for me with hundreds of bespoke branches for this and that builder, I wonder if we could discuss what would be a good way to handle this more generally.

Not sure if I am being clear but....let me know your thoughts.

### DIFF
--- a/regolith/templates/postdocad.tex
+++ b/regolith/templates/postdocad.tex
@@ -1,0 +1,41 @@
+\documentclass{letter}
+\usepackage{times}
+
+\begin{document}
+\begin{center}
+ {\bf
+  Post-Doctoral Fellowship Opportunity
+ }
+\end{center}
+\vskip 0.5 in 
+
+The Billinge group is seeking to fill a post-doctoral research position on {{positionOn}}.
+The goals of the project are {{projectGoals}}.
+The project will involve {{projectTasks}}.
+
+we seek candidates who are recent Ph.D. graduates in materials science, chemistry, physics, earth science, data science, computer science or a related field. {{requiredExperience}} Knowledge of x-ray, neutron or electron diffraction/scattering,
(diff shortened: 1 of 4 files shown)


regolith/templates/postdocad.tex
```diff
@@ -0,0 +1,41 @@
...
+\noindent{\bf Application Process}
+
+To apply, please submit the following materials to sb2896@columbia.edu with [{{latex_safe(k)}} post-doc] followed by your name in the subject line.  Candidates are encouraged to apply as early as possible and preferably before {{ applicationDeadline }}.
+
+Please submit as attachments to the email:
+\begin{itemize}
+\item A cover letter that explains your motivation for applying to this program
+\item A curriculum vitae (including a list of publications)
+\item PDF copies of the three publications that you are most proud of that describe work that was led and predominantly carried out by you.  
+\item {[optional]} links to any open-source software projects that you have contributed to significantly, and a description of your contributions.
```
CJ-Wright: Awesome!

CJ-Wright:
Context, we discussed the potential for plugin like things. Also should there be an HTML version of this as well? eg https://github.com/regro/regolith/blob/1b102f577f4a968464309d7a96c3d172f02d462b/regolith/templates/jobs.html

scopatz:
Build failures seem unrelated. Thanks @sbillinge!

PR #241, opened 2019-05-25
schema updates for proposals and contacts

sbillinge:
contacts is a lightweight people.  I thought of calling it collaborators, which is the main target, but I didn't want to restrict it to that.

### DIFF
--- a/regolith/schemas.py
+++ b/regolith/schemas.py
@@ -590,7 +590,7 @@ EXEMPLARS = {
             "nsf_pot_to_benefit_society": [],
             "requester": "Lane Wilson",
             "reviewer": "sbillinge",
-            "status": "invited,accepted,declined,downloaded,inprogress,submitted",
+            "status": "submitted",
             "summary": "dynamite proposal",
             "title": "A stunning new way to cure Malaria",
             "year": 2019,
@@ -923,7 +923,7 @@ SCHEMAS = {
         },
     },
     "contacts": {
-        "_description": {"a lighter version of people.  Fewer required fields"
+        "_description": {"description": "a lighter version of people.  Fewer required fields"
                          "for capturing people who are less tightly coupled"
                          },
         "_id": {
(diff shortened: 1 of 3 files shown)


regolith/schemas.py
```diff
@@ -1307,9 +1396,11 @@
             "required": False,
             "type": "string",
         },
+        # TODO: Can this be required only if status = active?
```
scopatz: I think so.  CC @CJ-Wright

scopatz:
Thanks for putting this in @sbillinge! Looks like there are some test failures related to the schema changes

regolith/schemas.py
CJ-Wright: Hmm I think this would work in reverse. You can do something similar to the code here: https://github.com/pyeve/cerberus/issues/274#issuecomment-262670804 
In this case do an `anyof` with `True` and `False` and set `True`'s dependencies so that all the needed fields must be supplied to be active.

sbillinge:
thanks @scopatz I will look into what is breaking things.

regolith/schemas.py
```diff
@@ -877,10 +922,60 @@
...
+            "required": False,
+            "anyof_type": ["list", "string"]
+        },
+        "title": {
+            "description": "how the person is addressed",
+            "required": False,
+            "type": "string",
+        },
+        "required": False,
+        "type": "dict",
```
CJ-Wright: Are these two lines needed?

sbillinge: I removed....let's see if that fixes it.

sbillinge:
@scopatz @CJ-Wright I wrestled this under control finally (thanks for the help CJ).  Ready for review and merge now.  I had to remove a ``dependency`` that wasn't working, by making ``position`` not required, but I think that is the right behavior in any case, but please take a look.

CJ-Wright:
lgtm

PR #246, opened 2019-06-24
Prop rev builder

sbillinge:
@scopatz Dang, you merged while I was on a flight working on the schemas!  

Thanks so much for doing that.  But now I think it _is_ just about finished.  It now builds proposal reviews for doe and nsf and has been tested in the real-world with a bunch of proposals........

I rebased onto upstream master so this should probably go in easily.


scopatz:
haha oh no!

PR #247, opened 2019-06-27
trying to get tests to pass on windows

sbillinge:
``test_round_trip`` now passes, but still some subprobcess errors on the builders that I am trying to hunt down.  This is stlil a WIP

### DIFF
--- a/news/TEMPLATE.rst
+++ b/news/TEMPLATE.rst
@@ -6,8 +6,6 @@
 
 **Removed:** None
 
-**Fixed:**
- - FIX: tests to run on windows OS by removing
-   removed directory paths
+**Fixed:** None
 
 **Security:** None
--- a/tests/test_broker.py
+++ b/tests/test_broker.py
@@ -1,19 +1,18 @@
 import os
 import subprocess
+import tempfile
 
 from regolith.broker import Broker, load_db
 
-import tempfile
 
-tmpdir = tempfile.gettempdir()
+TMPDIR = tempfile.gettempdir()
+
 
 def test_round_trip(make_db, tmpdir):
     repo = make_db
     os.chdir(repo)
     subprocess.check_call(["touch", "myfile.tex"], cwd=tmpdir)
     db = load_db()
-    db.add_file(
-        db["projects"]["Cyclus"], "myfile", os.path.join(tmpdir, "myfile.tex")
-    )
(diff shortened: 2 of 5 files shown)


sbillinge:
@scopatz @CJ-Wright I have the hard-coded filename things passing on windows now, so I can run tests.  Please take a look and merge if you like it.

I talked to CJ and the fix for the hard coded filepaths in the ``tests\outputs" is a little tricky so my workaround is just to avoid the test if it is one of the failing patterns.   The test will still fail if the output file is not successfully created by the test.   I think this is ok, because testing that the ``tmp`` directory filepath is a specific value is just testing how the test software creates tmp files which isn't testing the actual code functionality.....and I feel I have put enough time into this for the time being....want to get back to actually developing the code!

sbillinge:
closes #248

sbillinge:
@scopatz it would be a help if this could be pulled in and I can try fixing failed tests without using Travis....Thanks so much!

news/TEMPLATE.rst
```diff
@@ -6,6 +6,8 @@
 
 **Removed:** None
 
-**Fixed:** None
+**Fixed:**
+ - FIX: tests to run on windows OS by removing
+   removed directory paths
```
scopatz: I think you edited the template rather than the actual news file

tests/test_broker.py
```diff
@@ -3,6 +3,9 @@
 
 from regolith.broker import Broker, load_db
 
+import tempfile
```
scopatz: this import should be above with `os` and subprocess.

tests/test_broker.py
```diff
@@ -3,6 +3,9 @@
...
+
+tmpdir = tempfile.gettempdir()
```
scopatz: Global variables should be upercase, so `TMPDIR = ...`

news/TEMPLATE.rst
sbillinge: oops, sorry about that.

tests/test_broker.py
sbillinge: done

tests/test_broker.py
sbillinge: done

sbillinge:
@scopatz  this should be ready now.

sbillinge:
yipee, thanks!

PR #251, opened 2019-06-30
making date-range testing helper functions

sbillinge:
If I can give ``since`` and ``before`` date ranges on the command line, I would like to have easy to use helper functions that can test if something is valid or not in date range so I added some functions to tools.py.  Similar to ``is_current`` etc. in cpbuilder, but just tests two dates, but takes them in the format that they come from the db for convenience.

### DIFF
--- a/news/date_range_filter.rst
+++ b/news/date_range_filter.rst
@@ -1,5 +1,5 @@
 **Added:**
- * filters in tools.py that return true if a given date is since or before or
+ * filters in ``regolith.tools`` that return true if a given date is since or before or
    between other dates
 
 **Changed:** None
--- a/news/date_range_filter.rst
+++ b/news/date_range_filter.rst
@@ -1,0 +1,13 @@
+**Added:**
+ * filters in tools.py that return true if a given date is since or before or
+   between other dates
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


regolith/tools.py
```diff
@@ -116,6 +117,28 @@ def month_and_year(m=None, y=None):
     return "{0} {1}".format(SHORT_MONTH_NAMES[m], y)
 
 
+def is_since(y, sy, m=1, d=1, sm=1, sd=1):
```
scopatz: Can you please add docstrings for these functions and explain what the meaning of the function arguments is here? Thanks!

scopatz:
Thankd for putting this in @sbillinge! Just some docstrings needed

sbillinge:
yes, of course.  
I am torn and perhaps @scopatz @CJ-Wright  you can weigh in.  

Actually I started the docstrings but my pycharm helper was making rst format and I think we want numpy style, and I couldn't figure out how to switch that without googlin and I was on the plane.  Sorry, tmi, but it reminds me I need to google that....

CJ-Wright:
https://www.jetbrains.com/help/pycharm/settings-tools-python-integrated-tools.html

sbillinge:
@scopatz I think this is ready now, though I see one push I accidentally made on the wrong branch.....but it affects a different file.

news/date_range_filter.rst
```diff
@@ -0,0 +1,13 @@
+**Added:**
+ * filters in tools.py that return true if a given date is since or before or
```
CJ-Wright:
```suggestion
 * filters in ``regolith.tools`` that return true if a given date is since or before or
```

scopatz:
Thanks!

PR #255, opened 2019-07-10
added utf8 input encoding to the latex templates

### LINKED ISSUE #233: fix utf8 in presentation list builder
I built preslists recently and I saw an a with an umlaut in the db that didn't render.  Need to look into this at some point.

sbillinge:
closes #233

@scopatz this should be ready for review.  It allows us to store utf8 in the database and still have the latex build.


CJ-Wright:
You need to update the expected outputs to include the utf8 line.

sbillinge:
right.  I was thinking about how to test this last night.  I think I need
some utf8 characters in the exemplar schema also, right?

CJ-Wright:
Maybe? We don't compile the test examples.

sbillinge:
I see, you just create files and write stuff in there and see if it is the
same as the outputs files?  The real test is if the latex compiles with the
UTF8 or not, but I don't think we want latex compilation happening during
unit testing, and it is easy to catch that bug when it happens.....

CJ-Wright:
Yes

sbillinge:
@CJ-Wright  @scopatz not sure whey these tests are failing...they are on files I haven't touched.  It looks like some incompatibility on the last line of the built file ``\end{document}`` but I don't see any missing or added carriage returns or spaces.  These tests pass when I run these tests on windows,  ironically....

sbillinge:
OK, @scopatz @CJ-Wright  I think I need some help here.  I have no idea why these are failing, or why the outputs files think I have changed the ``\end{document}`` statement (which I didn't) and why the templates documents (on the whole) don't think I changed that line.....

sbillinge:
I will rebase this onto master after #258 is merged and it should pass tests at that time and we can merge this one too.  I would rather do it this way than make it pass and introduce conflicts.

sbillinge:
@scopatz @CJ-Wright This should be ready to merge now also.

Working towards getting our html to build without errors.  I think we are close now.

scopatz:
Thanks!

sbillinge:
Thanks for merging!

PR #258, opened 2019-07-28
remove that offending latex_safe

sbillinge:
so that publist will build.  Also, change colors of the links


CJ-Wright:
Can you update the tests? (the color change needs to be incorporated)

sbillinge:
@scopatz @CJ-Wright this should be ready for review now.

I finally got to the bottom of the issue that plagued this PR, and also the utf8 PR, which was the magically changing last lines of the latex code.  It turns out that the issue is that my PyCharm automatically adds a newline at the end of the file on save, which is normally good behavior.  However, jinja2 by default strips newlines from the end of the rendered file when it renders.  So the test output has to have no newline at its end.  By temporarily disabling the newline adding setting in PyCharm while I work on ``scopatz.tex`` I get the behavior that I want.  

I can no go and fix #233 in the same way.

Phew, that was driving me mad

CJ-Wright:
Yey! You found the thing!

CJ-Wright:
LGTM

sbillinge:
Yeh, so happy to find that thing, it was really annoying!

PR #259, opened 2019-07-29
FIX: removing duplicate street in institutions schema

sbillinge:
@scopatz @CJ-Wright this is a very small one.  I noticed street was defined twice in the institutions schema.


CJ-Wright:
LGTM

scopatz:
Thanks!

PR #262, opened 2019-08-07
WIP: Appraisalbuilder

sbillinge:
@scopatz  I wasn't going to add this because it was somewhat Columbia specific, but it is a builder for annual appraisal (answering the questions we have at Columbia).  However, I think it is valuable in general, and it has some updates to the schema that I would like to reuse in other developments, so I think I would like to consider merging it.

I will need to do some work to clean it so it passes tests etc., but will get back to you when I am ready.  Feel free to review at any time and give feedback.


scopatz:
No problem having institution specific code! Feel free to ping when needed

sbillinge:
@CJ-Wright could you have quick look at the failing test for the ann-appraisal builder.  It looks to me that the render command is not writing the tex file but I can't figure out why.

sbillinge:
btw, fixing the schema is still WIP, though I think the EXEMPLAR should be ok.

scopatz:
Looks like there is a test failure

scopatz:
And now a conflict with master

sbillinge:
yes, @scopatz still struggling with this one.  Trying to find time  to
finish it....

scopatz:
Ok!

sbillinge:
@mwinitch I hate to do this to you, but I really need help to get this merged.  It has some major functionality in it and schema work that is needed for other projects.  Here are the tasks that I think it needs to be merged:

- [ ] fix the date logic so it properly uses your new date logic!

- [ ] make it pass tests.  It was just failing validate inside the tests, where the schema validates against the database built from the exemplars.  this is hard to debug because there is a sys.exit that catches the error without dumping the stack.  It will take some fiddling around!  What in the exemplar is not validating in the schema.  I think it is in the grants collection but not sure, and there may be more validation fails

- [ ] change the name.  I think the builder, and all the stuff around it like the template and output filenames etc. should be called something like activity_summary instead of annual_appraisal.

- [ ] make sure there are no hard-coded things in there. When I first wrote it it just built for sbillinge.  I htink I fixed that but look for anything else that looks like that.

That's all I can think of right now.

btw, to access it, fork my fork and do hte normal workflow, making my fork upstream.  We can chat if it helps

sbillinge:
moved to #408

PR #263, opened 2019-08-28
FIX: month to int now properly handled

sbillinge:
@scopatz this is a bug fix PR

### DIFF
--- a/news/fix_int_month_preslistbuilder.rst
+++ b/news/fix_int_month_preslistbuilder.rst
@@ -1,0 +1,12 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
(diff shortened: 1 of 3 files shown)


sbillinge:
What does this mean? A bad database entry that causes it to crash?

CJ-Wright:
I would modify the exemplars so that the prior code tests would have failed, while the fix in the PR causes the test to pass.

sbillinge:
Ah gotcha

CJ-Wright:
It might not be truly needed for a simple bug fix like this, but it is good practice to make certain that the bug a PR is supposed to squash is the correct one and that we don't have the bug crop up again.

sbillinge:
OK, @CJ-Wright I checked and this new schema fails tests without the fix

CJ-Wright:
Thank you very much! LGTM

PR #269, opened 2019-09-17
adding a csv reader function for Columbia format classlists

sbillinge:
Right now it is for the version that comes from the Columbia Courseworks website, but it could be maybe made broader in application with a bit of thought.


scopatz:
LGTM! Thanks!

PR #272, opened 2019-10-17
WIP: starting to add handling of filtering based on cli

sbillinge:
now want to make use of those nice filtering options that we added to the CLI.  Not ready for review yet.


sbillinge:
Simon closing.  This is replaced by #283

PR #274, opened 2019-10-21
Is current to tools

sbillinge:
move is_current logic from current-pending builder to tools so it is easier to use it more widely.

@scopatz this is ready for review and merge


scopatz:
Thanks!

PR #278, opened 2019-10-29
Use recursively merging to update the default schemas according to user schemas

st3107:
I am Songsheng Tao in Billinge group. I would like to recommend that we use the recursively merging of two nested dictionary to update the schemas instead of iterating the first two levels of keys. It will provide the user the freedom to add their own yaml files to db and add their own properties for the entries in yaml files.

### DIFF
--- a/regolith/main.py
+++ b/regolith/main.py
@@ -11,6 +11,7 @@ from regolith import commands
 from regolith import storage
 from regolith.builder import BUILDERS
 from regolith.schemas import SCHEMAS
+from regolith.tools import merge_nested_dicts
 
 DISCONNECTED_COMMANDS = {
     "rc": lambda rc: print(rc._pformat()),
@@ -255,20 +256,6 @@ def create_parser():
         help="If provided only validate that collection",
     )
     return p
-
-
-def merge_nested_dicts(dict0: dict, dict1: dict):
-    """Merge dict1 into dict0 recursively."""
-    for key in dict1.keys():  # iterate keys in dict1
-        if (key in dict0) and isinstance(dict0[key], dict) and isinstance(dict1[key], dict):
-            # if the key is also in dict0 and both values are dictionary,
-            # recursively merge the sub-dictionary
-            merge_nested_dicts(dict0[key], dict1[key])
-        else:
-            # if the key is not in dict0 or either of the values is not dictionary,
-            # use value in dict1 to update the dict0
-            dict0[key] = dict1[key]
-    return
 
 
 def main(args=None):
--- a/regolith/main.py
+++ b/regolith/main.py
@@ -269,7 +269,6 @@ def main(args=None):
     rc._update(ns.__dict__)
     if "schemas" in rc._dict:
         user_schema = copy.deepcopy(rc.schemas)
-        print(user_schema)
         default_schema = copy.deepcopy(SCHEMAS)
         rc.schemas = update_schemas(default_schema, user_schema)
     else:
--- a/regolith/main.py
+++ b/regolith/main.py
@@ -2,7 +2,6 @@
 from __future__ import print_function
 import copy
 import os
-import collections
 from argparse import ArgumentParser, RawTextHelpFormatter
 
 from regolith.commands import INGEST_COLL_LU
--- a/regolith/main.py
+++ b/regolith/main.py
@@ -2,6 +2,7 @@
 from __future__ import print_function
 import copy
 import os
+import collections
 from argparse import ArgumentParser, RawTextHelpFormatter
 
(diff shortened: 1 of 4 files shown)


st3107:
@scopatz @CJ-Wright @sbillinge Hello, everyone. I would like to we use the recursively merging of two nested dictionary to update the schemas instead of iterating the first two levels of keys. Here is the PR. How do you think of it?

regolith/main.py
```diff
@@ -269,9 +283,7 @@ def main(args=None):
     if "schemas" in rc._dict:
         user_schema = copy.deepcopy(rc.schemas)
         rc.schemas = copy.deepcopy(SCHEMAS)
-        for k in user_schema:
-            for k2, v in user_schema[k].items():
-                rc.schemas[k][k2].update(v)
+        merge_nested_dicts(rc.schemas, user_schema)  # merge user schemas into the default schemas
```
sbillinge: I think we don't need the comment.  It can be deleted.

regolith/main.py
```diff
@@ -257,6 +257,20 @@ def create_parser():
     return p
 
 
+def merge_nested_dicts(dict0: dict, dict1: dict):
```
sbillinge: maybe remove type?

regolith/main.py
```diff
@@ -257,6 +257,20 @@ def create_parser():
...
+    """Merge dict1 into dict0 recursively."""
+    for key in dict1.keys():  # iterate keys in dict1
```
sbillinge: please remove comments, the intent should be interpretable from the code itself.

regolith/main.py
sbillinge: I wonder if this should be imported from ``tools.py``?

st3107: I put it in tools.

regolith/main.py
st3107: Comments are removed.

regolith/main.py
st3107: done.

regolith/main.py
st3107: done.

st3107:
Hello, @sbillinge . I add six test cases for the schemas updating function. They are in test_tools.py .  The other files will be further edited after we agree on the test cases.

I have a question about the use case: Shall we allow user to change the 'type' in the default schemas? For example, user changes the 'type' from 'string' to 'anyof_type': ['string', 'integer']?

sbillinge:
changing a type from 'string' to 'anyof_type: ['string, 'integer']' may not
be needed.  Such a change should probably be merged into the main regolith.

tests/test_tools.py
```diff
@@ -187,3 +187,257 @@ def test_number_suffix(input, expected):
...
+                },
+            },
+        },
+    },
+}
+
+USER_SCHEMA0 = {
+    "expenses": {
+        "itemized_expenses": {
+            "type": "list",
```
sbillinge: delete?

st3107:
Encounter some errors in tests. Need fix ...

st3107:
The failing of tests in travis is not due to my code. The master branch failed in the same way as my branch so the problem lies somewhere else.

scopatz:
Thanks @st3107! Can you help fix up the test suite please?

sbillinge:
@chiahaoliu and I dug around a bit and it looks as if it is a problem with migrating to python 3.8, which travis has done, but we haven't locally and tests are still passing.  Cerberus is giving an AttributeError so maybe something has been deprecated.  @st3107 can you dig around and see if you can figure out what is wrong; maybe talk to Tim if you need to?

sbillinge:
update, @chiahaoliu found that cerberus 1.3.7 won't work on Python 3.8.  Apparently cerberus 1.3.8 has been released but hasn't appeared on conda-forge yet.  So this should presumably resolve itself shortly.

sbillinge:
@st3107 the cerberus problems are now fixed.  Please can you push a new commit to generate a new build?  Maybe run black to check for syntax problems then commit and push?

st3107:
@scopatz @sbillinge The build passed. Could you merget the PR?

scopatz:
Thanks!

sbillinge:
@scopatz we want to set up CI on one of our databases and it needs the changes that are in this merged PR to work.  Would it be possible to do a minor release of regolith and push it to conda-forge for us?  Thanks so much.  There is one other PR that is ready for review that could be pulled in first, or not if it isn't ready.

Also, I should mention, that if you like I am happy to be given more rights on the regolith repo to replace CJ, but only if you are ready for that.

PR #279, opened 2019-11-23
adding helper functions because sometimes we want months as integer-like strings

sbillinge:
....but with leading zero's where necessary.

@scopatz this should be ready for review


sbillinge:
@scopatz it looks as if our cerberus problems are over.  Thanks for merging that conda-forge feedstock!  I will work to generate commits on all the PRs that should be passing.

scopatz:
Thanks!

PR #280, opened 2019-12-06
importing from collections.abc not collections

sbillinge:
@scopatz This should fix an out of date collections import in runcontrol.py


scopatz:
Thanks!

sbillinge:
:)

PR #302, opened 2020-02-11
fix for loss of support for syntax in jinja2

sbillinge:


sbillinge:
Our CI builds stopped working.  Tracked it down to a Jinja2 error where this syntax:
``  {{'p. ' + pub['pages'] if 'pages' in pub else ''}}``
was no longer working in Jinja2 2.11.1 but was working in 2.10.3

The fix is to wrap this line in an ``if pub['pages']:`` clause in the template.  This then works again, even though it is just different syntax for the same thing.

sbillinge:
@scopatz @CJ-Wright please could you take a quick look at this?

Our  build of our website (rg-=db-public) started failing and I traced it back to a change on conda-forge from jinja2 from 2.10.3 (working) to 2.11.1 (not working) that I could replicate locally, and then produced a change (wrapping the offending line in an explicit if-else-endif statement instead of the implicit one) which allowed the tests to pass locally with 2.11.1, so I want to merge this in, but then it produced a xonsh error deep down in the Travis process that I don't have too much idea about so this PR is now failing, but I am a bit stuck.  I wonder if it is something quick that you can see?

It would be greatly appreciated, because this is breaking everything on our database commits.....

Thanks so much

scopatz:
Thanks!

PR #310, opened 2020-02-21
reimbursement builder now requires people.  It needs tests fixing so …

sbillinge:
@scopatz @CJ-Wright this is ready to merge.  It forces a name to be specified on the command line for reimbursement builder, which is incredibly slow .

### DIFF
--- a/regolith/builders/reimbursementbuilder.py
+++ b/regolith/builders/reimbursementbuilder.py
@@ -43,6 +43,10 @@ class ReimbursementBuilder(BuilderBase):
         super().construct_global_ctx()
         gtx = self.gtx
         rc = self.rc
+        if not rc.people:
+            sys.exit("Please rerun specifying --people and a person or list of "
+                     "people")
+
         gtx["month_and_year"] = month_and_year
         gtx["people"] = sorted(
             all_docs_from_collection(rc.client, "people"),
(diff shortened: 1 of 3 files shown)


regolith/builders/reimbursementbuilder.py
```diff
@@ -55,103 +62,121 @@ def construct_global_ctx(self):
...
+                            total_amount += item.get("unsegregated_expense", 0)
+                        except TypeError:
+                            if item.get("unsegregated_expense", 0) == 'tbd':
+                                print("WARNING: unsegregated expense in {} is "
+                                      "tbd".format(ex["_id"]))
+                                item["unsegregated_expense"] = 0
+                            else:
+                                sys.exit(
+                                    "ERROR: unsegregated expense in {} is not "
+                                    "a number".format(ex["_id"]))
```
scopatz: You should not use sys.exit in library code.  Please raise an approriate error instead

sbillinge:
thanks @scopatz I fixed that so it reraises the TypeError.  I left the sys.exit in the global-context place.  Is it ok there?

scopatz:
I really don't think we should have any sys.exits

scopatz:
I really don't think we should have any sys.exits

sbillinge:
@scopatz I think this is ready for merge.  

Can we make a release and push it to conda-forge?   This seems to be needed for the  CI to run properly again.

scopatz:
doing a release now

PR #314, opened 2020-02-28
initial commit of date-gap/overlap finding function and tests

sbillinge:


sbillinge:
@scopatz @CJ-Wright I more or less forced this to pass the codecov tests by changing the ``.codecovrc.yml`` and allowing it to report success if it doesn't find files in ``project/tests``.  Is this acceptable or is there some other problem with the configuration that should be updated to fix this?  The codecov error that we were getting was:
``codecov/project/tests — No files found while filtering the report. ``

The codecov was working and then all of a sudden stopped working a week or so ago with this error and I have no idea why.  Any suggestions are welcome, but if we merge this PR with this workaround it will make most of the other PRs start to pass CI.  Please let me know

regolith/tools.py
```diff
@@ -749,10 +749,48 @@ def update_schemas(default_schema, user_schema):
...
+    if max(loading) > 1.0:
+        status = False
+        print("max {} at {}".format(max(loading),
+                datearray[list(loading).index(max(loading))]))
+    elif min(loading) < 1.0:
+        status = False
+        print("min {} at {}".format(min(loading),
+                                    datearray[list(loading).index(min(loading))]
+                                    ))
+    print
```
scopatz: probably don't need this

sbillinge: oops that's weird, thanks for catching it.  Should be fixed now.

PR #317, opened 2020-03-18
Filter grants copis

rohanpok:
@sbillinge 
Please review and merge


sbillinge:
@rohanpok this also contains a bunch of merged items.  Did you work with someone to do this correctly?  If so, your local master may be corrupted and we will need to take more drastic action.  Please could you work with Sani or Songsheng to try and clean things up?

PR #327, opened 2020-04-19
The first adder helper that will add an entry to a collection

sbillinge:


sbillinge:
@scopatz this should be ready for merge after the helper infrastructure is merged.  I branched off the helper branch so some of the commits are the same.  I guess this won't be a problem.

scopatz:
Thanks!

PR #292, opened 2020-02-05
Move all the date filters to the dates.py which may be a better place for them

sbillinge:
@scopatz this is a WIP.  Needs tests and all kinds of things.

PR #328, opened 2020-04-22
get dates returns datetime objects from dates in documents

sbillinge:


sbillinge:
made this PR in error.  These changes are on #292

PR #330, opened 2020-04-27
new adder helper for adding a reading list to reading_lists...

sbillinge:
...based on tags on entries in the group citations database


sbillinge:
@scopatz I have to fix tests in this one, but I am fixing them in #331 so when you merge that one and I rebase this one it should pass, so please wait on this for now.

PR #331, opened 2020-04-28
An adder helper for adding a projectum to the projecta collection

sbillinge:


sbillinge:
@scopatz this should be ready for review and merge.  I will finalize the schema later when I have used it a bit so it is not validating but there is an exemplar for testing.

for these helpers I will give you a few days in case you want to review, and then merge them myself because I think they are fairly orthogonal to how you were using regolith so it shouldn't do any harm.  This should save you some time and worry.  I will only merge if tests are written and passing and news is written

sbillinge:
I just broke this by mistake, but rather than fix it, I am adding new commits in #333 and will get that to work and merge that one, and we can close this one.

PR #394, opened 2020-06-05
Added news for fragment retrieval

shogho0:


news/fragment_retrieval.rst
```diff
@@ -0,0 +1,11 @@
+**Added:** Added function fragment_retrieval that returns a list of documents from a collection where the fragment is found anywhere in the given fields(similar to fuzzy_retrieval)
```
sbillinge, line 1: this should look like:
```
**Added:** 
* New function in tools, fragment_retrieval, that returns a list of documents from a collection where the fragment is found anywhere in the given fields (similar to fuzzy_retrieval)

**Changed**
```
and so on

PR #411, opened 2020-06-08
WIP: infrastructure to build an internal html with group meetings summarized.

sbillinge:
group meeting infrastructure

### DIFF
--- a/news/news0.rst
+++ b/news/news0.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Add test for internalhtmlbuilder in test_builder.py
+
+**Changed:** None
+
+**Deprecated:** None
+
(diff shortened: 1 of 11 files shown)


sbillinge:
@berrakozer it looks as if the main internalhtml helper got merged into master in one of your other helper merges.  I am not sure how that happened, but it was merged in without any tests (this is why that merge was failing all the codecov tests).    That was a bit of a mistake and  I didn't realize that the code for this helper was in there, but it means that we should probably close this PR, but we will need tests for helper that was merged.  Is that something that you could do, if you had some help?  It basically means writing the schema for the meetings collection in the db.  I can walk you through it.

sbillinge:
closed by #603

PR #417, opened 2020-06-09
Expense helper

### LINKED ISSUE #360: helper: new expense adder
add an expense to the expenses collection

AhmedS343:
closes #360


sbillinge:
Great start!

This already needs a rebase I am afraid.  Please talk to Michael or Songsheng if you are not sure (don't try and do it yourself).

Also there are formatting errors in your yaml, just fyi....  

Please make sure you are able to run tests locally and doing that too.  Talk to someone if you need help.

AhmedS343:
@sbillinge Sorry about the delay. Should I update schemas.py to reflect what is in the google doc?

regolith/helpers/a_expensehelper.py
```diff
@@ -14,42 +14,45 @@
 )
 
 TARGET_COLL = "projecta"
-ALLOWED_TYPES = ["nsf", "doe", "other"]
-ALLOWED_STATI = ["proposed", "started", "finished", "back_burner", "paused",
-                 "cancelled"]
-MILESTONES_ALLOWED_STATI = ["proposed", "scheduled", "finished", "cancelled"]
+ALLOWED_TYPES = ["business", "travel"] # need to check all expense types.
+ALLOWED_STATI = ["submitted", "unsubmitted"]
```
sbillinge, line 18: please add "reimbursed"

regolith/helpers/a_expensehelper.py
```diff
@@ -14,42 +14,45 @@
...
-    subpi.add_argument("--date",
-                       help="The begin_date for the projectum  Defaults to "
-                            "today's date."
+    subpi.add_argument("name", help="A short name for the expense",
+                       default=None
                        )
-    subpi.add_argument("-d", "--description",
-                       help="Slightly longer description of the projectum"
+    subpi.add_argument("-b", "--business",
+                       help="expense type is business"
```
sbillinge, line 28: let user know that if -b not specified it will assume it is a travel expense.

Also, you will have to add code that sets business to True if -b is set.

regolith/helpers/a_expensehelper.py
```diff
@@ -14,42 +14,45 @@
...
                        )
-    subpi.add_argument("-c", "--collaborators", nargs="+",
-                       help="list of outside collaborators who should  be in contacts"
-                            "collection"
+    subpi.add_argument("-t", "--travel",
```
sbillinge, line 30: not needed, see above

regolith/helpers/a_expensehelper.py
```diff
@@ -14,42 +14,45 @@
...
+                       help="expense type is travel"
                        )
-    subpi.add_argument("-m", "--group_members", nargs="+",
-                       help="list of group members other than the lead who are involved"
+    subpi.add_argument("-y", "--payee",
+                       help="payee of the expense, defaults to sbillinge"
                        )
     subpi.add_argument("-g", "--grants", nargs="+",
-                       help="grant or (occasionally) list of grants that support this work"
+                       help="grant or (occasionally) list of grants that cover this expense"
```
sbillinge, line 37: remove "(occasionally)"

regolith/helpers/a_expensehelper.py
```diff
@@ -14,42 +14,45 @@
...
                        )
-    subpi.add_argument("-u", "--due_date",
-                       help="proposed due date for the deliverable"
+    subpi.add_argument("-s", "--status",
+                       help=f"status, from {ALLOWED_STATI}. default is unsubmitted"
+                       )
+    subpi.add_argument("-z", "--segregated",
+                       help="Amount for the segregated expense, defaults to 0"
```
sbillinge, line 43: "Amount for any segregated expense, defaults to 0"

regolith/helpers/a_expensehelper.py
```diff
@@ -14,42 +14,45 @@
...
+                       )
+    subpi.add_argument("-w", "--where",
+                       help="Where the expense has been submitted"
+                       )
+    subpi.add_argument("-n", "--notes", nargs="+",
+                       help="list of notes for the expense, defaults to empty list"
+                       )
+    subpi.add_argument("-d", "--begin",
```
sbillinge, line 51: begin_date and end_date (to conform to standard).  Let users know the format YYYY-MM-DD.

regolith/helpers/a_expensehelper.py
```diff
@@ -14,42 +14,45 @@
...
+                       help="Input begin date for this expense. Defaults to today's date"
+                       )
+    subpi.add_arguemnt("-e,", "--end",
+                       help="Input end date for this expense. Defaults to today's date"
```
sbillinge, line 55: if you have a default, set it here and then remove if logic later

tests/test_helpers.py
```diff
@@ -17,6 +17,11 @@
       "--date", "2020-04-29", "-c", "afriend", "-d", "more work",
       "-m", "ascopatz", "-g", "SymPy-1.1", "-u", "2021-01-01"],
      "20ly_newprojectum has been added in projecta\n"),
+    (["helper", "a_expense", "159.18", "Payment to google", "-b", "-g", "dmref15", "mrsec14"],
+     "Payment to google has been added to expenses\n"),
+    (["helper", "a_expense", "159.18", "Payment to google", "-b", "-y",
```
sbillinge, line 22: I don't think you can currently run multiple tests with the same output name

AhmedS343: Thanks. I will update my tests.

regolith/helpers/a_expensehelper.py
```diff
@@ -25,34 +25,33 @@ def subparser(subpi):
...
-    subpi.add_argument("-t", "--travel",
-                       help="expense type is travel"
+                       help="expense type is business. If not specified defaults to travel"
                        )
     subpi.add_argument("-y", "--payee",
                        help="payee of the expense, defaults to sbillinge"
                        )
     subpi.add_argument("-g", "--grants", nargs="+",
-                       help="grant or (occasionally) list of grants that cover this expense"
+                       help="list of grants that cover this expense"
```
sbillinge, line 34: sorry, I may have mistyped...."grant, or list of grants, that cover this expense"   Let's have  default of "tbd"

AhmedS343: done

tests/outputs/a_expense/expenses.yaml
```diff
@@ -0,0 +1,25 @@
+2003sb_google:
+	expense_type: business
```
sbillinge, line 2: the indentation is almost certainly wrong here.  Make sure that this matches the actual output.

Also the travel expense is much larger so please could you build the test for that?  Please look in my new_presentation.py script for this (you can basically copy-paste it so it shouldn't take too long). Let me know if you need that.

regolith/helpers/a_expensehelper.py
```diff
@@ -0,0 +1,184 @@
...
+ALLOWED_STATI = ["submitted", "unsubmitted", "reimbursed"]
+
+
+def subparser(subpi):
+    subpi.add_argument("amount", help="expense amount",
+                       )
+    subpi.add_argument("name", help="A short name for the expense",
+                       default=None
+                       )
+    subpi.add_argument("-b", "--business",
```
sbillinge, line 27: you need an ``action='store_true'``.  Please see https://docs.python.org/2/library/argparse.html

AhmedS343: Thank you. I will fix that

tests/outputs/a_expense/expenses.yaml
AhmedS343: Thank you, i will add the test for it.

AhmedS343: I added an example of a travel expense present on the slack channel.

AhmedS343:
@sbillinge I updated my updater, but I just noticed that there seems to be a difference between the format of the yaml on the projecta log vs the slack channel. I used the slack channel's format, but please let me know if this needs to be changed. I will also make changes to itemized expenses (and possibly add even more arguments) to reflect the multiple purposes possible.

regolith/helpers/a_expensehelper.py
```diff
@@ -103,81 +118,61 @@ def db_updater(self):
...
-            'log_url': '',
-            'name': rc.name,
-            'pi_id': rc.pi_id,
-            'lead': rc.lead,
-        })
-        if rc.lead is "tbd":
-            pdoc.update({
-                'status': 'proposed'
-            })
+            "begin_year": int(begin_date.split('-')[0]),
```
sbillinge, line 123: just load the dates, don't split (this is a new pattern that it can be in date form or day/month/year form)

regolith/helpers/a_expensehelper.py
```diff
@@ -84,16 +87,28 @@ def construct_global_ctx(self):
...
+            end_date = dt.date.today()
+            end_date = end_date.strftime('%Y-%m-%d')
         else:
-            due_date = date_parser.parse(rc.due_date).date()
-        key = f"{str(now.year)[2:]}{rc.lead[:2]}_{''.join(rc.name.casefold().split()).strip()}"
+            end_date = rc.end_date
+
+        # defaults to sbillinge
+        if rc.payee:
+            key = f"{str(begin_date.year)[2:]}{rc.payee[:2]}_{''.join(rc.name.casefold().split()).strip()}"
```
sbillinge, line 107: pls check that this is the pattern for the key in expenses.  I am guessing not.

regolith/helpers/a_expensehelper.py
```diff
@@ -84,16 +87,28 @@ def construct_global_ctx(self):
...
+            begin_date = dt.date.today() # string with date time format
+            begin_date = begin_date.strftime('%Y-%m-%d')
         else:
-            now = date_parser.parse(rc.date).date()
-        if not rc.due_date:
-            due_date = now + relativedelta(years=1)
+            begin_date = rc.begin_date() # need to raise exception for wrongly formatted user input
+        if not rc.end_date:
+            end_date = dt.date.today()
+            end_date = end_date.strftime('%Y-%m-%d')
```
sbillinge, line 101: I think it is better to leave it as a datetime object.  yml can save datetime and date is in the schema as an allowed type

regolith/helpers/a_expensehelper.py
```diff
@@ -103,81 +118,61 @@ def db_updater(self):
...
-            pdoc.update({'grants': rc.grants})
+            percentages = [100 / len(rc.grants) for i in rc.grants]
         else:
-            pdoc.update({'grants': ["tbd"]})
-        if rc.group_members:
-            if isinstance(rc.group_members, str):
-                rc.group_members = [rc.group_members]
-            pdoc.update({'group_members': rc.group_members})
+            rc.grants = 'tbd'
+            percentages = 0 #No grant => no percentage?
```
sbillinge, line 143: I think this has to be a list.  We could set it to 100 if there is not grant but check, it might have to be [100]

regolith/helpers/a_expensehelper.py
```diff
@@ -103,81 +118,61 @@ def db_updater(self):
...
+
+        # segregated expense
+        if not rc.segregated:
+            rc.segregated = "0"
```
sbillinge, line 147: I think this should be an int not str

sbillinge:
looks like you need a rebase too.

regolith/helpers/a_expensehelper.py
AhmedS343: I will leave it as a list, and set grant percentage to 100 if there is no grant.

regolith/helpers/a_expensehelper.py
AhmedS343: I will convert this to an int, (I was converting it to an int at a later stage).

regolith/helpers/a_expensehelper.py
AhmedS343: Ok, thanks for the pointer.

regolith/helpers/a_expensehelper.py
AhmedS343: I am actually not sure what defines the name of an expense. I am guessing a user specifies the name, but there are names like 2003sb_google, where does the 2003 come from?

regolith/helpers/a_expensehelper.py
AhmedS343: done

tests/test_helpers.py
```diff
@@ -21,10 +21,8 @@
       "-a","Kurt Godel", "MC Escher", "Johann Sebastian Bach","-c", "Bitcoin",
       "--other_agencies", "Flatland","-n", "this is a sample added proposal"],
      "20_anewproposal has been added in proposals\n"),
-    (["helper", "a_expense", "159.18", "Payment to google", "-b", "-g", "dmref15", "mrsec14"],
-     "Payment to google has been added to expenses\n"),
-    (["helper", "a_expense", "159.18", "Payment to google", "-b", "-y",
-      "ashaaban", "-g", "dmref15", "mrsec14"],
+    (["helper", "a_expense", "159.18", "google", "Payment to Google Cloud to pay for cloud services",
+      "-b", "g", "dmref15", "mrsec14", "-d", "2020-06-18"],
```
sbillinge, line 25: I think this should be -g.  Also, to get more test coverage, add something for all the optional variables and write the test for the travel case, which is more complicated.

tests/outputs/a_expense/expenses.yaml
```diff
@@ -1,95 +1,103 @@
...
-		  year: 2020
-		  purpose: Buy the train ticket (Ronkonkoma LIRR station to Penn station)
-		  unsegregated_expense: 14.25
-		  segregated_expense: 0.0
-	notes: []
-	overall_purpose: Travel to the Brookhaven National Laboratory (BNL) to conduct x-ray
-		diffraction experiments
-	payee: sstao
-	status: submitted
+		- day: 1
```
sbillinge, line 26: these seem to have values entered.    Also, please use dates.  Please could you maybe get together with someone to work on getting the tests as good as they can be?

tests/test_helpers.py
AhmedS343: done

regolith/helpers/a_expensehelper.py
sbillinge: It may be better to specify a default in the parser, then you don't need this if statement.

regolith/helpers/a_expensehelper.py
sbillinge: this is a moving target.  We never defined a heuristic for this, so you can come up with one, but my sense is that it is more informational if it has info about the expense rather than just the name of the payee.  maybe something like ``2006sb_<sthg>`` as an example.

regolith/helpers/a_expensehelper.py
```diff
@@ -0,0 +1,179 @@
...
+    subpi.add_argument("-z", "--segregated",
+                       help="Amount for any segregated expense. Defaults to 0"
+                       )
+    subpi.add_argument("-w", "--where",
+                       help="Where the expense has been submitted"
+                       )
+    subpi.add_argument("-n", "--notes", nargs="+",
+                       help="List of notes for the expense. Defaults to empty list"
+                       )
+    subpi.add_argument("-d", "--begin_date",
```
sbillinge, line 51: I see the defaults mentioned in the help, but not any defaults set here.   Set them here if you can (check the argparse docs how to do it)

regolith/helpers/a_expensehelper.py
sbillinge: your comment says done but I don't see it done?  Something not pushed?

tests/outputs/a_expense/expenses.yaml
AhmedS343: Thank you. I will reach out to Daniela

AhmedS343: I will reach out to Daniela.

regolith/helpers/a_expensehelper.py
AhmedS343: I added a default, but I am not sure if should just leave it as a datetime object, or a string of today's date.

regolith/helpers/a_expensehelper.py
AhmedS343: For cases like sstao, from what I see an expense name would be 20st_name, not 20ss_name.

regolith/helpers/a_expensehelper.py
AhmedS343: done

regolith/helpers/a_expensehelper.py
sbillinge: I'm not sure, but it could be a fun project for you to work on.  It crops up elsewhere in the project.  don't waste too much time on it though, it is not that big of a deal, especially in expenses.

btw, please make sure you go through and respond to all my comments.  That way I know that you saw them.

regolith/helpers/a_expensehelper.py
```diff
@@ -0,0 +1,176 @@
...
+                       )
+    subpi.add_argument("-g", "--grants", nargs="+",
+                       help="grant, or list of grants that cover this expense. Defaults to tbd"
+                       )
+    subpi.add_argument("-s", "--status",
+                       help=f"status, from {ALLOWED_STATI}. Default is unsubmitted"
+                       )
+    subpi.add_argument("-z", "--segregated",
+                       help="Amount for any segregated expense. Defaults to 0",
+                       default="0"
```
sbillinge, line 45: is it allowed to be an int or it has to be a string?

btw, I think I meant for you to take the next step after I hinted the first step, and to think about defaults for all the optional args where it makes sense.....

regolith/helpers/a_expensehelper.py
sbillinge: date object, if it is allowed.

regolith/helpers/a_expensehelper.py
AhmedS343: Done

regolith/helpers/a_expensehelper.py
AhmedS343: Done

regolith/helpers/a_expensehelper.py
AhmedS343: Ok, for now I think it would be simpler to just ignore these exceptions.

regolith/helpers/a_expensehelper.py
AhmedS343: done

regolith/helpers/a_expensehelper.py
AhmedS343: Okay, I see now. I will fix this, before proceeding with the helper I am going to try, and make sure I have my test for travel expenses up.

regolith/helpers/a_expensehelper.py
sbillinge: cool.  Do me a favor though and throw up that job as an issue.....

we probably want to have a function in tools that builds the name-based part of the ID and not have the code that does it in every helper.  there are two patterns, YYfl and YYMMfl and the function could do both.  Don't make the function, just make the issue and let me know when it is done.

regolith/helpers/a_expensehelper.py
sbillinge: yes, perfect!  TDD.  You are getting it!  Thanks so much

regolith/helpers/a_expensehelper.py
AhmedS343: created new issue for the function.

regolith/helpers/a_expensehelper.py
AhmedS343: done

tests/test_helpers.py
```diff
@@ -21,9 +21,10 @@
       "-a","Kurt Godel", "MC Escher", "Johann Sebastian Bach","-c", "Bitcoin",
       "--other_agencies", "Flatland","-n", "this is a sample added proposal"],
      "20_anewproposal has been added in proposals\n"),
-    (["helper", "a_expense", "159.18", "google", "Payment to Google Cloud to pay for cloud services",
-      "-b", "g", "dmref15", "mrsec14", "-d", "2020-06-18"],
-     "Payment to google has been added to expenses\n"),
+    (["helper", "a_expense", "159.18", "google", "Payment to google cloud",
```
sbillinge, line 24: let's make it a bit more realistic, since these exemplars also form part of the documentation for users.  Since this is a travel expense, something like "Invited talk in timbuktoo" would be better.

tests/test_helpers.py
```diff
@@ -21,9 +21,10 @@
...
+      "-g", "mrsec14", "dmref15", "-y", "ashaaban", "-s", "submitted",
```
sbillinge, line 25: also, to make it more readable as documentation, let's use the full names of the variables here rather than the shortcuts, so ``--status`` rather than ``-s`` etc.

AhmedS343: done

tests/test_helpers.py
AhmedS343: done

tests/test_helpers.py
```diff
@@ -21,10 +21,10 @@
...
       "--other_agencies", "Flatland","-n", "this is a sample added proposal"],
      "20_anewproposal has been added in proposals\n"),
-    (["helper", "a_expense", "159.18", "google", "Payment to google cloud",
-      "-g", "mrsec14", "dmref15", "-y", "ashaaban", "-s", "submitted",
-      "-w", "bank", "-d", "2020-06-20", "-e", "2020-06-25"],
-     "2006as_google has been added in expense\n"),
+    (["helper", "a_expense", "159.18", "timbuktoo", "travel to timbuktoo",
+      "--grants", "mrsec14", "dmref15", "--payee", "ashaaban", "--status", "submitted",
+      "--where", "bank", "--begin_date", "2020-06-20", "--end_date", "2020-06-25"],
+     "2006as_timbuktoo has been added in expense\n"),
```
sbillinge, line 27: added in expenses (plural)

AhmedS343: done.

AhmedS343:
@sbillinge. My test is not passing because the actual output of my helper for date (in itemized) is:
date: *id002. I am not sure what is causing this.

regolith/helpers/a_expensehelper.py
```diff
@@ -11,9 +11,11 @@
 from regolith.tools import (
     all_docs_from_collection,
     get_pi_id,
+    get_dates
+
```
sbillinge, line 15: remove empty line

regolith/helpers/a_expensehelper.py
```diff
@@ -93,84 +89,148 @@ def construct_global_ctx(self):
...
+                    'unsegregated_expense': 0,
+                    'segregated_expense': 0
+                }]
+            })
+        pdoc.update({'notes': rc.notes,
+                     'overall_purpose': rc.purpose,
+                     'payee': rc.payee,
+                     })
+        if rc.status == 'submitted':
+            submission_day = begin_date.day
```
sbillinge, line 202: we are moving to dates rather than year month day so maybe if we are updating we should update with ``submission_date``.  It may need a schema update.

regolith/helpers/a_expensehelper.py
```diff
@@ -93,84 +89,148 @@ def construct_global_ctx(self):
...
+            submission_month ='tbd'
+            submission_year = begin_date.year
 
 
+        pdoc.update({'reimbursements': [
+            {
+                'amount': 0,
+                'day': 0,
+                'month': 'tbd',
+                'submission_day': submission_day,
```
sbillinge, line 216: same here...move to date

tests/outputs/a_expense/expenses.yaml
```diff
@@ -43,100 +49,81 @@
...
-
-
-
-
-
-
-
-
+  project: Cyclus
+
```
sbillinge, line 129: clean up end of file.  Not sure what the problem is here....

AhmedS343: done

regolith/helpers/a_expensehelper.py
AhmedS343: done

regolith/helpers/a_expensehelper.py
AhmedS343: done

regolith/helpers/a_expensehelper.py
AhmedS343: done

AhmedS343:
@sbillinge I am having problems fixing a particular issue if I place my dates in a list (eg for itemized expenses) it messes up the date representation and instead of getting something like date: 2020-06-22. I get something like: date: *id001. I am not sure what causes this behaviour but deepcopying the objects seems to fix this, but this is not really a good thing to do.

Also, for segregated expenses, I am not sure what behaviour the helper should follow for travel expenses. So, forexample, if a user specifies the value for a segregated_expense what am I mean to do with this since there are many dicts in the list.

sbillinge:
> 
> 
> @sbillinge I am having problems fixing a particular issue if I place my dates in a list (eg for itemized expenses) it messes up the date representation and instead of getting something like date: 2020-06-22. I get something like: date: *id001. I am not sure what causes this behaviour but deepcopying the objects seems to fix this, but this is not really a good thing to do.
> 
> Also, for segregated expenses, I am not sure what behaviour the helper should follow for travel expenses. So, forexample, if a user specifies the value for a segregated_expense what am I mean to do with this since there are many dicts in the list.

please crowdsource in the group for the first thing.  For the second thing, it should put the segregated and unsegregated expense in the right place, of course.  But is this an adder or an updater?  If it is an adder, just remove that as an option for the command line and put in 0 for all the values.  Adders just add emtpy documents.

AhmedS343:
@sbillinge  By crowdsource do you mean post the problem on the slack channel? I have also removed the argument for segregated expense.

AhmedS343:
@sbillinge Ready for final review.

regolith/helpers/a_expensehelper.py
```diff
@@ -87,11 +87,10 @@ def db_updater(self):
         # dates
         if rc.begin_date:
             begin_date = date_parser.parse(rc.begin_date).date()
-            date = date_parser.parse(rc.begin_date).date().strftime('%Y-%m-%d')
+            date_string = rc.begin_date
         else:
             begin_date = dt.date.today()
-            date = begin_date
-            date = dt.date.today().strftime('%Y-%m-%d')
+            date_string = dt.date.today().strftime('%Y-%m-%d')
```
sbillinge, line 93: why are you turning a date into a string, then back into a date later?  Just keep everything as date type.  

Also, don't change begin_date to date.  These two things have different meanings so keep it as date.  I think this should work:
```
        if rc.begin_date:
            begin_date = date_parser.parse(rc.begin_date).date()
        else:
            begin_date = dt.date.today()

File: regolith/helpers/a_expensehelper.py [hunk 29]
```diff
@@ -129,49 +128,49 @@ def db_updater(self):
         if expense_type == 'travel':
             pdoc.update({'itemized_expenses': [
                 {
-                    'date': date,
+                    'date': date_parser.parse(date_string).date(),
```
sbillinge, line 131: ...then here youshould just need ``'date': begin_date``

AhmedS343: This is the strange behaviour I was referring to, if I use this line the value for date becomes date: *id001. For some reason, putting the date object in a list more than once results in this behaviour. The fix I had in mind was to simply make a date object for each key, which is why I use the string since it allows me to call the constructor.

sbillinge:
> @sbillinge By crowdsource do you mean post the problem on the slack channel? I have also removed the argument for segregated expense.

I mean "ask someone in the group" and if they can't help "ask someone else...."

regolith/helpers/a_expensehelper.py
sbillinge: I see.  To avoid this we may have to modify the ``dump_yaml`` in fsclient to ignore aliases (see https://stackoverflow.com/questions/58091449/disabling-alias-for-yaml-file-in-python).  Why don't you try this trick on your branch and see if it works.

AhmedS343: Thank you! It does work.

regolith/helpers/a_expensehelper.py
AhmedS343: done

regolith/helpers/a_expensehelper.py
sbillinge: cool.  Please could you also include in "news" that we made this change to the yaml dumper.  This is a better fix because it will remove this problem from wherever dump is used.....

sbillinge:
let me know when it is ready for merge (and don't forget to remove the WIP.  I think you need a news for the helper too.

AhmedS343:
@sbillinge ready to merge.

sbillinge:
Great Job @AhmedS343 this looks great!   I will merge and start using it.  I will post any issues (which always come up when I start using these....)

PR #418, opened 2020-06-09
Fixed schema issue

shogho0:


regolith/schemas.py
```diff
@@ -1169,12 +1169,12 @@
...
             "description": "the month when the entry was created",
             "required": False,
-            "type": "integer",
+            "anyof_type": ["integer", "string"],
         },
         "year": {
             "description": "the year when the entry was created",
             "required": False,
-            "type": "integer",
+            "anyof_type": ["integer", "string"],
```
sbillinge, line 1177: this should be integer only.  Months can be "May" or 5 but years can only be 2020

shogho0: Done

shogho0:
@sbillinge This is ready to be merged to fix the circleci error

regolith/schemas.py
```diff
@@ -1169,12 +1169,12 @@
...
             "description": "the month when the entry was created",
             "required": False,
-            "type": "integer",
+            "anyof_type": ["integer", "string"],
         },
         "year": {
             "description": "the year when the entry was created",
             "required": False,
-            "type": "integer",
+            "anyof_type": ["integer"],
```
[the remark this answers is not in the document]
shogho0: Sorry about that-to clarify, I actually thought about changing it back to type:"integer", but then thought the anyof_type would be better as it could be more easily changed in the future. I'll make the appropriate change now.

shogho0:
Just pushed fix

sbillinge:
thanks @shogho0 !

PR #420, opened 2020-06-09
Edit to date logic in contactlisterhelper

shogho0:
I was doing some more stringent testing on the helper following the schema change this morning, and found a slight error in the dates logic. I've corrected that error now.


regolith/helpers/l_contactshelper.py
```diff
@@ -108,19 +108,24 @@ def sout(self):
...
-                                       dateutil.relativedelta.relativedelta(
-                                           months=int(
-                                               rc.range))).isoformat(),
-                        "end_date": (temp_dat +
-                                     dateutil.relativedelta.relativedelta(
-                                         months=int(
-                                             rc.range))).isoformat()}
-                    if is_current(temp_dict, now=temp_dat):
-                        date_list.append(stringify(contact))
+                if contact.get('day'):
```
sbillinge, line 121: this should be handled by ``get_dates()``.  If not, it may be a bug in get_dates().  Can you give more info about what is failing when you use get_dates?

shogho0: The error wasn't in using get_dates, it was that I was accidentally using rc.date as my "current date" to compare to rather than the actual date from the contact. I just uploaded a new commit that revises my code and uses get_dates().

sbillinge:
ouch, I am afraid that a recent merge means you will have to do a rebase.  so sorry.

PR #427, opened 2020-06-11
Fixed contacts helper bug

shogho0:
Solves @426

### DIFF
--- a/regolith/helpers/l_contactshelper.py
+++ b/regolith/helpers/l_contactshelper.py
@@ -23,6 +23,11 @@ HELPER_TARGET = "l_contacts"
 
 def subparser(subpi):
     subpi.add_argument(
+        "-u",
+        "--run",
+        required=True,
+        help='required argument: "y" must be entered here in order for the lister to run')    
+    subpi.add_argument(
         "-n",
         "--name",
         help='name or name fragment (single argument only) to use to find contacts')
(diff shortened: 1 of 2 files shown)


regolith/helpers/l_contactshelper.py
```diff
@@ -22,6 +22,11 @@
 
 
 def subparser(subpi):
+    subpi.add_argument(
```
sbillinge, line 25: I think it would be better if this is like  cmd not a flag... so one would type:
```
regolith helper l_contacts run
```

regolith/helpers/l_contactshelper.py
```diff
@@ -86,50 +91,51 @@ def construct_global_ctx(self):
...
-                curr_d = get_dates(contact)['date']
-                if is_current(temp_dict, now=curr_d):
-                    date_list.append(stringify(contact))
-            datel = set(date_list)
-        else:
-            datel = def_l
-        res_l = set.intersection(namel, instl, notel, datel)
-        for item in res_l:
-            print(item)
+        if rc.run == "y":
```
sbillinge, line 94: Doing it the way I suggest above I think you basically don't need any of this ``if`` logic.  If run is not typed the CLI will default to the help message and if it is typed it will continue

regolith/helpers/l_contactshelper.py
shogho0: Done

regolith/helpers/l_contactshelper.py
shogho0: Done

regolith/helpers/l_contactshelper.py
```diff
@@ -23,10 +21,8 @@
 
 def subparser(subpi):
     subpi.add_argument(
-        "-u",
-        "--run",
-        required=True,
-        help='required argument: "y" must be entered here in order for the lister to run')    
+        "run",
+        help='required argument that must be entered here in order for the contacts lister to run')
```
sbillinge, line 25: how about ``"run the lister.  To see allowed optional arguments type regolith helper l_contacts"``

shogho0:
Going to upload additional commit that resolves the test issue

sbillinge:
yipee....we love tests!

PR #428, opened 2020-06-12
WIP: filesystem to mongo

st3107:
Add functionality: transfer the database from fielsystem to mongodb

### DIFF
--- a/regolith/commands.py
+++ b/regolith/commands.py
@@ -7,9 +7,12 @@ from pprint import pprint
 
 from regolith.builder import builder, BUILDERS
 from regolith.deploy import deploy as dploy
+from regolith.emailer import emailer
 from regolith.helper import HELPERS, helpr
 from regolith.runcontrol import RunControl
 from regolith.tools import string_types
+
+email = emailer
 
 RE_AND = re.compile(r"\s+and\s+")
 RE_SPACE = re.compile(r"\s+")
--- a/regolith/commands.py
+++ b/regolith/commands.py
@@ -186,7 +186,7 @@ def yaml_to_json(rc):
 
 
 def fs_to_mongo(rc: RunControl) -> None:
-    """Import database from filesystem to mongo db.
+    """Convert database collection from filesystem to mongo db.
 
     Parameters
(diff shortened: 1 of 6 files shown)


st3107:
@sbillinge Ready for review. I am not sure why the codecov thought the coverage decreased. I add the test for the fs-to-mongo. This should cover almost all the code I added.

I cannot see the report of codecov so that I cannot know where to improve. I am not sure if someone could authorize me of the codecov.

regolith/commands.py
```diff
@@ -180,6 +185,23 @@ def yaml_to_json(rc):
         fsclient.yaml_to_json(inp, out)
 
 
+def fs_to_mongo(rc: RunControl) -> None:
+    """Import database from filesystem to mongo db.
```
sbillinge, line 189: I wonder if this is clearer if it says "Convert database collection from filesystem to mongo db" rather than import

st3107:
Done. Ready to merge.

sbillinge:
@scopatz @CJ-Wright This is some basic infrastructure for regolith so I would rather that you look over it quickly.  We are playing around with moving some of the databases to a mongo backend and Songsheng (@st3107 ) has written some code that should help with that.  I am happy to merge it, but would love for either or both of you to look it over quickly and make sure it is ok, or make suggestions.   Thanks

sbillinge:
@CJ-Wright @scopatz any thoughts?

st3107:
I need to do some further enhancement. There is a bug when trying to import database to mongod atlas

st3107:
Done. Ready for review.

sbillinge:
thanks @st3107 

I think that tools is causing coverage to fail.  Maybe we need a  quick hackathon to get more coverage in tools....

st3107:
I am not sure about the mechanism in calculation of coverage. I can make sure all the functions I add have been tested.

sbillinge:
yes on tests for functions, but we need to add tests for some functions that got merged before without tests.

I think one test is that coverage  >65% or so and another test on the change in coverage.

PR #436, opened 2020-06-15
Fixing list_projecta

### LINKED ISSUE #434: BUG: cannot trigger help message in l_projecta
**desired behavior:** 
have a way to return the list of allowed flags and the usage

**current behavior:**
if unknown flags are given it just lists all projecta and not a usage message

There is a workaround that shomik used in the l_contact helper and we could reuse that here (require a fake ``run`` command).  If you can think of a better way that would be better though.

### LINKED ISSUE #435: ENH: l_projecta
Would like a ``-g`` ``--grant`` filter that filters for a grant_id (e.g., efrc18) and returns all projecta associated with that grant.

mwinitch:
Closes #434 and closes #435


mwinitch:
@sbillinge As far as fixing #434 I have my given my solution. I am not sure if its what you are looking for. If you run `regolith helper l_projecta` or `regolith helper l_projeca -fakeflag` then only a help message will be printed. If you want all grants to be printed just run `regolith helper l_projecta -v` and it will print all of them. This does not require a run statement.

regolith/helpers/l_projectahelper.py
```diff
@@ -45,6 +45,9 @@ def subparser(subpi):
                        help="date range back from DATE to search over in days. If no "
                             "range is specified, search will be 7 days"
                        )
+    subpi.add_argument("-g", "--grant",
+                       help="Filter projecta by by a grant ID"
```
sbillinge, line 49: extra "by"

mwinitch:
So I can remove the news?

mwinitch:
@sbillinge This is ready for review and merge

regolith/helpers/l_projectahelper.py
```diff
@@ -86,6 +89,8 @@ def construct_global_ctx(self):
 
     def sout(self):
         rc = self.rc
+        if (not rc.lead) and (not rc.person) and (not rc.ended) and (not rc.grant) and (not rc.verbose):
```
sbillinge, line 92: wow, that's nice, I never saw that trick before!  fantastic!

PR #438, opened 2020-06-15
Fix get_dates function

### LINKED ISSUE #437: get_dates issues with date objects

mwinitch:
Closes #437

### DIFF
--- a/news/str_date.rst
+++ b/news/str_date.rst
@@ -1,5 +1,4 @@
-**Added:**
-    * get_dates function that now works with objects with datetime objects
+**Added:** None
 
 **Changed:** None
 
@@ -7,6 +6,7 @@
 
 **Removed:** None
 
-**Fixed:** None
+**Fixed:**
+    * get_dates function works with datetime objects and strings
 
 **Security:** None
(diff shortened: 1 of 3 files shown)


mwinitch:
@sbillinge This is ready for review. The get_dates function can have issues because when you do `thing.get('begin_date)` it will sometimes return a datetime object instead of a string. However, especially when it is getting dates from the schemas, it does return a valid string. This fixes the function so it works regardless if `.get` returns a datetime object or a string that has to be converted to a datetime object.

mwinitch:
Yes I added some tests for it. Just wondering how come this needs news but the two other fixes I did for l_projecta did not news?

mwinitch:
@sbillinge This is ready for review

news/str_date.rst
```diff
@@ -0,0 +1,12 @@
+**Added:**
+    * get_dates function that now works with objects with datetime objects
```
sbillinge, line 2: I think you didn't add a get_dates function, you fixed it.... (see below)

mwinitch: Fixed

mwinitch:
@sbillinge News is fixed and it is ready for merge

sbillinge:
Thanks Michael, this is really awesome!

PR #446, opened 2020-06-17
is fully appointed function

adibaejaz:
function in tools.py to check if a list of appointments for a person is fully loaded for a given interval of time


tests/test_tools.py
```diff
@@ -665,3 +666,121 @@ def test_fragment_retrieval(input, expected):
 def test_month_and_year(input,expected):
     assert(month_and_year(input[0],input[1]) == expected) 
 
+@pytest.mark.parametrize(
```
sbillinge, line 669: nice!   I would test using dates mostly (i.e., 2020-06-17) and just one test that uses day/month/year to test that pattern for the sake of backwards compatibility, as moving forward we are moving to dates which are easier to keep track of in the database.

regolith/tools.py
```diff
@@ -1287,4 +1289,31 @@ def get_id_from_name(coll, name):
...
     if person:
         return person["_id"]
-    else: return None
\ No newline at end of file
+    else: return None
+
+def is_fully_appointed(appts, begin_date, end_date):
+    status = [True, '']
+    if isinstance(begin_date,str):
+         begin_date = date_parser.parse(begin_date).date()
```
sbillinge, line 1329: indentation problem

regolith/tools.py
```diff
@@ -1287,4 +1289,31 @@ def get_id_from_name(coll, name):
                              case_sensitive=False)
     if person:
         return person["_id"]
-    else: return None
\ No newline at end of file
+    else: return None
+
+def is_fully_appointed(appts, begin_date, end_date):
```
sbillinge, line 1294: do we want to pass a person rather than a list of appointments?  Grammatically, ``is_fully_appointed(aejaz)`` seems to make more sense to me.

then logically it should return True or False.

I could imagine it taking dates as optional args.  default values for these could be the begin_date of the student (not the appointment) and end_date if it exists with now as default?  We may need a ``supported: True`` flag in the ``employment`` item to get meaningful begin and end dates for the student.   Maybe too brittle, I am not sure.  Easier would be to take as default begin and end dates as the earliest and the latest date in the list of appointments.

sbillinge: I like that your function identifies when the the person is over/under appointed.  Can we simply print this to stdOut as a warning rather than returning it as a value?

tests/test_tools.py
adibaejaz: done

regolith/tools.py
adibaejaz: done

regolith/tools.py
adibaejaz: I have changed it to take in a `person` and added default behaviour for the dates (earliest and latest appointment date)
As for the second thing - I think that is more elegant, but the problem is that it then prints a min every time the function is called in the helper file, which is a problem because then we have to move the `print(appointment)` into the loop, and the stdOut will display
`min 1.0 at 2012-12-32`
`aejaz f21`
i.e. the appointment below the minimum and not above, which is strange to read... and maybe some other problems I can't anticipate. do you have any suggestions for how to handle this?

sbillinge: not specifically but isn't that the behavior that I want?

From my perspective, it is helpful to know which appointments need to be adjusted because of an error in loading or grant etc. so if I get the person_id and the appointment_id of the bad appointment I can go and find it.    I guess if there is a gap in appointments, knowing the appointment_id of the preceding and following appointments would help me zoom into it.

sbillinge: ``min 1.0 at 2012-12-31`` is not a very helpful error message either tbn....

adibaejaz: I see what you mean: I returned a list earlier so that my code would display something like
`person: MC Escher       appointment: f20`
`min 0.4 at 2020-09-01`
`person: Kurt Godel       appointment: ss20`
`max 1.3 at 2020-09-01`
And changing the function to return True/False and print the message will display
`min 0.4 at 2020-09-01`
`person: MC Escher       appointment: f20`
`max 1.3 at 2020-09-01`
`person: Kurt Godel       appointment: ss20`
So the order of the display changes since the function is called before the appointment is printed.
I will make the error message more verbose so hopefully that's more helpful!

adibaejaz:
@sbillinge not sure what it means that this is failing codecov/project

regolith/tools.py
adibaejaz: done! I have manipulated it to print both on the same line

regolith/tools.py
```diff
@@ -1335,33 +1335,32 @@ def is_fully_appointed(person, begin_date=None, end_date=None):
             earliest = dates['begin_date']
         if dates['end_date'] > latest:
             latest = dates['end_date']
-    if not begin_date:
-        begin_date = earliest
-    if not end_date:
-        end_date = latest
+    begin_date = earliest if not begin_date else begin_date
+    end_date = latest if not end_date else end_date
     if begin_date > end_date:
```
sbillinge, line 1340: nice!

adibaejaz: thank you!

adibaejaz:
@sbillinge thank you! before I do that, should I remove the `is_fully_loaded` function in tools.py? it's the one I optimised, and I think was coded to serve a similar if not the same purpose

sbillinge:
> @sbillinge thank you! before I do that, should I remove the `is_fully_loaded` function in tools.py? it's the one I optimised, and I think was coded to serve a similar if not the same purpose. apart from that, this is ready for a final review!

Yes, remove it if it is doing the same thing.  I guess I forgot I put it there, but I guess that yours has more test coverage and is more recent.  You will have to find references to it everywhere  in regolith and replace the code to work with your new function.  Then let me know and I can merge.

adibaejaz:
@sbillinge done!

regolith/tools.py
```diff
@@ -1282,9 +1245,82 @@ def fragment_retrieval(coll, fields, fragment, case_sensitive = False):
...
+        >>> aejaz = {"name": "Adiba Ejaz", "_id": "aejaz", "appointments": appts}
+        >>> is_fully_appointed(aejaz, "2017-06-01", "2017-06-30")
+
+        In this case, we have an insufficient loading and minimum of 0.0 on 2017-06-16, hence it would return False and
+        print "aejaz is short of full loading with minimim 0.0 on 2017-06-16".
+        """
+
+    status = True
+    appts = person.get('appointments')
+    earliest = get_dates(appts[0])['begin_date']
```
sbillinge, line 1290: a cleaner way of doing this might be something like:
```
dates = [(get_dates(appt).get('begin_date'), get_dates(appt).get('end_date'))  for appt in appts]
if not begin_date:
    begin_date = min(dates[0])
if not end_date:
    latest = max(dates[1])
```
However, this looks to me like "magic" that we should remove it.  begin and end date should be required arguments and the function should blow up if they are not given.   The "no magic" rule.

regolith/tools.py
```diff
@@ -1282,9 +1245,82 @@ def fragment_retrieval(coll, fields, fragment, case_sensitive = False):
...
+    end_date = latest if not end_date else end_date
+    if begin_date > end_date:
+        raise ValueError("invalid begin and end dates")
+    if isinstance(begin_date, str):
+        begin_date = date_parser.parse(begin_date).date()
+    if isinstance(end_date, str):
+        end_date = date_parser.parse(end_date).date()
+    timespan = end_date - begin_date
+    loading, max, min, help = [], 0.0, 1.0, ''
+    for x in range(0, timespan.days):
```
sbillinge, line 1308: fyi, you don't need the 0 here.  OK to have it, but personally I think it is more readable without tbh....

regolith/tools.py
```diff
@@ -1282,9 +1245,82 @@ def fragment_retrieval(coll, fields, fragment, case_sensitive = False):
...
+        loading.append(0.0)
+        day = begin_date + relativedelta(days=x)
+        for appt in appts:
+            if is_current(appt, now=day):
+                loading[x] = loading[x] + appt.get("loading")
```
sbillinge, line 1313: couldn't you everywhere in here just have loading as a scalar variable, i.e., ``day_loading = day_loading + appt.get("loading")`` then at the very end have ``loadings.append(loading)``?  

What behavior do we want for the function? Do we want it to report the first bad loading it finds?  In that case, as soon as we find a bad one we can print the ``"WARNING: {} exceeds full loading on {}".format(person.get('_id'), max, str(day))"`` or whatever.  Do we want it to find the one, worst, bad loading (which I think is what your version is doing). Or do we want it to print all the bad loadings?

From my perspective, the first is fine but the easiest to debug my appointments might be to have a list of all the bad appointments.  This could make a huge long list if there are many bad days, so more elegant would be logic that would find a day that is not fully appointed and print just once for that whole "appointment gap", but in the spirit of agile programming, just dumping a long list of bad days in print statements is quick and easy.    None of these UCs require you to build the loadings array and find its max I think, so the code in here could be much less (which means easier to understand and less to maintain)

regolith/tools.py
adibaejaz: Thank you for the logic, it's very elegant! I will keep it in mind for future code, and change these arguments to required

news/is_fully_appointed.rst
```diff
@@ -5,7 +5,8 @@
 
 **Deprecated:** None
 
-**Removed:** None
+**Removed:**
+* old is_fully_loaded function in tools.py and its tests in test_tools.py since the new function is an improved version that does the same thing
```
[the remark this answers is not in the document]
adibaejaz: done

regolith/tools.py
adibaejaz: done

regolith/tools.py
adibaejaz: Done for the loading array - it seems I didn't need it at all! I have also included functionality to print "appointment gaps" and tested this locally by returning the `messages` in a list and checking in my tests, but I have pushed a version which instead prints to stdout and returns  a bool! the different edge cases make the code a little more obtuse, but the helper messages are condensed and quite useful. let me know what you prefer!

regolith/tools.py
sbillinge: both should be required, just as you did in the code.  Remember, no magic.   You can spend half a lifetime looking for bugs generated by magic.

regolith/tools.py
sbillinge: I like your solution!   It will be easier for me to test when I try it.   As long as you are confident that it is working correctly I am good.

sbillinge:
Thanks @adibaejaz I will merge.   We can test more when I start using the appointments helper.

Issue #443, opened 2020-06-17
BUG: contact updater requires institution when updating


sbillinge:
desired behavior:
institution required when adding a new contact but not when updating

actual behavior:
inst is always required

dy2403:
I think it was solved by https://github.com/regro/regolith/pull/460 !

sbillinge:
Hi @dy2403 I tested and it seems to be so, thanks.

PR #447, opened 2020-06-17
Contact helper issues

### LINKED ISSUE #444: ENH: contact updater can take an id as an optional argument
When updating (rather than adding) it would be nice to be able to specify an id rather than first/last name only.

Come to think of it, it would be nice to be able to specify an id in any case, for example, with chinese names (and possibly other nationalities in the future) our "standard" behavior is to put two initials from the first name (so Songsheng Tao is sstao, not stao)

dy2403:
closes #444 and #443 

hi @sbillinge !
can you please review it?
I am trying to figure out why the error is happening with your empty list (issue 445). 
the problem seems to be linked to current, and not to notes though.

### DIFF
--- a/regolith/helpers/u_contacthelper.py
+++ b/regolith/helpers/u_contacthelper.py
@@ -16,12 +16,12 @@ TARGET_COLL = "contacts"
 
 def subparser(subpi):
     subpi.add_argument("name", help="first name space last name in quotes")
-    subpi.add_argument("--id", help="id of the person, e.g., first letter first name "
-                                    "plus last name, but unique")
-    subpi.add_argument("--inst", help="person's institution.  Can be inf "
-                                      "short form such as columbiau and will "
-                                      "be retrieved from institutions collection. "
-                                      "required to create a new contact")
+    subpi.add_argument("-d", "--id", help="id of the person, e.g., first letter first name "
+                                            "plus last name, but unique")
+    subpi.add_argument("-i", "--institution", help="person's institution. It can be "
+                                                   "institution id or anything in the "
+                                                   "aka or name from institutions collection. "
+                                                   "it is required to create a new contact")
     #FIXME
     # subpi.add_argument("-e", "--email",
     #                    help="email address")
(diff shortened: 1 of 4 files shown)


regolith/helpers/u_contacthelper.py
```diff
@@ -16,9 +16,12 @@
 
 def subparser(subpi):
     subpi.add_argument("name", help="first name space last name in quotes")
-    subpi.add_argument("institution", help="person's institution.  Can be inf"
-                                            "short form such as columbiau and will "
-                                            "be retrieved from institutions collection")
+    subpi.add_argument("--id", help="id of the person, e.g., first letter first name "
+                                    "plus last name, but unique")
+    subpi.add_argument("--inst", help="person's institution.  Can be inf "
```
sbillinge, line 21: maybe have a single letter abbreviation, ``-i`` for institution and ``-d` for id?

Also inf is a typo?

regolith/helpers/u_contacthelper.py
```diff
@@ -82,7 +88,11 @@ def db_updater(self):
             notes = current.get('notes', [])
             aliases = current.get('aka', [])
         else:
-            pdoc.update({"date": dt.today()})
+            if not rc.inst:
```
sbillinge, line 91: I think you need to raise an exception here, not just a print statement

regolith/helpers/u_contacthelper.py
```diff
@@ -82,7 +88,11 @@ def db_updater(self):
...
+                print("Institution (--inst) is required to create a new contact")
+                return
+            pdoc.update({"date": dt.date.today()})
+            pdoc.update({"institution": rc.inst})
             notes = []
             aliases = []
             UUID = str(uuid.uuid4())
```
sbillinge, line 98: this should be lower-case for a string variable.

dy2403: changed to uniqueidentifier, it gets angry if I try to use uuid

regolith/helpers/u_contacthelper.py
dy2403: changed to:
`            if not rc.institution:
                raise RuntimeError("Institution is required to create a new contact")`

regolith/helpers/u_contacthelper.py
dy2403: done!

dy2403:
@sbillinge yes, can you review it and merge? 
I wasn't sure about the news, hope I did it correctly

sbillinge:
thanks @dy2403

PR #460, opened 2020-06-18
u_contact helper 

dy2403:
hi @sbillinge !
I changed it, and now it retrieves a list and accepts either a fragment of the name or the id!
Can you please review it?

### DIFF
--- a/regolith/helpers/u_contacthelper.py
+++ b/regolith/helpers/u_contacthelper.py
@@ -73,26 +73,23 @@ class ContactUpdaterHelper(DbHelperBase):
     def db_updater(self):
         rc = self.rc
         name = HumanName(rc.name)
-        frag = fragment_retrieval(self.gtx['contacts'], ["_id", "aka", "name"], rc.name)
-        ids = set(i.get('_id') for i in frag)
-        sorted_ids = sorted(ids)
-        index = 1
-        all_names = {}
-        for i in sorted_ids:
-            currentid = rc.client.find_one(rc.database, rc.coll, {'_id': i})
-            index += 1
-            all_names[str(index)] = {"_id":i, "name":currentid.get("name")}
+        sorted_ids = sorted(set(i.get('_id') for i in fragment_retrieval(self.gtx['contacts'],
+                                            ["_id", "aka", "name"], rc.name)))
+        index = list(range(len(sorted_ids)))
         if not rc.number:
             print("Please choose from one of the following to update/add:")
-            print(f"1. {rc.name} as a new contact")
-            for k, v in all_names.items():
-                print(f"{k}. {v['name']}    id: {v['_id']} ")
+            print(f"{0}. {rc.name} as a new contact")
+            print(*[f"{i+1}. {rc.client.find_one(rc.database, rc.coll, {'_id': j})['name']}    id: {j}\n"
+                                                                    for i, j, in zip(index, sorted_ids)])
             return
         pdoc = {}
-        if rc.number == '1':
+        if rc.number == 0:
             if not rc.institution:
                 raise RuntimeError("institution is required to create a new contact")
-            key = str(name.first[0].lower().replace(" ", "") + name.last.lower().replace(" ", ""))
+            if not rc.d:
+                key = str(name.first[0].lower().replace(" ", "") + name.last.lower().replace(" ", ""))
+            else:
+                key = rc.d
             pdoc.update({"name": name.full_name})
             pdoc.update({"date": dt.date.today()})
             pdoc.update({"institution": rc.institution})
@@ -101,7 +98,7 @@ class ContactUpdaterHelper(DbHelperBase):
             uniqueidentifier = str(uuid.uuid4())
             pdoc.update({'uuid': uniqueidentifier})
         else:
-            key = all_names[str(rc.number)]['_id']
+            key = sorted_ids[int(rc.number)-1]
             current = rc.client.find_one(rc.database, rc.coll, {'_id': key})
             notes = current.get('notes', [])
             aliases = current.get('aka', [])
--- a/regolith/helpers/u_contacthelper.py
+++ b/regolith/helpers/u_contacthelper.py
@@ -30,7 +30,7 @@ def subparser(subpi):
                              "quotes separated by a space")
     subpi.add_argument("--notes", nargs='+',
                         help="notes.  As many notes as you like, each one in "
-                             "quotes and separated by a space, such as where"
+                             "quotes and separated by a space, such as where "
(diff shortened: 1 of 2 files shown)


regolith/helpers/u_contacthelper.py
```diff
@@ -75,9 +75,10 @@ def db_updater(self):
         name = HumanName(rc.name)
         frag = fragment_retrieval(self.gtx['contacts'], ["_id", "aka", "name"], rc.name)
         ids = set(i.get('_id') for i in frag)
+        sorted_ids = sorted(ids)
         index = 1
         all_names = {}
-        for i in ids:
+        for i in sorted_ids:
             currentid = rc.client.find_one(rc.database, rc.coll, {'_id': i})
```
sbillinge, line 82: I am not sure what all this code is.  can't you do it with something like this?
```
sorted_ids = list(set(i.get("_id") for n in fragment_retrieval(self.gtx['contacts'], ["_id", "aka", "name"], rc.name))).sort()
index = list(range(len(sorted_ids)))
print(*[f"{i}. {j}\n" for i, j, in zip(index, sorted_ids))
```
if you need other things from the retrieved entries you can assign the return from fragment retrieval to a variable and get the ids from that.

regolith/helpers/u_contacthelper.py
```diff
@@ -117,7 +118,7 @@ def db_updater(self):
         pdoc.update({"aka": aliases})
         pdoc.update({"notes": notes})
         pdoc.update({'updated': now})
-
+        print(all_names)
```
sbillinge: is this a debug print?

dy2403: yes, I delete it

regolith/helpers/u_contacthelper.py
dy2403: ok! done!

tests/test_helpers.py
```diff
@@ -84,6 +80,21 @@
...
+    (["helper", "u_contact", "afriend", "--number", "1",
+      "--notes", "Test note", "-a", "Friend", "--date", "2020-01-02"],
+     "afriend has been added/updated in contacts\n"
+     ),
+    (["helper", "u_contact", "Anthony", "--date", "2020-01-02"],
+     "Please choose from one of the following to update/add:\n"
+     "0. Anthony as a new contact\n"
+     "1. Anthony B Friend    id: afriend\n"
+     "\n"
+     ),
```
dy2403, line 92: I am annoyed at this last \n, but spent more time than I should trying to get rid of it. 
do you have any solution?

regolith/helpers/u_contacthelper.py
```diff
@@ -73,26 +73,23 @@ def construct_global_ctx(self):
...
+        sorted_ids = sorted(set(i.get('_id') for i in fragment_retrieval(self.gtx['contacts'],
+                                            ["_id", "aka", "name"], rc.name)))
+        index = list(range(len(sorted_ids)))
         if not rc.number:
             print("Please choose from one of the following to update/add:")
-            print(f"1. {rc.name} as a new contact")
-            for k, v in all_names.items():
-                print(f"{k}. {v['name']}    id: {v['_id']} ")
+            print(f"{0}. {rc.name} as a new contact")
+            print(*[f"{i+1}. {rc.client.find_one(rc.database, rc.coll, {'_id': j})['name']}    id: {j}\n"
```
sbillinge, line 82: I guess what I was trying to avoid was multiple calls to the db which are slow.  I think the fragment_retrieval gives you everything you need so we can avoid the second db call

sbillinge: just to expand on that...you will have to modify the code I gave you to do it in two steps, sthg like
```
found_contacts = fragment_retrieval....
sorted_ids = list(set([i.get("_id") for i in found_contacts])).sort()
```
higher up, but then everything you need for the print you can get directly from ``found_contacts``

dy2403: done!

tests/test_helpers.py
```diff
@@ -85,15 +85,15 @@
      "afriend has been added/updated in contacts\n"
      ),
     (["helper", "u_contact", "Anthony", "--date", "2020-01-02"],
-     "Please choose from one of the following to update/add:\n"
+     "Please, rerun the helper with the same name as previously inputted,"
```
sbillinge, line 88: How about "please rerun the helper specifying '-n list-index' to update item number 'list-index'"

dy2403: done!

regolith/helpers/u_contacthelper.py
```diff
@@ -4,34 +4,33 @@
...
 
 
 TARGET_COLL = "contacts"
 
 def subparser(subpi):
-    subpi.add_argument("name", help="first name space last name in quotes")
+    subpi.add_argument("name", help="name, name fragment (single argument only) or id "
+                                    "used to find an existing contact. "
+                                    "please, inform full name if this is a new contact")
+    subpi.add_argument("--number", help="number in the numbered list")
```
sbillinge, line 20: I think this should maybe be called ``--index`` and ``-i`` and help is ``index of the item in the enumerated list chosen to update``

regolith/helpers/u_contacthelper.py
```diff
@@ -71,34 +73,39 @@ def construct_global_ctx(self):
...
             pdoc.update({"institution": rc.institution})
             notes = []
             aliases = []
             uniqueidentifier = str(uuid.uuid4())
             pdoc.update({'uuid': uniqueidentifier})
-        #FIXME
-        # if rc.e_email:
-        #     pdoc.update({"email": rc.email})
+        else:
+            current = found_contacts[int(rc.number)-2]
```
sbillinge, line 101: I would type-check rc.number right at the top and throw an exception if it is not an int.  then remove the int here. Otherwise the user may get unexpected behavior (no magic rule)

dy2403: I added type=int to the subparser,  and I think it will make sure that it is an integer!
I did a test passing a date and it yelled at me.

regolith/helpers/u_contacthelper.py
dy2403: done!

dy2403:
hi @sbillinge !
I think this is ready for a final review! also, does this needs a news?

regolith/helpers/u_contacthelper.py
sbillinge: nice work!

sbillinge:
thanks @dy2403

PR #466, opened 2020-06-21
grant burn function

adibaejaz:
a function to get the amounts of a grant for a given period of time and list of people


adibaejaz:
@sbillinge I have pushed a rudimentary version of the `get_grant_amount` function called in the `makeappointments` helper. please let me know if I am understanding the requirements correctly!

regolith/tools.py
```diff
@@ -1287,4 +1289,44 @@ def get_id_from_name(coll, name):
...
         return person["_id"]
-    else: return None
\ No newline at end of file
+    else: return None
+
+
+def get_grant_amount(grant, people, begin_date=None, end_date=None):
+    grant_amounts = []
+    if not grant.get('budget'):
+        return "This grant has no specified budget"
```
sbillinge, line 1298: can you print the id of the offending grant maybe?

regolith/tools.py
```diff
@@ -1287,4 +1289,44 @@ def get_id_from_name(coll, name):
                              case_sensitive=False)
     if person:
         return person["_id"]
-    else: return None
\ No newline at end of file
+    else: return None
+
+
+def get_grant_amount(grant, people, begin_date=None, end_date=None):
+    grant_amounts = []
```
sbillinge, line 1296: nice!   I am officially disappointed by two things though.
1) no tests!  Remember, tests first, code after.
2) no docstring.  Remember, docstring comes even before the tests as it forces you to describe in words what the function will do.

regolith/tools.py
adibaejaz: done

regolith/tools.py
adibaejaz: sorry about that! I have pushed the docstring. the tests are quite tedious to write since there are so many parameters, but I am working on them now! I will take it as an exercise in imagination to pretend I haven't written the code

sbillinge: Thanks.  Tests are tedious, until you have been coding for a while and you realize how transformative they are.

adibaejaz: I have pushed the tests!

regolith/tools.py
```diff
@@ -1322,3 +1322,63 @@ def is_fully_appointed(person, begin_date, end_date):
             else:
                 print("appointment gap for {} on {}".format(person.get('_id'), str(day)))
     return status
+
+
+def get_grant_amount(grant, ppl_coll, begin_date=None, end_date=None):
```
sbillinge, line 1327: Thanks @adibaejaz , let's start our work here.

Before I look at the tests, lets discuss what this function is supposed to do.  This is captured in the name, and in the docstring.  From the name I see ``get_grant_amount``.  A function with this name I would probably expect to take one argument (grant_identifier) and have it return a float (the amount of the grant).  That is probably not what you have in mind so we probably want to work on this.

Then the docstring says "Parameters" but there is no description of what the function does.  I can maybe piece it together from the parameters and the returns below, but the overall objective is not clear.

Next, it would be helpful to think a bit about the design of the code.  What functions you will make and are they th eright functions.  Code is more modular and reusable if you make functions that do small, granular (and preferrably abstractish) things and then you make functions to do specific things by chaining them together.  

btw, we should issue a warning or better, throw and exception, if someone has concurrent appointments of a different type.  They can't be a pd and a gra at the same time....

adibaejaz: Thank you Professor!
1. I have edited the docstring to include a description but it might not be very succinct due to the complications of the function itself (which probably suggests I should work on breaking the function down, as you said)
2. As for the name, do you think something like `get_net_grant_amount` or `get_net_grant_budget` be more informative?
3. Do you think the inconsistent appointment type check might be better suited to the `is_fully_appointed` function?  Or should I add it here?
4. I am trying without much success to think of ways to make this function more abstract and/or granular although I completely see the need, especially since so much of it is dependent on the internal properties of the `grant` object.  I'm not sure how to make this more granular, but on making it more abstract, I could possibly redesign the function to take in any object `thing_gives_money` that has a `budget` field, sum over all its `budget_periods`, and deduct values from it by iterating through a list of objects `thing_takes_money` which are appointed on `thing_gives_money`, and then return the amounts over a period of time? Please let me know what you think!

adibaejaz:
@sbillinge I have also edited the function to raise `ValueError`'s instead of returning an error message and added  a test for exception checking!

regolith/tools.py
sbillinge: Thanks, this is very helpful.  So I thought a bit about this and I think what we are actually trying to do is to build a "burn", which can be thought of as a plot of amount vs. day of each grant

How about the function name ``burn(grant, appointments, begin_date, end_date)``.  

How about we pass in a collection of appointments rather than a people collection, since we don't care about the people.  This means that we can have a function that collects appointments. so a structure a bit like:
```
def collect_appts(people, grant_id, begin_date=None, end_date=None):
    check that either both begin and end date are specified, or neither.
    apps = []
    for person in people:
       for appt in person['appts']:
           if appt[grant] == grant_id:
                 if begin_date and end_date:
                        if is_current(appt, begin_date, end_date):
                              apps.append(appt)
                 else:
                       apps.append(appt)
   ....
   return appts

def burn(grant_coll, appts_coll, begin_date, end_date):
    burn, day = [], []
    for day in date_range:
         net_load = 0
         for budget in grant:
               add budgeted amounts to net_load
         for appt in appts:
               subtract appt amounts
          check for over-under and do whatever with that info
          burn.append[net_load]
          day.append[day]                     #not sure the best format here...could be just an index but could be a date object?
    return burn, day
          

all_appts = collect_appts(people)
grant1_burn, days = burn(grant1_coll, all_appts, begin_date, end_date)
plot(days, grant1_burn)
```
Of course this will reuse most of your existing code, but just distribute it around a bit, but I think this gives us the flexibility and reusability that we are after.

adibaejaz:
Yes, thank you for this! I will work on restructuring the code. However, for the `burn` function, I think it might be better if we collect all appointments, rather than those only in the date range provided by the user, since the grant amount may depend also on its expenditure before the `begin_date` provided. So I will still include `begin_date` and `end_date` parameters in `coll_appts` since this would be useful, but will default it to collecting all appointments in the database

regolith/tools.py
adibaejaz: Yes, thank you very much for this! I will work on restructuring the code. However, for the burn function, I think it might be better if we collect all appointments, rather than those only in the date range provided by the user, since the grant amount may depend also on its expenditure before the begin_date provided. So I will still include begin_date and end_date parameters in coll_appts since this would be useful, but will default it to collecting all appointments in the database

adibaejaz: also a workflow question - is it ok if I work on both of these functions in the same PR? or should I create a new PR for the `collect_appts` function?

adibaejaz: also Professor, how valuable is it to let the user pass in a `grants_coll` instead of simply a `grant`?  for the purposes of checking overspent/ underspent/ appointment on grant out of money in the `makeappointments` helper, we only need to check one grant at a time. adding a multiple grants functionality complicates the code significantly  with several nested loops and such. if we want to plot the aggregate burn, I could write another function that calls this one for multiple grants, and let the `burn` function tackle one grant at a time?

sbillinge: yes, I think you are right on both points.  I was thinking about it being burn per grant, and then total burn is just the sum of these.  Not sure why I put grant_coll there.

sbillinge:
> 
> 
> @sbillinge I have also edited the function to raise `ValueError`'s instead of returning an error message and added tests for exception checking!

great!

regolith/tools.py
```diff
@@ -1497,21 +1497,21 @@ def collect_appts(ppl_coll, filter_key=None, filter_value=None, begin_date=None,
...
+    Retrieves the total burn of a grantt over an interval of time by integrating over all appointments
+    made on the grant.
 
     Parameters
     ----------
     grant: dict
-        The person whose appointments need to be checked
-    ppl_coll: collection (list of dicts)
-        The people collection that contains the grant appointments
+        The grant_id or grant object whose burn needs to be retrieved
```
sbillinge, line 1508: this doesn't seem to be consistent.  It says type dict but then you say grant_id is ok

regolith/tools.py
```diff
@@ -1497,21 +1497,21 @@ def collect_appts(ppl_coll, filter_key=None, filter_value=None, begin_date=None,
                 appts[-1].update({'person': p.get('_id')})
     return appts
 
-def get_grant_amount(grant, ppl_coll, begin_date=None, end_date=None):
+def grant_burn(grant, appts, begin_date=None, end_date=None):
     """
-    Given a grant, a database of people appointed on assorted grants, and two dates marking the limits of an interval,
-    it returns a list of the net amount of the grant on each day of the interval.
+    Retrieves the total burn of a grantt over an interval of time by integrating over all appointments
```
sbillinge, line 1502: typo in grant

regolith/tools.py
```diff
@@ -1497,21 +1497,21 @@ def collect_appts(ppl_coll, filter_key=None, filter_value=None, begin_date=None,
...
     grant: dict
-        The person whose appointments need to be checked
-    ppl_coll: collection (list of dicts)
-        The people collection that contains the grant appointments
+        The grant_id or grant object whose burn needs to be retrieved
+    appts: collection (list of dicts)
+        The collection of appointments made on assorted grants
     begin_date: datetime, string, optional
-        The start date of the interval of time to check the grant amount for
+        The start date of the interval of time to retrieve the grant burn for
```
sbillinge, line 1512: needs more I think.  What is the default behavior?  If string, does it have to be ain a particular format?

tests/test_tools.py
```diff
@@ -875,40 +875,8 @@ def test_collect_appts(people, key, value, start, end, expected):
...
-     },
-    {'name': 'GEM Anscombe', '_id': 'ganscombe', 'appointments': []},
-    {'name': 'Sophie Germain', '_id': 'sgermain',
-     'appointments':
-         [{'_id': 'A', 'begin_date': '2019-09-02', 'end_date': '2019-09-06', 'grant': 'grant4', 'loading': 1.0,
-           'type': 'ss'}]
-     },
-    ]
+
+appts = collect_appts(appointed_people)
```
sbillinge, line 879: haha, nice!  I'm loving this new coding style! 👍

tests/test_tools.py
```diff
@@ -973,21 +941,20 @@ def test_collect_appts(people, key, value, start, end, expected):
          {'date': '2019-09-06', 'postdoc_days': 30.5, 'ss_days': 25.5, 'student_days': 30.5},
          {'date': '2019-09-07', 'postdoc_days': 30.5, 'ss_days': 25.5, 'student_days': 30.5}]
          ),
-        ({'_id': 'malicious_grant', 'alias': 'very_malicious_grant'}, grant_people, '2012-12-23', '2013-01-24',
+        ({'_id': 'malicious_grant', 'alias': 'very_malicious_grant'}, appts, '2012-12-23', '2013-01-24',
```
sbillinge, line 944: its a bit picky, but I don't think anyone in our group is being malicious.  Do you mean misbehaving_grant? or if we want more humour in our code (we do!) maybe "mischievous"

tests/test_tools.py
```diff
@@ -973,21 +941,20 @@ def test_collect_appts(people, key, value, start, end, expected):
...
           'malicious_grant has no specified budget'
          ),
-        (grant1, [{'name': 'Malicious Person', '_id': 'mperson',
-         'appointments': [{'_id': 'A', 'begin_date': '2019-09-01', 'end_date': '2019-09-05', 'loading': 1.0,
-                          'grant': 'grant1', 'type': 'imaginary'}]}],
-         None, None, 'invalid  type for appointment A of mperson'
+        (grant1, [{'person': 'malicious person', '_id': 'A', 'begin_date': '2019-09-01', 'end_date': '2019-09-05',
```
sbillinge, line 947: pls see above

regolith/tools.py
adibaejaz: removed the `grant_id` option! I thought we could allow both and use `client.find_one` to retrieve the grant, but I'm not sure that would work nested in the function since the grants database varies

regolith/tools.py
adibaejaz: edited!

regolith/tools.py
adibaejaz: done!

tests/test_tools.py
adibaejaz: thank you!

tests/test_tools.py
adibaejaz: yes sorry! the grant and person are now `magical` rather than `malicious`

regolith/tools.py
adibaejaz: done!

sbillinge:
lovely!   I removed WIP from the PR title and I merged

PR #467, opened 2020-06-22
WIP: todolist helper

jiaweizang:
I set the default user name in ~/.config/regolith/user.json:
{ "default_user_id": "jwzang",
 "user_first_name": "Jiawei"
}

There are three parts: 
l_todo: list the to-do list, milestones and actions from group meeting
a_todo: add a task to the list
u_todo: mark a task as finished by regolith helper u_todo mark-number (or delete that task with -d). The mark-number can be found by "regolith helper l_todo -v"

There are still some bugs and I will fix them later~

### DIFF
--- a/regolith/helpers/a_todohelper.py
+++ b/regolith/helpers/a_todohelper.py
@@ -16,23 +16,27 @@ from regolith.tools import (
 )
 
 TARGET_COLL = "todolist"
-Importance = [0, 1, 2, 3, 4, 5]
+Importance = [0, 1, 2]
 
 
 def subparser(subpi):
     subpi.add_argument("description", help="the description of the to_do task",
                        default=None)
+    subpi.add_argument("due_date",
+                       help="Proposed due date for the to_do task. Please enter a number. 5 means 5 days from the "
+                            "begin_date. "
+                       )
+    subpi.add_argument("-b", "--begin_date",
+                       help="Begin date of the task. Default is today."
+                       )
     subpi.add_argument("-i", "--id", help="ID of the to_do list's owner. The default id is saved in user.json. ")
-    subpi.add_argument("-d", "--due_date",
-                       help="Proposed due date for the to_do task."
-                       )
     subpi.add_argument("-p", "--importance",
-                       help=f"The importance of this task from {Importance}. Default is 0 (unimportant).",
-                       default=0
+                       help=f"The importance of this task from {Importance}. Default is 1.",
+                       default=1
                        )
-    subpi.add_argument("-e", "--estimated_time",
-                       help="The estimated time that the task will take. The unit is minute. Default is 60 ",
-                       default = 60
+    subpi.add_argument("-d", "--duration",
+                       help="The estimated duration that the task will take. The unit is minute. Default is 60 ",
+                       default=60
                        )
     return subpi
 
@@ -65,7 +69,6 @@ class TodoAdderHelper(DbHelperBase):
         gtx["str"] = str
         gtx["zip"] = zip
 
-
     def db_updater(self):
         rc = self.rc
         if not rc.id:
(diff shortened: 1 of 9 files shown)


jiaweizang:
I can run these three helpers successfully in command line, but I also updated the schema file, so maybe there are some bugs in schemas.py (and may also in three helper files).

jiaweizang:
After debugging, all tests can passed. (The bug is that I wrote "int" instead of the correct "integer" in schemas..)
Up to now, l_todo helper can pass the tests, I will later test a_todo and u_todo. And we can also discuss what we need to improve for l_todo. 
Currently users need to provide their first name, cause we need to use this as key to filter tasks from actions in grouping meetings.yml. 
By the way, the order of the to_do list helper is first by time order, then by importance order.

sbillinge:
> 
> 
> I set the default user name in ~/.config/regolith/user.json:
> { "default_user_id": "jwzang",
> "user_first_name": "Jiawei"
> }
> 
> There are three parts:
> l_todo: list the to-do list, milestones and actions from group meeting
> a_todo: add a task to the list
> u_todo: mark a task as finished by regolith helper u_todo mark-number (or delete that task with -d). The mark-number can be found by "regolith helper l_todo -v"
> 
> There are still some bugs and I will fix them later~

Here are what I think would be useful helpers:
 - a_todo
 - l_todo (both these same as your definition)
 - finish_todo (maybe think of a shorter name!)   given a todo-id or position in the todo list, mark it as done and set the end_date to today (if not specified)
 - u_todo - all other updates.  Add notes to notes list, assign to someone else, change importance or due date (basically allow anything to be changed but make common things easier maybe?)
 - delete_todo

regolith/helpers/a_todohelper.py
```diff
@@ -0,0 +1,126 @@
...
+def subparser(subpi):
+    subpi.add_argument("description", help="the description of the to_do task",
+                       default=None)
+    subpi.add_argument("-i", "--id", help="ID of the to_do list's owner. The default id is saved in user.json. ")
+    subpi.add_argument("-d", "--due_date",
+                       help="Proposed due date for the to_do task."
+                       )
+    subpi.add_argument("-p", "--importance",
+                       help=f"The importance of this task from {Importance}. Default is 0 (unimportant).",
+                       default=0
```
sbillinge, line 31: default importance should probably be somewhere in the middle maybe but perhaps we need to start using this to learn how to do this best.  I am guessing something less granular might be better 0,1,2 but we will see.

regolith/helpers/a_todohelper.py
```diff
@@ -0,0 +1,126 @@
...
+                       )
+    subpi.add_argument("-e", "--estimated_time",
```
sbillinge, line 33: estimated duration

regolith/helpers/a_todohelper.py
```diff
@@ -0,0 +1,126 @@
...
+        rc = self.rc
+        if not rc.id:
+            try:
+                rc.id = rc.default_user_id
+            except AttributeError:
+                raise RuntimeError(
+                    "Please set default_user_id and user_first_name in '~/.config/regolith/user.json',\
+                     or you need to input your id in the command line")
+        now = dt.date.today()
+        if not rc.due_date:
```
sbillinge, line 79: I wonder if it is better to make due date required (but easy to type.... so select from tomorrow, two days, one week, one month or allow an ISO date if there is a more precise idea)

jiaweizang:
Travis doesn't work after rebasing, so I will close this pull request, and create separate branches for l_todo, a_todo, f_todo..

sbillinge:
> 
> 
> Travis doesn't work after rebasing, so I will close this pull request, and create separate branches for l_todo, a_todo, f_todo..

OK, good plan.

PR #474, opened 2020-06-24
first version of l_todo helper

jiaweizang:
I tested several cases and all passed.  If the basic function of l_todo can be settled as this, maybe we could merge l_todo first, then add a_todo, f_todo... one by one?


jiaweizang:
@sbillinge Hi Simon please see the new codes. 
By the way, currently the l_todo will print "notes" only in --verbose case. 
I also gather and print "finished" items from todolist.yml in --verbose, so users can see all tasks and related marks, in case they want to update finished tasks identified by marks when they use u_todo. Finished tasks from milestones won't be printed.

sbillinge:
Thanks so much @jiaweizang I am excited to try this.... I am sure I will have many requests!

PR #475, opened 2020-06-24
General lister

### LINKED ISSUE #452: helper: general lister
UC1:
1. simon types ``regolith helper list people``
1. lister lists the ids of all the people in people

UC2:
1. S types ``rh list people --kv_filter name simon``
1. l lists the id of all people in people who's name contains simon

UC3:
1. S types ``rh list people --kv_filter name simon -r name position``
1. l lists the id, name and position of all people in people who's name contains simon

UC4:
1. S types ``rh list people -k``
1. l returns a list of the available keys and their description (if available) in collection people

dy2403:
closes #452 
Hi, Prof. @sbillinge !
I wrote the tests, can you please take a look at this? 
Also, I am not sure what UC4 should really do, can you help me? 

**UC1:**
simon types regolith helper list people
lister lists the ids of all the people in people
**UC2:**
S types rh list people --kv_filter name simon
l lists the id of all people in people who's name contains simon
**UC3:**
S types rh list people --kv_filter name simon -r name position
l lists the id, name and position of all people in people who's name contains simon
**UC4:**
S types rh list people -k
l returns a list of the available keys and their description (if available) in collection people


tests/test_helpers.py
```diff
@@ -105,6 +105,21 @@
      "Please rerun the helper specifying '-n list-index' to update item number 'list-index':\n"
      "1. Maria as a new contact\n"
      ),
+    #UC1
+    (["helper", "list", "people"],
+     "scopatz\nsbillinge\n"),
+    #UC2
+    (["helper", "list", "--kv_filter", "name", "simon"],
```
sbillinge, line 112: doesn't this need a collection?

tests/test_helpers.py
```diff
@@ -105,6 +105,21 @@
...
+     "sbillinge\n"),
+    #UC3
+    (["helper", "list", "--kv_filter", "name", "simon", "-r", "name", "position"],
+     "sbillinge, name: Simon J. L. Billinge, position: professor\n"),
+    #UC4?
+    (["helper", "list", "people", "-k"],
```
sbillinge, line 118: is this general or specific?  If you put this in it is not a general lister.  UC4 is a specific example that needs to be supported by giving a specific set of inputs to your general lister

tests/test_helpers.py
dy2403: yes, I does! I will add it!

tests/test_helpers.py
dy2403: would it be like: if the person gives the collection projecta, then it would give: 
--leader : (description)
--group_members: (description)
etc.. 
and for the people collection it would be:
--name: name of the person
--position: position of the person
?

dy2403:
Hi @sbillinge!
I have a question about the -k and exactly it should retrieve. (From UC4)
Are keys like name, position, etc.? Should it be retrieved from schemas?
Or should I establish the keys available for each collection in the function?

sbillinge:
> Hi @sbillinge!
> I have a question about the -k and exactly it should retrieve. (From UC4)
> Are keys like name, position, etc.? Should it be retrieved from schemas?
> Or should I establish the keys available for each collection in the function?

-k there stands for --keys and it lists all the keys that I can give to the lister either to search over, or to return.  Otherwise I may not remember what I have available to search over.

dy2403:
@sbillinge ready for review and merge!

regolith/helpers/l_generalhelper.py
```diff
@@ -7,10 +7,74 @@
     all_docs_from_collection,
     get_pi_id,
     search_collection,
-    collection_str
+    collection_str,
 )
 HELPER_TARGET = "lister"
+KEYS = {"abstracts":{"firstname": "first name of the author.",
```
sbillinge, line 13: no, this is the wrong approach.  

more like
```
if rc.print_keys
    print(*[f"{key}\n" for key in rc.coll.keys()])
    return
```

dy2403: hi @sbillinge!
Does this looks better?

regolith/helpers/l_generalhelper.py
```diff
@@ -0,0 +1,84 @@
...
+def subparser(subpi):
+    subpi.add_argument(
+        "coll",
+        help='collection where the lister is going to run.')
+    subpi.add_argument(
+        "--kv_filter",
+        nargs="+",
+        help='search the given collection by key element pairs')
+    subpi.add_argument(
+        "-r",
```
sbillinge, line 23: I think this needs a long name, --return_fields or sthg.  Help could "keys of fields whose values will be printed after filtering."

regolith/helpers/l_generalhelper.py
```diff
@@ -0,0 +1,84 @@
...
+    collection_str,
+)
+HELPER_TARGET = "lister"
+
+def subparser(subpi):
+    subpi.add_argument(
+        "coll",
+        help='collection where the lister is going to run.')
+    subpi.add_argument(
+        "--kv_filter",
```
sbillinge, line 19: this needs a short name, like -f (for filter) (long name is ok as is)

regolith/helpers/l_generalhelper.py
```diff
@@ -0,0 +1,84 @@
...
+        action="store_true",
+        help='list of the available keys and their description')
+    return subpi
+
+class GeneralListerHelper(SoutHelperBase):
+    """Helper for finding and listing contacts from the contacts.yml file
+    """
+    # btype must be the same as helper target in helper.py
+    btype = HELPER_TARGET
+    needed_dbs = ["abstracts", "assignments","beamplan", "beamtime",
```
sbillinge, line 37: this is not maintainable, and will make this function very slow.  Maybe it can work if you put the variable assignment into the ``construct_global_ctx`` method and define it as:
```
needed_dbs = [rc.coll]
```

tests/test_helpers.py
```diff
@@ -115,6 +115,44 @@
      "Please rerun the helper specifying '-n list-index' to update item number 'list-index':\n"
      "1. Maria as a new contact\n"
      ),
+    (["helper", "lister", "people"],
```
sbillinge, line 118: fantastic!

regolith/helpers/l_generalhelper.py
dy2403: done!

regolith/helpers/l_generalhelper.py
dy2403: done!

regolith/helpers/l_generalhelper.py
dy2403: done!

regolith/helpers/l_generalhelper.py
```diff
@@ -0,0 +1,78 @@
...
+HELPER_TARGET = "lister"
+
+def subparser(subpi):
+    subpi.add_argument(
+        "coll",
+        help='collection where the lister is going to run.')
+    subpi.add_argument(
+        "-f","--kv_filter",
+        nargs="+",
+        help='search the given collection by key element pairs')
```
sbillinge, line 21: this needs an example or it will be too hard for people to know what to do here.  If you search around you should see a nice help message for this in other listers, written by Michael.

Also, let's decide if we are going to put periods at the end of help messages or not, and make it all consistent.  I think yes, have a period.  It's better English.

dy2403:
I do have a prum! Should I finish it when it is merged, or when we guarantee that it is working when we test it?

regolith/helpers/l_generalhelper.py
dy2403: Does it look better?

regolith/helpers/l_generalhelper.py
```diff
@@ -14,19 +14,22 @@
 def subparser(subpi):
     subpi.add_argument(
         "coll",
-        help='collection where the lister is going to run.')
+        help='Collection where the lister is going to run.')
     subpi.add_argument(
         "-f","--kv_filter",
         nargs="+",
-        help='search the given collection by key element pairs')
+        help="Search the given collection by key element pairs. "
```
sbillinge, line 21: "key-value pairs".  We don't call those things "elements"

regolith/helpers/l_generalhelper.py
```diff
@@ -14,19 +14,22 @@
...
+             "e.g. < -f name simon >, will return the id of all the people "
```
sbillinge, line 22: put in the full command, so "e.g., 'regolith helper lister -f name simon' will return...."

regolith/helpers/l_generalhelper.py
```diff
@@ -14,19 +14,22 @@
...
+             "who's name contains simon.")
     subpi.add_argument(
         "-r",  "--return_fields",
         nargs="+",
-        help='keys of fields whose values will be printed after filtering.')
+        help="Specify what keys to return values from when running --filter. "
```
sbillinge, line 27: I think this should work independent of whether -f is specified?  Why not?  What if I want to get certian key values but not apply a filter so from everything in the collection?  It seems like a reasonable request, right?

wording -> "Specify from which keys to print values, e.g.,' regolith helper lister people -r name status title"

dy2403: done!

regolith/helpers/l_generalhelper.py
dy2403: done!

regolith/helpers/l_generalhelper.py
dy2403: done!

dy2403:
Hi prof. @sbillinge,
I think this is ready to review and merge!

tests/test_helpers.py
```diff
@@ -115,6 +115,49 @@
...
+     "sbillinge    \nscopatz\n"),
+    (["helper", "lister", "projecta"],
+     "20ly_newprojectum    \n20sb_firstprojectum\n"),
+    (["helper", "lister", "contacts"],
+     "afriend\n"),
+    (["helper", "lister", "proposals"],
```
sbillinge, line 124: I think these tests are testing the same functionality so after the first one they can be removed.  You don't have to iterate through all the collections....

tests/test_helpers.py
```diff
@@ -115,6 +115,49 @@
...
+     "20_anewproposal    \ndmref15    \nmypropsal\n"),
+    (["helper", "lister", "abstracts"],
+     "Mouginot.Model\n"),
+    (["helper", "lister", "people", "--kv_filter", "name", "simon"],
+     "sbillinge\n"),
+    (["helper", "lister", "projecta","--kv_filter", "lead", "ascopatz"],
+     "20sb_firstprojectum\n"),
+    (["helper", "lister", "contacts","--kv_filter", "institution", "columbiau"],
```
sbillinge, line 132: again, you only need additional tests if it is testing new functionality, or an edge case.  e.g., for edge cases what happens if the person gives an arg that is not found at all, or is invalid in some way.  What do you want the program to do?  for the different functionality, for example pulling from a nested field (though we don't have that UC implemented yet, but it would be different functionality because it would be testing retrieving info from a nested field rather than a top-level field)

tests/test_helpers.py
```diff
@@ -115,6 +115,49 @@
...
+     ),
+    (["helper", "lister", "projecta", "--keys"],
+     "['_id', 'begin_date', 'collaborators', 'deliverable', 'description', "
+     "'end_date', 'grants', 'group_members', 'kickoff', 'lead', 'log_url', "
+     "'milestones', 'name', 'pi_id', 'status']\n"),
+    (["helper", "lister", "contacts", "--keys"],
+     "['_id', 'aka', 'day', 'department', 'email', 'institution', 'month', 'name', "
+     "'notes', 'title', 'updated', 'uuid', 'year']\n"
+     ),
+    (["helper", "lister", "people", "--return_fields", "name", "position"],
```
sbillinge, line 156: good.

tests/test_helpers.py
```diff
@@ -115,6 +115,49 @@
...
+     "afriend    name: Anthony B Friend\n"),
+    (["helper", "lister", "people", "--keys"],
+     "['_id', 'activities', 'aka', 'appointments', 'avatar', 'bio', 'bios', "
+     "'committees', 'education', 'email', 'employment', 'facilities', 'funding', "
+     "'github_id', 'google_scholar_url', 'hindex', 'home_address', 'initials', "
+     "'membership', 'miscellaneous', 'name', 'office', 'orcid_id', 'position', "
+     "'publicity', 'research_focus_areas', 'service', 'skills', 'teaching', "
+     "'title']\n"
+     ),
+    (["helper", "lister", "projecta", "--keys"],
```
sbillinge, line 148: see above

regolith/helpers/l_generalhelper.py
```diff
@@ -0,0 +1,86 @@
...
+            if rc.return_fields:
+                print((search_collection(coll, rc.kv_filter, rc.return_fields)).strip())
+                return
+            else:
+                print((search_collection(coll, rc.kv_filter)).strip())
+                return
+        if rc.return_fields:
+            print(collection_str(coll, rc.return_fields).strip())
+            return
+        if rc.keys:
```
sbillinge, line 82: you probably don't have adequate tests for testing edge cases because the structure of this code doesn't look right.  For example,  what if someone specifies a kv_filter and --keys?  What do you want the code to do, etc.?  Think about all the different scenarios, build a test for that, then rework the code.

btw you code is looking very clean and nice, s that is awesome!!!!!  You are turning into a really good pythonista!

dy2403: Hi @sbillinge!
I added tests for when --keys is called with kv_filter or return_fields. It will now return the keys after the results!
Does it look ok?

tests/test_helpers.py
dy2403: Done!

tests/test_helpers.py
dy2403: I added a ValueError for when the key is not valid, and when the value is not found it now prints message!

sbillinge: I am not sure you need value error because I think the parser always load string, but you might want to check that.

dy2403: If you give a not valid key, for example:
` rh lister people  --returrn_fields 'favorite color' 'id' `
it will return:
sbillinge       favorite color: None
dyano           favorite color: None

If the None is okay, I can just discard the ValueError. What do you think?

regolith/helpers/l_generalhelper.py
```diff
@@ -0,0 +1,99 @@
...
+             "e.g. 'regolith helper lister people -r name status title' "
+             "will return the name, status, and title of all the people "
+             "in the collection. If no argument is given the default is just the id.")
+    subpi.add_argument(
+        "-k", "--keys",
+        action="store_true",
+        help="List of the available keys in the collection.")
+    return subpi
+
+def check_key(keys, list_keys):
```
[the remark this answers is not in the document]
dy2403: I think I was overcomplicating things! 
Can you look if it looks better?

regolith/helpers/l_generalhelper.py
```diff
@@ -0,0 +1,99 @@
...
+    get_pi_id,
+    search_collection,
+    collection_str,
+)
+HELPER_TARGET = "lister"
+
+def subparser(subpi):
+    subpi.add_argument(
+        "coll",
+        help='Collection where the lister is going to run.')
```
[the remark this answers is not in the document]
dy2403: done!

tests/test_helpers.py
sbillinge: I see.  I think the preferred behavior would probably be to raise a ``KeyError`` if the key is not found in that case.

sbillinge:
great job @dy2403 , this is an exciting one!

Please can you work on your commit messages to try and make them more informational.  Things like "fix after rebase" essentially contains zero information.  But other than that, this is looking great!

PR #479, opened 2020-06-25
DEV: make regolith work on could based mongodb

st3107:
Hi @sbillinge . I make the regolith work for the cloud based mognodb (also local host mognodb). I test it by building the beamplan from the cloud based rg-db-group in my cloud. Ready for review.

I deprecate the one functionality of the mongoclient in regolith: start the mongo server and close it after the run is finished. The motivation is that (1) starting and closing server take time (2) usually the server will always run in the backend and we don't really need to open and close it every time we run a regolith command (3) if we encounter some error in the run, the server will not be closed and user usually doesn't notice that and when the user tries to run again, it will produce error because regolith will try to start a server that is already running.

### DIFF
--- a/regolith/mongoclient.py
+++ b/regolith/mongoclient.py
@@ -242,7 +242,7 @@ class MongoClient:
             The dictionary of data base information, such as 'name'.
         """
         host = getattr(self.rc, 'host', None)
-        uri = db.get('uri', None)
+        uri = db.get('dst_url', None)
         dbpath = dbpathname(db, self.rc)
         dbname = db['name']
         import_jsons(dbpath, dbname, host=host, uri=uri)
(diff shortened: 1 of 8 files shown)


regolith/mongoclient.py
```diff
@@ -102,6 +105,27 @@ def import_yamls(dbpath: str, dbname: str, host: str = None, uri: str = None) ->
     return
 
 
+def load_mongo_col(col: Collection) -> dict:
```
sbillinge, line 108: is there a way to write a test for this?

st3107: Yes. I will work on it.

st3107:
I tested the mongodb backend on all the builders. It works for all builders except for building figures. I am not sure why. This bug probably takes time to fix. I don't except the mongo backend works for helpers so I didn't add tests for that. This is only the inital step and I think we can first merge it and then start a another step forward.

st3107:
I will skip the figure test for mongo. Not ready for merge.

st3107:
Done.

sbillinge:
thanks @st3107

PR #493, opened 2020-06-29
kv_filter expansion

### LINKED ISSUE #491: ENH: filter and print key functions not so tightly coupled.
Basically, I kind of want the powerful kv_filtering available but then do other interesting things with the filtered data than just print ids or a key value.  Can we integrate the kv_filtering into the main flow?

mwinitch:
Closes #491

### DIFF
--- a/regolith/helpers/l_membershelper.py
+++ b/regolith/helpers/l_membershelper.py
@@ -69,12 +69,12 @@ class MembersListerHelper(SoutHelperBase):
     def sout(self):
         rc = self.rc
         if rc.filter:
-            database = key_value_pair_filter(self.gtx["people"], rc.filter)
+            collection = key_value_pair_filter(self.gtx["people"], rc.filter)
         else:
-            database = self.gtx["people"]
+            collection = self.gtx["people"]
         bad_stati = ["finished", "cancelled", "paused", "back_burner"]
         people = []
-        for person in database:
+        for person in collection:
             if rc.current and not person.get('active'):
                 continue
             people.append(person)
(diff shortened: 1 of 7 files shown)


mwinitch:
I have modified `l_members` to show what I think could be the functionality you want. Basically I have made it so if the filter flags are given the new valid collection becomes the database the lister goes through, otherwise, the lister goes through the default collection. If this looks good to you I can modify the other listers in a similar way.

regolith/helpers/l_membershelper.py
```diff
@@ -68,16 +69,21 @@ def construct_global_ctx(self):
     def sout(self):
         rc = self.rc
         if rc.filter:
-            results = search_collection(self.gtx["people"], rc.filter, rc.keys)
-            print(results, end="")
-            return
+            database = key_value_pair_filter(self.gtx["people"], rc.filter)
```
sbillinge, line 72: This is just right!   super!

I think this function returns a collection, not a database.  Let's make sure we use the jargon correctly to both avoid confusing people and also training them in the jargon.

regolith/helpers/l_membershelper.py
```diff
@@ -68,16 +69,21 @@ def construct_global_ctx(self):
...
         bad_stati = ["finished", "cancelled", "paused", "back_burner"]
         people = []
-        for person in self.gtx["people"]:
+        for person in database:
             if rc.current and not person.get('active'):
                 continue
             people.append(person)
-        #people.sort()
+
+        if rc.filter and not rc.verbose:
```
sbillinge, line 82: yes, I think this will work too.

mwinitch:
@sbillinge What are your thoughts on the updates?

sbillinge:
These look good.  Is it ready for merge (don't forget to remove WIP and ask for the merge)

mwinitch:
@sbillinge This is ready for review and merge

sbillinge:
thanks @mwinitch !

PR #497, opened 2020-06-30
added helper test for appointments

jc-umana:
new PR for old attempt at adding helper test for appointments by grant lister

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -120,6 +120,9 @@ helper_map = [
      "Please rerun the helper specifying '-n list-index' to update item number 'list-index':\n"
      "1. Maria as a new contact\n"
      ),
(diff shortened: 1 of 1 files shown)


tests/test_helpers.py
```diff
@@ -120,6 +120,9 @@
      "Please rerun the helper specifying '-n list-index' to update item number 'list-index':\n"
      "1. Maria as a new contact\n"
      ),
+    (["helper", "l_appointments", "-g", "mrsec14"], "f19, scopatz, loading=0.75, 2019-09-01 to 2019-10-31\n"
```
sbillinge, line 123: a new standard for our tests is to use the long-form of the argument keys and not the short form, so that the sense of the tests can become a bit more readable and  like documentation, so here ``-g`` -> ``--grant``

this test is otherwise ok, but can you also add tests for some edge cases, e.g., what if the code can't find the grant_id that is given, what do we want it to do?

PR #502, opened 2020-07-01
changing types for appts to dict from list

adibaejaz:
I did something rather unfortunate: for the 3 functions I've written that handle appointments,  `is_fully_appointed`, `collect_appts`, and `grant_burn`, I've assumed that `person.get("appointments")` is a list, when in fact it is a dict as per the schema. this PR is to fix that!


adibaejaz:
@sbillinge `grant_burn(appts)` presently only takes in a list of dicts, assuming the user runs `collect_appts` on their object/database. should I also update it to do a different thing if `isinstance(appts, dict)`? or is it safe to assume the user will enter a list of appointments and not a dictionary of appointments?

regolith/tools.py
```diff
@@ -1543,6 +1543,8 @@ def grant_burn(grant, appts, begin_date=None, end_date=None):
             pd_val += b.get('postdoc_months') * 30.5
         if b.get('ss_months'):
             ss_val += b.get('ss_months') * 30.5
+    if isinstance(appts, dict):
+        appts = collect_appts([{"appointments": appts}])
```
adibaejaz, line 1547: @sbillinge  a little bit of cheating. if this is not very readable, I will added a parallel for-loop logic instead for the dict case, but this saves quite some space!

adibaejaz:
@sbillinge ready for review!

sbillinge:
I hate to say this but I probably made a misatake makeing appointments as a dict and a better fix would have been to change it to a list tbh.  It means updating my databases, and it may break other things, but now is probably the right time to do that.   would it be possible to revert the commit that changed those functions and update the schema to make them a list?

sbillinge:
Actually, I changed my mind as I really need this functionality.  I will merge this so you can move forward with the appointments helper, but please could you make an issue to refactor back to lists for the appointments?  We can return to it in afterwards?  Thanks so much.

adibaejaz:
@sbillinge I have opened an issue!

PR #515, opened 2020-07-03
added new milestone types

jc-umana:
Defined ALLOWED_MILESTONE_TYPES and put them in types field

### DIFF
--- a/regolith/schemas.py
+++ b/regolith/schemas.py
@@ -7,6 +7,7 @@ from cerberus import Validator
 from .sorters import POSITION_LEVELS
 
 SORTED_POSITION = sorted(POSITION_LEVELS.keys(), key=POSITION_LEVELS.get)
+ALLOWED_MILESTONE_TYPES = ["meeting", "pr", "deliverable", "approved_file"]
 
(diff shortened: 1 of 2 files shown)


jc-umana:
I've tried... and it would seem that I have failed.

regolith/schemas.py
```diff
@@ -1168,14 +1169,14 @@
                           'initial project plan',
              'audience': ['lead', 'pi', 'group_members'],
              'status': 'proposed',
-             'type': 'meeting'
+             'type': ALLOWED_MILESTONE_TYPES,
```
sbillinge, line 1172: this is not right here.  see if you can figure out the logic.....  What is the EXEMPLAR doing and what is the SCHEMA doing?  How are they doing it?

jc-umana: If I'm being honest, from the file, I have no idea what the difference between the exemplar and the schema is or if they are necessarily separate.

sbillinge: Thanks for the honesty.   Maybe a good topic for a convo with one of the other group members?

PR #516, opened 2020-07-03
WIP: Cleaning up l_projecta

### LINKED ISSUE #480: ENH: projecta lister better handling of whether pra are open or closed
current behavior:
Not sure, so documentation that is displayed by argparse help needs to be a bit richer, but I think by default it only lists pra that are open (not finished, cancelled or back_burner for example)

desired behavior:
other than a fix to the documentation to make this explicit, if the behavior is to list only open (not closed) pra by default, then a command line arg that allows this to be turned off (so ``--all`` prints all pra).  If, by default, it is printing all pra, then change default to just returning open pra.

### LINKED ISSUE #482: MNT: remove -s from l_projecta
the functionality is not working, but also it is now handled by ``-f status finished`` that is working

mwinitch:
Closes #480, #482

### DIFF
--- a/regolith/helpers/l_projectahelper.py
+++ b/regolith/helpers/l_projectahelper.py
@@ -21,7 +21,6 @@ from regolith.tools import (
 
 TARGET_COLL = "projecta"
 HELPER_TARGET = "l_projecta"
-ALLOWED_STATI = ["proposed", "started", "finished", "back_burner", "paused", "cancelled"]
 
 
 def subparser(subpi):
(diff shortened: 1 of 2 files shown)


mwinitch:
@sbillinge Once I add news I think it is ready for review and merge. However, is there anything you would like me to do with this helper. Any other flags or change anything else about it?

sbillinge:
> 
> 
> @sbillinge Once I add news I think it is ready for review and merge. However, is there anything you would like me to do with this helper. Any other flags or change anything else about it?

Let's release it like this and I can start to use it and give feedback then.  It is hard to change the inputs till I try to use it for something real.....then I figure out very quickly what I need.....   Thx!

PR #517, opened 2020-07-03
modified fragment retrieval  and changed to u_institution

dy2403:
Hi @sbillinge!
It should work properly now! It was only retrieving by id, and now I added aka and name. 
Also, do you think a verbose could be useful? I can try to add it if you wish!


sbillinge:
thanks @dy2403 .  It is an updater, not a lister, so I don't think we need -v.  However, we need to be able to preview the things we might want to update, including department and school if it exists.

PR #518, opened 2020-07-04
deleted mdy function from reimb builder

### LINKED ISSUE #514: MNT: move mdy function from reimb builder to dates and add tests

dy2403:
Closes #514 
Hi prof. @sbillinge,
I moved the mdy function from reimb builder to dates and added tests.
Can you please review it?

### DIFF
--- a/regolith/builders/reimbursementbuilder.py
+++ b/regolith/builders/reimbursementbuilder.py
@@ -3,9 +3,8 @@
 import os
 
 import openpyxl
-import datetime as dt
 from regolith.builders.basebuilder import BuilderBase
-from regolith.dates import get_dates, month_to_int
+from regolith.dates import get_dates
 from regolith.sorters import position_key
 from regolith.tools import all_docs_from_collection, month_and_year, \
     fuzzy_retrieval
--- a/regolith/builders/reimbursementbuilder.py
+++ b/regolith/builders/reimbursementbuilder.py
@@ -1,23 +1,14 @@
 """Builder for Resumes."""
 
-import datetime
 import os
 
 import openpyxl
 
 from regolith.builders.basebuilder import BuilderBase
-from regolith.dates import month_to_int, get_dates
+from regolith.dates import get_dates, mdy
 from regolith.sorters import position_key
 from regolith.tools import all_docs_from_collection, month_and_year, \
     fuzzy_retrieval
-
-
-
-def mdy(month, day, year, **kwargs):
-    return "{}/{}/{}".format(
-        str(month_to_int(month)).zfill(2), str(day).zfill(2), str(year)[-2:]
-    )
-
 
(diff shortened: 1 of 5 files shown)


regolith/dates.py
```diff
@@ -420,3 +420,28 @@ def is_between(thing, start=None, end=None):
...
+def mdy(month, day, year, **kwargs):
+    """
+    Given a month, day, and year, returns a string with the date
+    in the following format: MM/DD/YY
+
+    Parameters
+    ----------
+    month: integer or string
+         the month.  if a string should be resolvable using regolith month_to_int
+
```
sbillinge, line 433: remove empty lines.  If you want to see the format, please google for "numpy docstrings"

sbillinge:
Actuallly, come to think of it, I think we don't need this at all because if we are passing dates around we can use ``strftime`` method, so maybe just remove this function from everywhere it is used and replace with strftime

regolith/builders/reimbursementbuilder.py
```diff
@@ -118,9 +109,9 @@ def excel(self):
                             se_column = 14
                         dates.append(expdates.get("date"))
                         item_ws.cell(row=r, column=2, value=i)
-                        item_ws.cell(row=r, column=3, value=mdy(expdates.get("date").month,
-                                              expdates.get("date").day,
-                                              expdates.get("date").year))
+                        item_ws.cell(row=r, column=3, value=dt.datetime(int(expdates.get("date").year),
```
sbillinge, line 112: doesn't 
```
dt.datetime(expdates.get("date)).strftime('%x')
```
work?

sbillinge: also I see the same mdy function being used in coa builder.  Please could you do the same cleaning there as well?  Thx.

dy2403:
@sbillinge I am trying to understand what gets as input to this builder and how tests work to builders. Thus, I am not fully confident of what I am doing here.

regolith/builders/reimbursementbuilder.py
dy2403: I think the mdy function was defined there, but wasn't used. Thus, I just deleted it.

sbillinge: cool.  Remove WIP and request merge if you are ready.  Make sure there is a news.

sbillinge:
> 
> 
> @sbillinge I am trying to understand what gets as input to this builder and how tests work to builders. Thus, I am not fully confident of what I am doing here.
> Here, I am treating dates as date time objects, which I guess they are because min(dates) and max(dates) are used, is it correct?

yes, correct.

regolith/dates.py
dy2403: this was deleted!

sbillinge:
Thanks @dy2403

PR #523, opened 2020-07-05
WIP: Regolith Proposal Report Builder

vivianlin2000:


regolith/builders/grantreportbuilder.py
```diff
@@ -54,14 +54,21 @@ def latex(self):
...
+
         # Major Goals
 
         # Accomplishments
 
         # Opportunities for Training and Professional Development
         valid_presentations = []
-        for p in self.gtx["people"]:
-            if p["active"] and p["_id"] is not "sbillinge":
+        for p in people:
```
sbillinge, line 70: for person in people

regolith/builders/grantreportbuilder.py
```diff
@@ -54,14 +54,21 @@ def latex(self):
         begin_year, begin_month, begin_day = int(today[0])-1, int(today[1]), int(today[2])
         end_date = date(end_year, end_month, end_day)
         begin_date = date(begin_year, begin_month, begin_day)
+
+        # Get All Active Members
+        people = []
+        for p in self.gtx["people"]:
```
sbillinge, line 60: ``current_members = [person for person in self.gtx['people'] if person['active']]``

regolith/builders/grantreportbuilder.py
```diff
@@ -54,14 +54,21 @@ def latex(self):
...
         # Major Goals
 
         # Accomplishments
 
         # Opportunities for Training and Professional Development
         valid_presentations = []
-        for p in self.gtx["people"]:
-            if p["active"] and p["_id"] is not "sbillinge":
+        for p in people:
+            if p["_id"] is not "sbillinge":
```
sbillinge, line 71: my name shouldn't be hard-coded in here

regolith/builders/grantreportbuilder.py
```diff
@@ -70,6 +77,12 @@ def latex(self):
 
         # Plans for Next Reporting Period to Accomplish Goals
 
+        # Individuals that have worked on project
+        individuals_data = []
+        for p in people:
```
sbillinge, line 82: you keep looping over people so why not put each of these actions inside a single outer loop over people?

regolith/builders/grantreportbuilder.py
```diff
@@ -25,22 +25,10 @@ def construct_global_ctx(self):
...
-            all_docs_from_collection(rc.client, "presentations"), key=_id_key
-        )
-        gtx["grants"] = sorted(
-            all_docs_from_collection(rc.client, "grants"), key=_id_key
-        )
-        gtx["institutions"] = sorted(
-            all_docs_from_collection(rc.client, "institutions"), key=_id_key
-        )
-
+        for dbs in rc.needed_dbs:
```
sbillinge, line 28: so pretty!  👍

regolith/builders/grantreportbuilder.py
```diff
@@ -51,36 +39,31 @@ def latex(self):
...
+        begin_year, begin_month, begin_day = int(today[0]) - 1, int(today[1]), int(today[2])
         end_date = date(end_year, end_month, end_day)
         begin_date = date(begin_year, begin_month, begin_day)
 
         # Get All Active Members
-        people = []
-        for p in self.gtx["people"]:
-            if p["active"]:
-                people.append(p)
+        current_members = [person for person in self.gtx['people'] if person['active']]
```
sbillinge, line 47: :) nice!

regolith/builders/grantreportbuilder.py
```diff
@@ -36,12 +43,23 @@ def construct_global_ctx(self):
...
     def latex(self):
         """Render latex template"""
+        rc = self.rc
         # Get Dates
-        today = str(date.today()).split("-")
-        end_year, end_month, end_day = int(today[0]), int(today[1]), int(today[2])
-        begin_year, begin_month, begin_day = int(today[0]) - 1, int(today[1]), int(today[2])
-        end_date = date(end_year, end_month, end_day)
-        begin_date = date(begin_year, begin_month, begin_day)
+        start_date = rc.start_date.split("-")
```
sbillinge, line 48: this is heading us back down a road to the old days (passing dates around as strings and ints) rather than the brave new world of working with dates.   I very much doubt this is the direction we want to go in here....

vivianlin2000: Yea I realized this too! Will work with date objects instead of the strings and ints.

regolith/builders/grantreportbuilder.py
```diff
@@ -45,19 +45,20 @@ def latex(self):
         """Render latex template"""
         rc = self.rc
         # Get Dates
-        start_date = rc.start_date.split("-")
-        end_date = rc.end_date.split("-")
-        begin_year, begin_month, begin_day = int(start_date[0]), int(start_date[1]), int(start_date[2])
-        end_year, end_month, end_day = int(end_date[0]), int(end_date[1]), int(end_date[2])
-        end_date = datetime.datetime(end_year, end_month, end_day)
-        begin_date = datetime.datetime(begin_year, begin_month, begin_day)
+        start_date = datetime.strptime(rc.start_date, "%Y-%m-%d")
```
sbillinge, line 48: we usually use a date parser from dateutils.   pls could you look in one of the other helpers for the syntax?

regolith/builders/grantreportbuilder.py
```diff
@@ -45,19 +45,20 @@ def latex(self):
...
         grant_name = "dmref"
-        # Get people linked to grant and active during reporting period
-        people = [person for person in self.gtx['people']
-                  if grant_name in person['grant']]
-        # Get prums linked to grant and active/finished during the reporting period
+        # Get people and prum linked to grant and active during reporting period
+        prums = [prum for prum in self.gtx['projecta']
+                 if grant_name in prum['grants']
+                 and
+                 ((start_date <= datetime.strptime(prum['end_date'], "%Y-%m-%d") <= end_date
```
sbillinge, line 57: use get_dates to get dates and is_current etc. (fully tested functions in tools.py to do this work if poss.

regolith/templates/grantreport.txt
```diff
@@ -28,17 +27,39 @@ Goal 4: Apply this to real scientific problems
 
 * What was accomplished under these goals (you must provide information for
 at least one of the 4 categories below)?
-Major Activities:
-Specific Objectives:
-Significant Results:
+------------------------------------------------------------------------------
+- Major Activities (currently worked on projecta):
```
sbillinge, line 31: again, can't have me hard-coded in here.

If there is something that is "me" specific that doesn't change often it should probably be in people.  However, since this is a grant-specific thing it might better be in grants?   If it is mostly me-specific but will be described in the grant report of multiple grants, this should probably be in people but have a field that is ``grant``.  This is how we discover how the world is organized!

regolith/templates/grantreport.txt
```diff
@@ -28,17 +27,39 @@ Goal 4: Apply this to real scientific problems
...
-{%- for opportunity in presentations %}
-    - {{opportunity -}}
+{%- for opportunity in trainingAndProfessionalDevelopment %}
+    - Authors: {{ opportunity['authors'] }}
+    - Title: {{ opportunity['title'] }}
+    - Location: {{ opportunity['meeting_name'] }}
 {% endfor -%}}
 
 * How have the results been disseminated to communities of interest?
+1. Publications
```
sbillinge, line 60: This is an interesting one.   It should probably be in people too so any user of Regolith can decide for themselves.

regolith/builders/grantreportbuilder.py
```diff
@@ -90,13 +91,32 @@ def latex(self):
...
         # thesis defendings
         # how do i access people.yml in rg-db-public vs the people.yml file in rg-db-group?
+        defended_theses = []
         for id in grant_people:
             person = self.gtx['people'][id]
-            if person['']
-        # How have results been disseminated
+            for education in person['education']:
+                if 'phd' in education['degree'].lower() and 'columbia' in education['institution'].lower() and \
+                        rp_start_date.year <= education['end_year'] <= rp_end_date.year:
```
sbillinge, line 99: since you will be doing something like this a lot I guess, you may want to use a function.  Can you use is_current which has already been written?  I would probably try and use get_dates to get dates and then is_current.

It may not work, but using functions that are well tested and can handle edge cases well will save time later.

regolith/builders/grantreportbuilder.py
```diff
@@ -90,13 +91,32 @@ def latex(self):
...
 
-        # Plans for Next Reporting Period to Accomplish Goals
+        # Products
+        # need rg-db-public's citation.yml... how to access that instead of the rg-db-group's citation.yml file
+        publications = filter_publications(self.gtx["citations"], grant_people, since=rp_start_date, before=rp_end_date)
 
+        # Participants/Organizations
+        participants = {}
+        for person in grant_people:
+            p = self.gtx["people"].get(person)
```
sbillinge, line 109: please don't be lazy with variable names, it makes the code hard to read.  What is p?

Here I think the logic is that p contains the person document for the person of name person, so I would pick variable names like this:
```
for name in grant_people:
    person = self.gtx["people"].get(person)
```
I think this logic assumes that the names in grant_people are valid id's is that right?  This might be a bit brittle.

regolith/builders/grantreportbuilder.py
```diff
@@ -90,13 +91,32 @@ def latex(self):
...
+            months = 0
+            if p['active']:
+                difference = datetime.today() - rp_start_date
```
sbillinge, line 112: is it correct to do this calculation from today?

regolith/templates/grantreport.txt
```diff
@@ -1,4 +1,4 @@
-NSF {{beginYear}}-{{endYear}} Annual Report by Simon J. L. Billinge, (Columbia University)
+NSF {{beginYear}}-{{endYear}} Annual Report
```
sbillinge, line 1: Don't you want a name here (just not my name hard-coded)

tests/outputs/grantreport/billinge_grant_report.txt
```diff
@@ -51,32 +51,17 @@ Key outcomes or Other achievements:
 3. Presentations at conferences and seminars
 
 * What do you plan to do during the next reporting period to accomplish the goals?
+#############################################################
```
sbillinge, line 54: this looks like copy-pasted text....maybe assign this string to a variable and use that?

regolith/builders/grantreportbuilder.py
```diff
@@ -23,6 +23,7 @@
 
 
 def subparser(subpi):
+    subpi.add_argument("authorFullName", help="full name of the author of this grant report")
```
sbillinge, line 26: this is very brittle.  I doubt I will ever remember the full name of people and whether it is capitalized and spaces and whatnot.  We have two choices here.
1. require the person id (e.g., you would be vlin).  This is easier to remember because we use a mostly standard naming scheme.
1. allow anything that is in id, name or alias and then use fuzzy retrieval to get the person back.

(2) is probably better because (1) is also covered by (2)

regolith/builders/grantreportbuilder.py
```diff
@@ -100,32 +101,78 @@ def latex(self):
...
-            p = self.gtx["people"].get(person)
-            months = 0
-            if p['active']:
-                difference = datetime.today() - rp_start_date
-                if difference.day > 15:
-                    months = months + 1
-                months = difference.year * 12 + difference.month
-            else:
+        for name in grant_people:
+            person = self.gtx["people"].get(name)
```
sbillinge, line 110: to get things from a collection we use things like `fuzzy_retrieval` in tools.py or `find_one` etc. in `fsclient` and `mongoclient` You might also find `get_person` from tools.py to be useful.

PR #527, opened 2020-07-06
u_contact issue 524

dy2403:
Hi @sbillinge,

I made changes, but I am still unsure of what I should do with email. 
I don't really see how the email validator should behave, especially bc it is called even when there is no email (and it is an optional field). Also, it does not seem practical to have email url, cred, etc., specified in the command line when adding a new email. I guess it should be good to have a validator to an email address, but maybe we should move them somewhere else? 

Screen Shot 2020-07-06 at 1 21 37 PM


dy2403:
@sbillinge ready to review and merge!

Issue #513, opened 2020-07-03
MNT: help simon get reading_lists PR #291 merged


sbillinge:
Simon needs help getting this PR merged.  Please reach out to Simon about how to proceed.

jiaweizang:
@sbillinge Hi Simon! Is it a rebase problem? I can fork that branch and create a new PR to merge.

sbillinge:
OK, fine.

PR #532, opened 2020-07-07
Todo adder ready to be merged

jiaweizang:
Hi @sbillinge, I create this new PR in order not to rebase.

### DIFF
--- a/tests/outputs/a_todo/people.yaml
+++ b/tests/outputs/a_todo/people.yaml
@@ -1,0 +1,526 @@
+sbillinge:
+  activities:
+    - type: teaching
+      name: course development
+      year: 2018
+      other: Developed a new course for Materials Science
+  aka:
+    - Billinge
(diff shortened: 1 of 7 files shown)


tests/outputs/a_todo/people.yaml
```diff
@@ -0,0 +1,526 @@
...
+      month: August
+      organization: Columbia University
+      position: professor
+      semester: Fall
+      syllabus: https://drive.google.com/open?id=0BxUpd34yizZreDBCMEJNY2FUbnc
+      year: 2019
+  title: Dr.
+  todos:
+    - description: read paper
+      due_date: '2020-07-19'
```
jiaweizang, line 270: I keep this as string type, because the first two tasks are from schemas. In the exemplars of schemas, it seems that the date can only be written with quotation mark, or it won't pass the test.

sbillinge:
Thanks @jiaweizang I will start to play with it!

I have a project for you that we can talk about.  I will send a prum.  But in the meantime, please could you work on issue #513

PR #533, opened 2020-07-07
WIP: Sm_reading_list

jiaweizang:
Hi @sbillinge, I checked out a new branch, then merged sbillinge/sm_reading_lists to that new branch. Is there other things to do for this issue?


tests/outputs/reading-lists/rlistbibfile.txt
```diff
@@ -0,0 +1,10 @@
+Reading List
```
sbillinge, line 1: something wrong here.... this looks like a template but it should look like the built file, so the builder isn't building anything in your test.  Same for the bib file above.

jiaweizang: Before it didn't pass the test, saying there aren't output files, so I try adding a template to the outputs, and suddenly it pass...Yeah the order I did is wrong, I will modify it soon!

sbillinge: I think it is probably just that you added an outputs file.  Does it still pass if you remove the template files but leave behind an empty outputs directory!

Please let me reiterate frequently and often, the purpose of tests is not to make them pass, it is to make them ensure the code is behaving as expected....... always remember that and design them with that in mind.  If it were oherwise all our tests would look like 
```
def test_something()
    assert True
```
and we would always pass tests!

jiaweizang: Thanks! It turns out that the test program will check the output file only when the file name is the same as the generated output file. 
Please see the new code. The format of the new outputs file is the same as that you sent me during on boarding.

tests/outputs/a_grppub_readlist/reading_lists.yaml
```diff
@@ -15,7 +15,7 @@ getting_started_with_pdf:
   year: 2019
 test_the_lister:
   papers:
-    doi: 10.1021/nn501591g
-    text: The description of symbolic computing in Python
```
jiaweizang, line 19: I also modified a_grppub_readlist helper a little bit, because previously, the format here wasn't list type, and it conflicted with the new schemas.

regolith/builders/readinglistsbuilder.py
```diff
@@ -0,0 +1,123 @@
...
+    def latex(self):
+        """Render latex template"""
+
+        for rlist in self.gtx["reading_lists"]:
+            listid = rlist["_id"]
+            outfile_bib = listid
+
+            n = 1
+            for paper in rlist['papers']:
+                # fixme: code here to get info from crossref
```
jiaweizang, line 55: not sure what information should be got from crossref...

jiaweizang, line 55: not sure what information should be got from crossref...

tests/outputs/a_grppub_readlist/reading_lists.yaml
sbillinge: good!

regolith/builders/readinglistsbuilder.py
sbillinge: We get the citation info from crossref.  But that should already be in there.  I will look at your whole file changes to see what is going on.

sbillinge: OK, I see.  The code is below (starting ``article = self.cr.works(ids=doi)``

We can just delete this comment.  Also the commented out code below with the ``pass`` can be deleted.

regolith/builders/readinglistsbuilder.py
```diff
@@ -0,0 +1,123 @@
...
+                    n += 1
+
+            self.render(
+                "rlistbibfile.txt",
+                outfile_bib + ".txt",
+                rlist=rlist,
+            )
+
+
+"""            self.render(
```
sbillinge, line 118: let's delete this too.

sbillinge:
@jiaweizang I looked back through this and I am worried about the merge way back in the beginning.

Please could we close this PR.  I will rebase my original PR and then could you make PRs into that and we will merge that?  Thx.

jiaweizang:
@sbillinge Sure!

PR #537, opened 2020-07-07
makeappointments run bug

adibaejaz:
made a `run` required argument for the `makeappointments` helper to work as in `l_contacts` after noticing a bug while trying it in my local!


sbillinge:
Nice, thanks @adibaejaz !

PR #538, opened 2020-07-07
TST: use word by word comparison to fix the file path issue

st3107:
Hi @sbillinge @scopatz . I thought out a solution to fix the file path issue in test builders. I use the word by word comparison between two strings and ignore the comparison if certain substring is contained. How do you think of it?


sbillinge:
@st3107 this looks good to me.  I merged, let's try it out....

PR #539, opened 2020-07-07
changed output (regular and verbose) l_contacts

### LINKED ISSUE #536: ENH: contact lister
Hi Daniela, this is getting good now.

Next, please could you 
1. return the id as well as name etc. in the short form.  
1. in the short form, put it all info on one line, maybe separated by | or whatever looks good.  No need to prefix with ``id:``, it should be pretty self evident what the info means.
1. in the verbose form, the formatting looks great, but if a field is empty, simply don't return the field at all, except perhaps for institution where you can put None.
1. Dereference the full name of the institution from institutions.yml (use fuzzy retrieval).  If the institution can't be found, return whatever  is in the institution field as now.

OK, that's it for now.

dy2403:
Closes #536 
Hi @sbillinge!
I made changes to the l_contacts. My changes were mostly done to the function stringify, which I am not sure if it is optimal. However, to make it differently, I think I would need to make big changes to the structure of the helper, because right now it is using this function to get all the data and transform to a string, so I would have to change it back to a dictionary to work on it. 
Please let me know what your thoughts are on this! I would be happy to make it differently if you wish!

### DIFF
--- a/regolith/helpers/l_contactshelper.py
+++ b/regolith/helpers/l_contactshelper.py
@@ -13,7 +13,7 @@ from regolith.tools import (
     all_docs_from_collection,
     get_pi_id,
     fragment_retrieval,
-    search_collection,
+    fuzzy_retrieval,
     key_value_pair_filter,
     collection_str
 )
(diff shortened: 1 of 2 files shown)


sbillinge:
> Closes #536
> Hi @sbillinge!
> I made changes to the l_contacts. My changes were mostly done to the function stringify, which I am not sure if it is optimal. However, to make it differently, I think I would need to make big changes to the structure of the helper, because right now it is using this function to get all the data and transform to a string, so I would have to change it back to a dictionary to work on it.
> Please let me know what your thoughts are on this! I would be happy to make it differently if you wish!

yes, this is almost certainly a bad way to do it.  I didn't really like it when Michael used this but I let it slide.  it is much better to pass around documents (dicts) and lists of documents, and then build hte strings at the end.  Could you try and refactor and remove stringify altogether?  Let me know if you run into problems, but I don't think you will.

dy2403:
hi @sbillinge!
I made changes, and deleted that function. Can you please review it?

PR #542, opened 2020-07-08
AddedUserNameReq

zthatch:
Oddly I am getting an error on the html builder, but I don't think that has anything to do with this change. This has to do with our previously discussed change to require that the user have some sort of identifier to be automatically inserted into the commit message for the regolith heap


regolith/storage.py
```diff
@@ -88,18 +88,19 @@ def push_git(store, path, rc):
     subprocess.check_call(cmd, cwd=storedir)
     if "default_user_id" in rc:
         cmd = ["git", "commit", "-m", f"{rc.default_user_id} regolith auto-store commit"]
```
sbillinge, line 90: see my comment on the other PR.  It looks as we may need ``default_github_id`` in regolithrc.json

PR #543, opened 2020-07-08
added webinar as an allowed type to presentations

### LINKED ISSUE #456: webinar as allowed presentation type
update schema, exemplar and preslist builder etc. to handle the new post-covid world

dy2403:
Closes #456 
Hi @sbillinge!
Can you please review and merge this?

### DIFF
--- a/regolith/schemas.py
+++ b/regolith/schemas.py
@@ -3252,6 +3252,7 @@ SCHEMAS = {
                 "poster",
                 "seminar",
                 "tutorial",
+                "webinae"
             ],
             "required": True,
             "type": "string",
--- a/regolith/schemas.py
+++ b/regolith/schemas.py
@@ -3252,7 +3252,7 @@ SCHEMAS = {
                 "poster",
                 "seminar",
                 "tutorial",
-                "webinae"
+                "webinar"
             ],
             "required": True,
             "type": "string",
--- a/regolith/schemas.py
+++ b/regolith/schemas.py
@@ -8,7 +8,7 @@ from .sorters import POSITION_LEVELS
 
 SORTED_POSITION = sorted(POSITION_LEVELS.keys(), key=POSITION_LEVELS.get)
 PRESENTATIONS_TYPE = ["award", "colloquium", "contributed_oral", "invited", "keynote",
-                      "plenary", "poster", "seminar", "tutorial", "webinar"]
+                      "plenary", "poster", "seminar", "tutorial"]
 EXEMPLARS = {
     "abstracts": {
         "_id": "Mouginot.Model",
@@ -1224,6 +1224,7 @@ EXEMPLARS = {
             "status": "accepted",
             "title": "Graphitic Dephenestration",
             "type": "award",
+            "webinar": False,
         },
         {
             "_id": "18sb_nslsii",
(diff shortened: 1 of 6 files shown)


regolith/schemas.py
```diff
@@ -3252,6 +3252,7 @@
                 "poster",
                 "seminar",
                 "tutorial",
+                "webinar"
```
sbillinge, line 3255: since we are touching this code now, I want to change how we do these eallowed by defining a global variable at the top of the schemas file and then insert it here.  the advantage of this is that we can import this into helpers to get the allowed lists that are synchronized.   We could call it by default ``<COLLECTION>_<FIELD1>_<FIELD2>`` etc., so this one would be called ``PRESENTATIONS_TYPE`` and it would be set to this list at the top of the file.

regolith/templates/preslist.tex
```diff
@@ -20,7 +20,7 @@
 \item {{ latex_safe(pres['authors'])-}}, {\bf {{ latex_safe(
 sentencecase(pres['title']))-}} },
 {%-set inst = pres['institution'] %}
-{%-if pres['type'] in ['colloquium', 'seminar']%}
+{%-if pres['type'] in ['colloquium', 'seminar', 'webinar']%}
```
sbillinge, line 23: I guess webinar could be a seminar or a conference presentation.  I think we will have to handle it separately.  

Actually, come to think of it, it is not really a type of presentation, but a mode of delivering it, so I guess we may need a new field called ``webinar:`` that is False by default but can be set to true.    If it is true we just change some text somewhere to indicate this?  Could you play with it a bit and see if something makes sense?  Take a look at my presentation list (I will send it) and see what you think it should read for a webinar, then try to code it up.

regolith/schemas.py
dy2403: done!

regolith/templates/preslist.tex
dy2403: I tried! Can you look at this?

regolith/schemas.py
```diff
@@ -7,7 +7,8 @@
 from .sorters import POSITION_LEVELS
 
 SORTED_POSITION = sorted(POSITION_LEVELS.keys(), key=POSITION_LEVELS.get)
-
+PRESENTATIONS_TYPE = ["award", "colloquium", "contributed_oral", "invited", "keynote",
```
sbillinge, line 10: this is now not correct.   webinar is not a valid type in presentations.

regolith/schemas.py
```diff
@@ -3242,21 +3244,16 @@
...
-            ],
+            "eallowed": PRESENTATIONS_TYPE,
             "required": True,
             "type": "string",
         },
+        "webinar": {
+            "description": "true if a webinar",
+            "required": False,
+            "type": "boolean",
+            "default": False,
```
sbillinge, line 3255: does this work, setting default in the schema?  Actually, I don't know the answer, this is not a rhetorical question.....

regolith/schemas.py
dy2403: Done!

regolith/schemas.py
dy2403: I am not sure, but I think I do not even need this.
 In the code, I am only adding the webinar if it is true, thus it is not necessary to set it to false by default, bc if it is none or false, it will do nothing. Or should I guarantee that when there is nothing in the 'webinar' field it should be defined as false?

sbillinge: no, you are right, just remove this.

sbillinge:
wow, nice.....merging all over the place for you today!

PR #561, opened 2020-07-16
amount type from int to float

dragonyanglong:
@sbillinge , please review and merge.


sbillinge:
thanks @dragonyanglong !

PR #568, opened 2020-07-22
add u_todo

jiaweizang:


sbillinge:
Thanks @jiaweizang  I am just working with the todo infrastructure and was just thinking "I need the todo updater!"

I have some comments on things that I need worked on based on my usage.  I will send these shortly, plus any comments I have on the updater.....   thanks so much!

PR #573, opened 2020-07-24
fix some bugs and add order algorithm in l_todo

jiaweizang:
Hi @sbillinge! I haven't sync printed lists in l_todo, f_todo and u_todo. I will do it soon after we decide which way to choose (please see comments in issue#572). 
In this PR, there are several things:
1. delete begin-date and end-date in u_todo. 
2. add flag --end_date in f_todo, cuz when we use f_todo to mark a task as finished, I think it's better to set end_date as today automatically, or specify an end_date.
3. add an algorithm in l_todo to sort tasks. I tried:  order = importance + 1/[1 + exp(|days to due date|)]. The second part  "1/[1 + exp(|days to due date|)] " ranges from 0 to 0.5. So by using that algorithm, importance 2 will always stay top, importance 1 is at the middle, and 0 is at the bottom. In each importance level, the tasks will be sorted by the absolute value of "days to due date", and list from small to large.  We can try this first and discuss if there are other things to consider.

### DIFF
--- a/regolith/helpers/f_todohelper.py
+++ b/regolith/helpers/f_todohelper.py
@@ -22,6 +22,8 @@ def subparser(subpi):
     subpi.add_argument("-i", "--index",
                         help="Index of the item in the enumerated list to mark as finished.",
                         type = int)
+    subpi.add_argument("-e", "--end_date",
+                       help="End date of the task. Default is today.")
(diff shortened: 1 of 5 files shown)


sbillinge:
super, thanks @jiaweizang

PR #575, opened 2020-07-24
update chainDB

jiaweizang:
Previously, if 
z =ChainDB [
                      {"a": {"b": 1}},  
                      {"a": {"c": [-1,-2]}}, 
                      {"a": {"c": [-3,-4]}}
 ] 
And when we want to get z["a"]["c"]: 
results = [__main__.ChainDBSingleton object, [-1,-2], [-3,-4]], and it will return [-3,-4].

By updating the codes,
results = [[-1,-2], [-3,-4]], and it will return [-1,-2,-3,-4].

### DIFF
--- a/tests/test_chain_db.py
+++ b/tests/test_chain_db.py
@@ -104,7 +104,7 @@ def test_exactness_setting_multi2():
     assert isinstance(z["a"]["m"], list)
     assert isinstance(z["a"].maps[0], dict)
     assert g is z["a"].maps[1]["mm"]
-    assert z["a"]["mm"] is g
+    assert z["a"]["mm"] == g
     assert z["a"]["m"] == [1, 2, 3, 4]
 
 
@@ -121,7 +121,7 @@ def test_exactness_setting_multi_novel():
     assert isinstance(z["a"]["m"], list)
     assert isinstance(z["a"].maps[0], dict)
     assert g is z["a"].maps[0]["mm"]
-    assert g is z["a"]["mm"]
+    assert g == z["a"]["mm"]
 
 
 def test_dicts_in_lists():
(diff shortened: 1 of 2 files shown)


tests/test_chain_db.py
```diff
@@ -104,7 +104,7 @@ def test_exactness_setting_multi2():
     assert isinstance(z["a"]["m"], list)
     assert isinstance(z["a"].maps[0], dict)
     assert g is z["a"].maps[1]["mm"]
-    assert z["a"]["mm"] is g
+    assert z["a"]["mm"] == g
```
jiaweizang, line 107: Before, when we want to get z["a"]["mm"], chainDB will gather values, and results = [singleton, g], where g = [-1, -2]. Because not all the items are list type, it will return to the last value g, so z["a"][mm"] is g.

Now, after removing singleton,  results =  [g]. The file will tell that all the items in "results" are list, and it will merge all the items. It still returns to [-1, -2], because there is only "g" in the results, but z["a"][mm"] is not g, instead, it is a list that merge all the items, even though it still equals to [-1,-2]

sbillinge: @CJ-Wright @scopatz please could you look at this?  We found what we think is a bug in chainDB that if the collection is in 3 dbs and a key has a list-value in two but nothing in the other it doesn't return the merged list, which I think should be desired behavior (I certainly desire it!}.  this is a fix for that.  But let us know if you think it will introduce any unintended consequences (all tests are passing btw)

sbillinge:
@scopatz @CJ-Wright I am going to merge this as we have tested and it seems to be working.  I tis a nice granular PR so we can revert it if it causes issues elsewhere, but I don't think it will, and it is passing all tests.

PR #577, opened 2020-07-24
WIP: makeappts bug fixes

adibaejaz:


adibaejaz:
@sbillinge plots should show now - I think the reason they were not displaying is because you replaced `plt.show()`  with `plt.draw()`. from some googling, it appears `plt.draw()` is called after `plt.show()` in case we have made some changes to the graph in interactive mode and want to redraw it accordingly, whereas  `plt.show()` is used to display the figure

sbillinge:
> 
> 
> @sbillinge plots should show now - I think the reason they were not displaying is because you replaced `plt.show()` with `plt.draw()`. from some googling, it appears `plt.draw()` is called after `plt.show()` in case we have made some changes to the graph in interactive mode and want to redraw it accordingly, whereas `plt.show()` is used to display the figure

please test it carefully.  I replaced show with draw because show wasn't working.  You have basically put it back to the version that wasn't working for me.  The issue is multiple grants.  Do you have a test repo for that case.  If it is working for your repo but not mine we can see if it is an issue with my repo, but I think my repo is ok.  Let me know when you have tested that and it passes.

PR #579, opened 2020-07-24
added --current, fixed --all

sheneric88:
@sbillinge

So now --current shows everything ongoing (proposed or started), and --all now shows every projecta available. Though it looks like nothing is marked as ended so they both show every projecta for now.

Should any overall projecta be marked finished or anything? Or just certain prums? (assuming I'm understanding prums are the small parts of a bigger projecta)

I guess is this what l_projecta should do in general or something else?

### DIFF
--- a/regolith/helpers/l_projectahelper.py
+++ b/regolith/helpers/l_projectahelper.py
@@ -25,9 +25,14 @@ HELPER_TARGET = "l_projecta"
 
 def subparser(subpi):
     subpi.add_argument("--all", action="store_true",
-                       help="Lists all projecta that have not ended"
+                       help="Lists all projecta in general"
                        )
-    subpi.add_argument("-v", "--verbose", action="store_true", help='increase verbosity of output')
+    subpi.add_argument("-c", "--current", action="store_true",
+                       help='List ongoing projecta'
+                       )
+    subpi.add_argument("-v", "--verbose", action="store_true",
+                       help='Increase verbosity of output'
+                       )
     subpi.add_argument("-l", "--lead",
                        help="Filter milestones for this project lead"
                        )
@@ -102,7 +107,7 @@ class ProjectaListerHelper(SoutHelperBase):
         else:
             collection = self.gtx["projecta"]
 
-        if (not rc.lead) and (not rc.person) and (not rc.ended) and (not rc.grant) and (not rc.verbose) and (not rc.grp_by_lead) and (not rc.filter) and (not rc.all):
+        if (not rc.lead) and (not rc.person) and (not rc.ended) and (not rc.grant) and (not rc.verbose) and (not rc.grp_by_lead) and (not rc.filter) and (not rc.all) and (not rc.current):
             return
         if rc.date:
             desired_date = date_parser.parse(rc.date).date()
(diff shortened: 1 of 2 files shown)


sbillinge:
thans @sheneric88 .  Who is helping you with the rebases, because whatever you are doing, it is probably not a rebase....which results in you having to do a ``git push --force-with-lease``...

A prum is a projectum (the whole thing).  The smaller things inside the projecta are milestones.  There definitely are finished prums in the group database, but there may not be in the schema against which we run tests.

Otherwise, this looks good.  is it ready to merge?

sheneric88:
I followed a few guides online, but it looks like the rebase method wasn't the right one.

Hmmm, I see the finished prums, but `-e` doesn't output them. There's multiple `if rc.ended` functions that might be causing something?

Besides that `--current` and `--all` run fine, so yep ready to merge

regolith/helpers/l_projectahelper.py
```diff
@@ -118,13 +123,17 @@ def sout(self):
...
         for projectum in collection:
-            if rc.all and projectum.get('status') != "finished":
-                projecta.append(projectum)
-                continue
             if isinstance(projectum.get('group_members'), str):
                 projectum['group_members'] = [projectum.get('group_members')]
+            if rc.all and (projectum.get('status') == True):
+                projecta.append(projectum)
+                continue
+            if rc.current and (projectum.get('status') == "proposed" or "started"):
```
sbillinge, line 134: I think this will be hard to maintain.  I suggest that we make lists at the top with relevant things in them such as 
```
ACTIVE_STATI = ["proposed", "started"]
INACTIVE_STATI = [...]
```
and then say 
```
if rc.current and projectum.get('status') in ACTIVE_STATI:
...
```
Also, no need for those parens which have a different meaning in coding than in math (you just made a tuple there but luckily it has the desired behavior).   
Rather than littering our code with these tests, I think it is better if we make the projectum status to be required in the schema, so I prefer if you remove the ``projectum.get('status') == True`` tests everywhere (in this case the code crashing is a feature as it forces us to make the database compliant)

regolith/helpers/l_projectahelper.py
```diff
@@ -118,13 +123,17 @@ def sout(self):
...
+                projecta.append(projectum)
+                continue
```
sbillinge, line 136: doing it this way will mean that none of the other filters will work.  Is that the desired behavior?  I don't remember the discussion.

tests/test_helpers.py
```diff
@@ -53,6 +53,9 @@
     (["helper", "l_projecta", "--all"],
      "20ly_newprojectum\n20sb_firstprojectum\n"
```
sbillinge, line 54: it looks as if we are not properly testing this.  We need a prum in the schema exemplar that is not current to properly test the functionality.  I guess ideally we would have prums in there that have each status to make sure the filters work properly.  This will probably cause all kinds of tests to fail on other helpers, but I think it is worth putting in the effort now to get the tests good.  rn they are not really testing much of the desired behaviors.

Actually, the first step may be for us to really decide what we want the behavior to be, make UCs then code up the tests, then finally work on the code....proper TDD

PR #580, opened 2020-07-27
add running_index

jiaweizang:
@sbillinge Hi Simon! Please check the new feature. 

Now in u_todo and f_todo, user can update tasks from all databases by entering running_index.

### DIFF
--- a/regolith/helpers/f_todohelper.py
+++ b/regolith/helpers/f_todohelper.py
@@ -5,6 +5,7 @@ import datetime as dt
 import dateutil.parser as date_parser
 from dateutil.relativedelta import relativedelta
 import sys
+import math
 
 from regolith.helpers.basehelper import DbHelperBase
 from regolith.fsclient import _id_key
@@ -25,6 +26,8 @@ def subparser(subpi):
     subpi.add_argument("-e", "--end_date",
                        help="End date of the task. Default is today.")
     subpi.add_argument("-a", "--assigned_to", help="ID of the member to whom the task is assigned. Default id is saved in user.json. ")
+    subpi.add_argument("-c", "--certain_date",
+                       help="Enter a certain date so that the helper can calculate how many days are left from that date to the deadline. Default is today.")
     return subpi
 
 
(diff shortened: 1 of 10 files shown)


regolith/helpers/f_todohelper.py
```diff
@@ -63,47 +66,63 @@ def db_updater(self):
...
+            print("started:")
+            num = 1
             for todo in todolist:
-                if todo.get('status') not in ["finished"]:
-                    print(f"{todo.get('index')}. {todo.get('description')}")
-                    del todo['index']
-            return
+                if todo.get('status') == "started":
+                    print(
+                        f"{num:>2}. ({todo.get('running_index')}) {todo.get('description')}({todo.get('days_to_due')}|{todo.get('importance')}|{str(todo.get('duration'))})")
```
sbillinge, line 99: Is this printing two indices?  num and running_index.  This will be confusing, so I suggest to just print running index.

Also, I think I asked for it before but please put a space between the description and the info in the parens.

maybe also ``.strip()`` the description to remove any rogue whitespace?

regolith/helpers/f_todohelper.py
```diff
@@ -63,47 +66,63 @@ def db_updater(self):
...
+                if not todo.get('importance'):
+                    todo['importance'] = 1
+                if type(todo["due_date"]) == str:
+                    todo["due_date"] = date_parser.parse(todo["due_date"]).date()
+                todo["days_to_due"] = (todo.get('due_date') - today).days
+                todo["order"] = todo['importance'] + 1 / (1 + math.exp(abs(todo["days_to_due"])))
+            todolist = sorted(todolist, key=lambda k: (-k['order'], k.get('duration', 10000)))
+            print("-" * 50)
             print("Please choose from one of the following to update:")
+            print("  (running_index) action (days to due date|importance|expected duration(mins))")
```
sbillinge, line 93: in English we always put space between words and parens but no space inside parens, so (this is good) but(this is not good) and ( neither is is this ).

Can we put this informational line above the    ``print("-" * 50)`` line.  My eyes have found this confusing having it mixed iwth the actual data.

We may want another informational print statement above this that says, "if the list numbers are far from being in numerical order, please reorder them by running ????"

regolith/helpers/l_todohelper.py
```diff
@@ -117,31 +116,45 @@ def sout(self):
...
             for todo in gather_todos[::-1]:
                 if todo.get('duration') is None or float(todo.get('duration')) > float(rc.short_tasks):
-                    gather_todos.remove(t)
-        num = 1
+                    gather_todos.remove(todo)
         print("-" * 50)
-        print("    action (days to due date|importance|expected duration(mins))")
-        for todo in gather_todos:
-            if todo.get('status') not in ["finished", "cancelled"]:
+        print("  (running_index) action (days to due date|importance|expected duration(mins))")
```
sbillinge, line 126: since we are reusing code in different places, it makes sense to break it out into a function or sthg and call that function in each place.  Much easier to maintain, and ensures the same behavior in each place.

regolith/helpers/u_todohelper.py
```diff
@@ -96,51 +103,81 @@ def db_updater(self):
...
+                db_name = rc.databases[i]["name"]
+                person_idx = rc.client.find_one(db_name, rc.coll, filterid)
+                todolist_idx = person_idx.get("todos", [])
+                if len(todolist_idx) == 0:
+                    continue
+                else:
+                    for todo in todolist_idx:
+                        if todo.get('status') == "started":
+                            todo["running_index"] = index
+                            index += 1
```
jiaweizang, line 119: First, give each task a temporary running_index that will be used to identify each task later

regolith/helpers/u_todohelper.py
```diff
@@ -96,51 +103,81 @@ def db_updater(self):
...
+            else:
+                today = date_parser.parse(rc.certain_date).date()
+            for todo in todolist:
+                if not todo.get('importance'):
+                    todo['importance'] = 1
+                if type(todo["due_date"]) == str:
+                    todo["due_date"] = date_parser.parse(todo["due_date"]).date()
+                todo["days_to_due"] = (todo.get('due_date') - today).days
+                todo["order"] = todo['importance'] + 1 / (1 + math.exp(abs(todo["days_to_due"])))
+            todolist = sorted(todolist, key=lambda k: (-k['order'], k.get('duration', 10000), k['status']))
```
jiaweizang, line 142: Second, mix all tasks and sort them

regolith/helpers/u_todohelper.py
```diff
@@ -96,51 +103,81 @@ def db_updater(self):
...
+                for i in range(0, len(rc.databases)):
+                    db_name = rc.databases[i]["name"]
+                    person_idx = rc.client.find_one(db_name, rc.coll, filterid)
+                    todolist_idx = person_idx.get("todos", [])
+                    if len(todolist_idx) != 0:
+                        for todo in todolist_idx:
+                            if todo.get('status') == "started":
+                                index = index_match[todo["running_index"]]
+                                todo["running_index"] = index
+                        rc.client.update_one(db_name, rc.coll, {'_id': rc.assigned_to}, {"todos": todolist_idx},
```
jiaweizang, line 159: Then, pair the temporary index with the updated order, and saved them in index_match. 
Finally, update tasks again..

The logic looks complex, while I couldn't think up an easier one... At least these codes work fine currently!

PR #585, opened 2020-07-28
fix bugs and add news for todo

jiaweizang:


sbillinge:
this looks great, thanks @jiaweizang !

PR #588, opened 2020-07-29
Mongodb Test on Windows

zthatch:
Modified the tests such that windows will now run mongodb tests if mongodb is installed, and all computers will now be able to selectively run filesystem tests and skip mongo tests. There will be a printout saying that the mongo tests were skipped, a link to where to install mongo, and a xonsh shell error saying that the mongo commands do not exist if the user has not installed mongodb community edition.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,12 +10,12 @@ from pymongo import MongoClient
 from pymongo import errors as mongo_errors
 from xonsh.lib import subprocess
 from xonsh.lib.os import rmtree
-from xonsh.proc import PopenThread
 
 from regolith.fsclient import dump_yaml
 from regolith.schemas import EXEMPLARS
 
 OUTPUT_FAKE_DB = False  # always turn it to false after you used it
+REGOLITH_MONGODB_NAME = "test"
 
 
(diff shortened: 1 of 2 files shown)


zthatch:
@st3107 Please take a look at my file changes to see if you agree with the implementation.

sbillinge:
Thanks @zthatch this looks great!  Is it ready to merge?

zthatch:
> Thanks @zthatch this looks great! Is it ready to merge?

@sbillinge I think so. I imagine that @st3107 may want to approve of the implementation. I will send him a question tomorrow morning asking if he wants to review of if I should just suggest a merge.

zthatch:
I have updated this implementation so that it is actually creating a database in the temporary directory for windows, and then utilizing it for tests rather than using the local database in the default location (None of this was actually happening in previous commits). However, there are still missing keys when I inspect the database. I need to make sure that I am actually looking at the database in the temporary location rather than the database in the default location, since that default database is always running as a windows service. This may be an optional mongo command that needs to be added when connecting the pymongo client to the local host on windows that enforces the database location.

sbillinge:
Thanks Zach, this is great

PR #589, opened 2020-07-29
initial PR for expense updater helper

AhmedS343:


sbillinge:
Nice start, thanks @AhmedS343 

I have played more with the todo updater.  I wonder if it would be better if we use that as a template?  I am not sure how much it deviates from the milestone updater, but it might be good synchronize them so they have the same behavior.  @jiaweizang is working on the todo updater.

The first thing to work on is the tests that implement the use-cases before we code in any case.  I suggest that you have a call with Jiawei and work together to harvest the UCs that she and i have developed for the todo updater, and adapt them, then write the tests.  I can join that call too, since it has another purpose which is to discuss exactly the workflow that we want for just about all the updater helpers.

PR #590, opened 2020-07-30
add assigned_by and update sorting in todolist

jiaweizang:
Hi @sbillinge 
This PR mainly updates:
1. sorting of finished tasks
2. add -b,  --assigned_by:
    if enter -b, then helper lists tasks assigned by the id in user.json, if -b jwzang, then lists tasks assigned by jwzang.
3. add -s,  --stati:
    if no -s, then helper lists started tasks, if -s finished, then lists finished tasks, if -s finished started, then lists all tasks.
4. make the shortcuts consistent

I will soon make a separate PR for kv filter after this PR is merged, in case too many codes are changed.

### DIFF
--- a/regolith/helpers/a_todohelper.py
+++ b/regolith/helpers/a_todohelper.py
@@ -29,14 +29,17 @@ def subparser(subpi):
     subpi.add_argument("duration",
                        help="The estimated duration the task will take in minutes.",
                        )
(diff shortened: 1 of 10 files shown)


sbillinge:
thanks SO MUCH @jiaweizang !

PR #591, opened 2020-07-30
fix bugs in l_todo

jiaweizang:


regolith/helpers/l_todohelper.py
```diff
@@ -153,6 +153,8 @@ def sout(self):
         print("-" * 42)
         print("tasks from projecta and other collections:")
         print("-" * 42)
+        if rc.stati == ["started"]:
+            rc.stati = ["started", "converged", "proposed"]
```
sbillinge, line 157: it is better not to hard-code this here but to make a global variable at the top of the file that defines this.  It is easier to maintain because if we decide to add a new "active status" like ``fun`` or sthg, you just add it there, and everywhere where sthg is filtering for active stati it will be correctly updated.

PR #595, opened 2020-07-31
fixed --all, --ended, and --current

sheneric88:
@sbillinge 

--all, --current, and --ended (with -d, -r, and standalone) all work now...at least locally.

### DIFF
--- a/regolith/helpers/l_projectahelper.py
+++ b/regolith/helpers/l_projectahelper.py
@@ -146,7 +146,9 @@ class ProjectaListerHelper(SoutHelperBase):
                 continue
             if rc.ended:
                 if projectum.get('status') not in ACTIVE_STATI:
-                    if projectum.get('status') != 'finished' or projectum.get('end_date') == None or type(projectum.get('end_date')) != dt.date:
+                    if projectum.get('status') == 'cancelled':
+                        continue
+                    elif projectum.get('status') != 'finished' or projectum.get('end_date') == None or type(projectum.get('end_date')) != dt.date:
                         error_projecta.append(projectum)
                     else:
                         end_date = projectum.get('end_date')
--- a/regolith/helpers/l_projectahelper.py
+++ b/regolith/helpers/l_projectahelper.py
@@ -27,42 +27,33 @@ ACTIVE_STATI = ["proposed", "started"]
 
 def subparser(subpi):
     subpi.add_argument("--all", action="store_true",
-                       help="Lists all projecta in general"
-                       )
+                       help="Lists all projecta in general")
     subpi.add_argument("-c", "--current", action="store_true",
                        help="Lists all active projecta")
     subpi.add_argument("-v", "--verbose", action="store_true",
-                       help='increase verbosity of output'
-                       )
+                       help="increase verbosity of output")
     subpi.add_argument("-l", "--lead",
-                       help="Filter milestones for this project lead"
-                       )
+                       help="Filter milestones for this project lead")
     subpi.add_argument("-p", "--person",
-                       help="Filter milestones for this person whether lead or not"
-                       )
+                       help="Filter milestones for this person whether lead or not")
     subpi.add_argument("-e", "--ended", action="store_true",
-                       help="Lists projecta that have ended. Use the -d and -r flags to specify"
-                            " up to what date and how many days around that"
-                       )
+                       help="Lists projecta that have ended. Use the -d and -r flags to specify up to "
+                            "what date and how many days before then. The default is 7 days before today.")
     subpi.add_argument("-d", "--date",
-                       help="projecta with end_date within RANGE before this date will be listed."
-                            " Default is today"
-                            " Be wary some projecta do not have a specified end date"
-                       )
+                       help="projecta with end_date within RANGE before this date will be listed. "
+                            "The default is today. Some projecta don't have an end date and won't appear in a search")
     subpi.add_argument("-r", "--range",
-                       help="date range back from DATE to search over in days. If no "
-                            "range is specified, search will be 0 days"
-                       )
+                       help="date range back from DATE to search over in days. "
+                       "If no range is specified, search range will be 7 days")
     subpi.add_argument("-g", "--grant",
-                       help="Filter projecta by a grant ID"
-                       )
+                       help="Filter projecta by a grant ID")
     subpi.add_argument("--grp_by_lead", action='store_true',
                        help="Lists all projecta by their lead")
     subpi.add_argument("-f", "--filter", nargs="+",
-                       help="Search this collection by giving key element pairs"
-                       )
-    subpi.add_argument("-k", "--keys", nargs="+", help="Specify what keys to return values from when running "
-                                                       "--filter. If no argument is given the default is just the id.")
+                       help="Search this collection by giving key element pairs")
+    subpi.add_argument("-k", "--keys", nargs="+",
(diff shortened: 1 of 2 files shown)


regolith/helpers/l_projectahelper.py
```diff
@@ -22,29 +22,36 @@
...
-                       help="Lists projects that have ended. Use the -d and -r flags to specify"
-                            "from one date and how many days"
+                       help="Lists projecta that have ended. Use the -d and -r flags to specify"
+                            " up to what date and how many days around that"
                        )
     subpi.add_argument("-d", "--date",
                        help="projecta with end_date within RANGE before this date will be listed."
-                            "Default is today"
+                            " Default is today"
+                            " Be wary some projecta do not have a specified end date"
```
sbillinge, line 50: what is a user supposed to do with this information?  Give the users useful information that helps guide their actions.  I would delete this, but perhaps have a conversation about what your intent was here first?"

regolith/helpers/l_projectahelper.py
```diff
@@ -22,29 +22,36 @@
...
                        )
     subpi.add_argument("-r", "--range",
                        help="date range back from DATE to search over in days. If no "
-                            "range is specified, search will be 7 days"
+                            "range is specified, search will be 0 days"
```
sbillinge, line 54: Is this correct?  I thought the desired default range was "ended in the last 7 days", so that when we run this with -e it will show all the prums that ended in the previous week, which is what we want to do each group meeting.

regolith/helpers/l_projectahelper.py
```diff
@@ -112,17 +119,22 @@ def sout(self):
...
-            raise RuntimeError(f"please specify either lead or person, not both")
+            raise RuntimeError(
+                f"please specify either lead or person, not both")
         for projectum in collection:
-            if rc.all and projectum.get('status') != "finished":
+            if rc.all:
                 projecta.append(projectum)
                 continue
+            if rc.current:
+                if projectum.get('status') in ACTIVE_STATI:
```
sbillinge, line 135: nice!!!!

regolith/helpers/l_projectahelper.py
```diff
@@ -140,32 +152,31 @@ def sout(self):
...
+                                end_date = date_parser.parse(
+                                    end_date).date()
+                            low_range = desired_date - \
+                                dt.timedelta(days=num_of_days)
+                            high_range = desired_date + \
+                                dt.timedelta(days=num_of_days)
+                            if num_of_days != 0:
+                                if low_range <= end_date <= high_range:
+                                    end_projecta.append(projectum)
+                            else:
```
sbillinge, line 172: is this needed?  Ithink it should work with just the "<=" above, no?

regolith/helpers/l_projectahelper.py
sheneric88: Ah ok, if this is for just inside the group then yep I'll edit that.

regolith/helpers/l_projectahelper.py
sheneric88: wait sorry, what do you mean by this?

Actually, I was guessing the behaviour, so I was thinking if it's set to 0 days, then it'd show all the prums finished before the set date.

sheneric88: should the behaviour be that it shows finished prums for the n days before?

regolith/helpers/l_projectahelper.py
sheneric88: The criteria for it to show up in -e is that it has to 1. have some sort of finished/submitted/cancelled string typed into 'status' and 2. the date has to fall in the range specified.

So since some projecta don't have the end_date entry, they won't come up in the search.

regolith/helpers/l_projectahelper.py
sbillinge: thanks

regolith/helpers/l_projectahelper.py
sbillinge: yes, that's right.  prums finished in the n days preceding date.  By default date is today and n is 7 but these can be set to anything you want'

regolith/helpers/l_projectahelper.py
sbillinge: almost., it should be status = finished only.

If status = finished and there is no end-date, then print a warning.

sbillinge: in the warning, ask people to add an end date and remind people to use ``f_prum`` which automatically sets status to finished and adds an end date.

sheneric88: what do you think about prums with status = cancelled?

sheneric88: I'll set it up to print out an error_projecta list when you run -e. thought I feel like not many people would be using -e in general, so should this be something brought up to have people fix?

while getting the code to work the first time around, I did end up with a list of improperly formatted prums, so I already know who should fix status and end_date at least.

sheneric88:
@sbillinge 

Overhauled descriptions and tuned the -e output/input.

regolith/helpers/l_projectahelper.py
sbillinge: cancelled means they weren't finished but are not active any more, but there may be info in there that we don't want to lose per se, so it should disappear off all lists basically, unless we want to recover them later which we could do using the kv filter.

Yes, you are right on the error code, but we will see the error list if it is run at group meeting....so people will know to fix it.  Over time this error rate will go down as people start using f_prum, which didn't exist before.

regolith/helpers/l_projectahelper.py
```diff
@@ -140,32 +144,24 @@ def sout(self):
...
+                    if projectum.get('status') != 'finished' or projectum.get('end_date') == None or type(projectum.get('end_date')) != dt.date:
+                        error_projecta.append(projectum)
+                    else:
+                        end_date = projectum.get('end_date')
+                        if isinstance(end_date, str):
+                            end_date = date_parser.parse(
+                                end_date).date()
+                        low_range = desired_date - \
+                            dt.timedelta(days=num_of_days)
+                        if low_range <= end_date <= desired_date:
```
sheneric88, line 158: @sbillinge 
Just realized I was missing part of the range filtering here.
Can you merge/check this before group meeting tmrw?

sbillinge:
@sheneric88 I'm happy to merge.   Please respond to all the comments though, so I know you saw them and how you responded.

regolith/helpers/l_projectahelper.py
sheneric88: yep, I adjusted it so it does default today + 7days back

regolith/helpers/l_projectahelper.py
sheneric88: fixed it to 7 days and changed up some of the other wording too

regolith/helpers/l_projectahelper.py
sheneric88: ok, so current functionality won't show cancelled or anything else under finished prum searches

sheneric88:
@sbillinge 
cool, replied to everything, and added a line so cancelled prums won't show up in the errors part, since they're technically correct

sbillinge:
Thanks @sheneric88 .   This is still failing one test, but then I can merge it.

sheneric88:
@sbillinge 

Found the test it was failing and fixed it!
A line spacing error

PR #596, opened 2020-08-03
HTML Builder - changed function finish, adding a if statement

dy2403:
Hi Prof. @sbillinge!
Here is the pull request with the changes we talked about today.


sbillinge:
Thanks @dy2403 .  Not sure where the is build error came from but please could you fix it?  Maybe something wrong in a schema or ouput file?  2008_einstein....

dy2403:
> Thanks @dy2403 . Not sure where the is build error came from but please could you fix it? Maybe something wrong in a schema or ouput file? 2008_einstein....

I am not sure why it was different, but I just changed the year and now it passed.

PR #597, opened 2020-08-07
Appointments now does some small amount of projections

sbillinge:


sbillinge:
@adibaejaz this is the latest version of the appointments maker.  I am actually using it to do my appointments and it working well.

Please could you do me a favor and take this and make it pass tests?  Also, I need a test written for a function I put in tools.py (sorry, I should have used TDD! :( )

One more thing I really need is for the budget increments to happen at the beginning of each budget period, and not all on day 1 of the grant.  I am not sure how much work that is.  Let's discuss that.  For now, if you could clone this and then push up a PR that passes tests I can merge it, that would be great.

Apart from the incrementing by budget period, there are some additions I would want to make this truly useful, but even in its current form it is night and day for me, so thank you so much for your efforts on this!   Oh, when you have it working, could you make it flake8 compliant?  This should be done on a separate commit because it will reformat the file a lot so you really don't want to mix it with any code edits!  Get everything passing tests, then run flake8 or black just on the files you have changed (you can also reformat the file automatically inside pycharm.  Reach out if you want to know how.  but as I said, don't mix that work with code edits.  Get everything passing tests, then clean)

sbillinge:
@adibaejaz yeh it worked!  Thanks so much!   I have merged this.....

PR #615, opened 2020-09-10
patch to milestonehelper so it updates stati correctly

sbillinge:


sbillinge:
close and reopen to trigger travis

Issue #662, opened 2020-12-18
missing year when create new prum


dragonyanglong:
When use the a_projectum helper function, it doesn't include the year in the prum name anymore.

See `a_projectumhelper.py` file:
`key = f"{rc.lead[:2]}_{''.join(rc.name.casefold().split()).strip()}"`

sbillinge:
this is desired behavior.  I realized that was needed when I was using the file-system for organizing things but not when using a db because I can write a lister that lists by date, for example.

sbillinge:
@dragonyanglong thanks for catching that though!

dragonyanglong:
opppps. sorry

PR #663, opened 2020-12-18
BUG: get the year when create new prum.

dragonyanglong:
@sbillinge , I found a bug when creating new prum. It seems like someone changed the prum name, not including the year anymore... 

I created this issue #662  and fixed it in this PR. Please review and merge.


sbillinge:
Thanks for the PR, but I will close it.  Pls see the comment on the issue.

PR #680, opened 2021-01-08
WIP: abstract lister helper

priyankaanehra:
helper.py updated with l_abstract and initial version of l_abstracthelper 
This work was based on using the template of l_contactshelper.py so far.


sbillinge:
@priyankaanehra nice job, I see this!  Keep pushing updates here and I can comment

priyankaanehra (commented):
Message should be -m "Added two tests for abstract lister, one with all info used to filter and one with some information to filter (where the more information given, the better it filters)"

tests/test_helpers.py
```diff
@@ -144,6 +144,41 @@
      "    aka:\n"
      "        -A. B. Friend\n        -AB Friend\n        -Tony Friend\n"
      ),
+    (["helper", "l_abstract", "run", "-a", "scopatz”, "-y", "2018", "-l", ”upton", "-t", "tape"],
+     "---------------------------------------\n"
+     "Authors: - scopatz - afriend\n"
+     "Meeting to check flexibility on dates\n"
+     "columbiau, apam\n"
+     "Location: Upton NY\n"
+     "Start: 5-22-2018\n"
```
sbillinge, line 153: let's keep dates in ISO format (so 2018-05-22)

tests/test_helpers.py
```diff
@@ -144,6 +144,41 @@
      "    aka:\n"
      "        -A. B. Friend\n        -AB Friend\n        -Tony Friend\n"
      ),
+    (["helper", "l_abstract", "run", "-a", "scopatz”, "-y", "2018", "-l", ”upton", "-t", "tape"],
```
sbillinge, line 147: in the test function we don't use the abbreviation for the command line switches, to make it easier to read, so --author etc. here

tests/test_helpers.py
```diff
@@ -144,6 +144,41 @@
...
+     "---------------------------------------\n"
+     "Authors: - scopatz - afriend\n"
+     "Meeting to check flexibility on dates\n"
+     "columbiau, apam\n"
+     "Location: Upton NY\n"
+     "Start: 5-22-2018\n"
+     "Type: award, Status: accepted\n"
+     "Project: 18sob_clustermining\n"
```
sbillinge, line 155: don't need project listed so please delete this line

tests/test_helpers.py
```diff
@@ -144,6 +144,41 @@
...
      "        -A. B. Friend\n        -AB Friend\n        -Tony Friend\n"
      ),
+    (["helper", "l_abstract", "run", "-a", "scopatz”, "-y", "2018", "-l", ”upton", "-t", "tape"],
+     "---------------------------------------\n"
+     "Authors: - scopatz - afriend\n"
+     "Meeting to check flexibility on dates\n"
+     "columbiau, apam\n"
+     "Location: Upton NY\n"
+     "Start: 5-22-2018\n"
+     "Type: award, Status: accepted\n"
```
sbillinge, line 154: don't need these either for this purpose

tests/test_helpers.py
```diff
@@ -144,6 +144,41 @@
...
+     "Project: 18sob_clustermining\n"
+     "Title: We pulled apart graphite with tape\n"
+     "Abstract: Graphitic Dephenestration"
```
sbillinge, line 157: Let's make it look like this in the output:
```
Title:   We pulled apart graphite with tape

Anthony B. Scopatz,  Anthony F. Friend

Abstract: Graphitic Dephenestration
```
to get the names you need to dereference from the people/contacts collections. There is a function in tools that does that `get_person_contact()`.  look in preslistbuilder to see how to use it.

tests/test_helpers.py
```diff
@@ -144,6 +144,41 @@
      "    aka:\n"
      "        -A. B. Friend\n        -AB Friend\n        -Tony Friend\n"
      ),
+    (["helper", "l_abstract", "run", "-a", "scopatz”, "-y", "2018", "-l", ”upton", "-t", "tape"],
+     "---------------------------------------\n"
+     "Authors: - scopatz - afriend\n"
+     "Meeting to check flexibility on dates\n"
```
sbillinge, line 150: to keep it consistent let's have
"Meeting name/department: Meeting to...."

tests/test_helpers.py
```diff
@@ -144,6 +144,41 @@
      "    aka:\n"
      "        -A. B. Friend\n        -AB Friend\n        -Tony Friend\n"
      ),
+    (["helper", "l_abstract", "run", "-a", "scopatz”, "-y", "2018", "-l", ”upton", "-t", "tape"],
+     "---------------------------------------\n"
+     "Authors: - scopatz - afriend\n"
```
sbillinge, line 149: This will be moved below (see below)

sbillinge (commented):
@priyankaanehra this is super-awesome!  Well done!   Please see my comments inline

tests/test_helpers.py
```diff
@@ -144,6 +144,22 @@
      "    aka:\n"
      "        -A. B. Friend\n        -AB Friend\n        -Tony Friend\n"
      ),
+    (["helper", "l_abstract", "run", "---author", "scopatz”,”--location”,”upton"],
```
sbillinge, line 147: spaces after commas (see PEP8 for syntactic standards, but if you are using PyCharm you can have Pycharm modify the whole file automatically)

sbillinge (commented):
tests look great!  now we are ready to code up the helper till the tests pass!  Do you want another call to get to work on that, or do you want to just dive in?

btw, I didn't look at the helper code at all yet (I don't switch focus to that until I think the test is right), but I can take a look and make a separate review if you think it is ready for me to look at.

regolith/helpers/l_abstracthelper.py
```diff
@@ -0,0 +1,105 @@
...
+    return subpi
+
+
+
+class AbstractListerHelper(SoutHelperBase):
+    """Helper for finding and listing abstracts from the presentations.yml file
+    """
+    # btype must be the same as helper target in helper.py
+    btype = HELPER_TARGET
+    needed_dbs = [f'{TARGET_COLL}', 'authors']
```
sbillinge, line 50: there is no `authors `collection.  You will need `people `and `contacts`

regolith/helpers/l_abstracthelper.py
```diff
@@ -0,0 +1,105 @@
...
+        """Constructs the global context"""
+        super().construct_global_ctx()
+        gtx = self.gtx
+        rc = self.rc
+        if "groups" in self.needed_dbs:
+            rc.pi_id = get_pi_id(rc)
+        rc.coll = f"{TARGET_COLL}"
+        try:
+            if not rc.database:
+                rc.database = rc.databases[0]["authors"]
```
sbillinge, line 62: not sure what this is but something is wrong here. need to remove ["authors"] I think

regolith/helpers/l_abstracthelper.py
```diff
@@ -0,0 +1,105 @@
...
+        gtx["all_docs_from_collection"] = all_docs_from_collection
+        gtx["float"] = float
+        gtx["str"] = str
+        gtx["zip"] = zip
+
+    def sout(self):
+        rc = self.rc
+        collection = self.gtx["presentations"]
+
+        if rc.year is None:
```
sbillinge, line 82: I think you are using rc.year to filter, so don't set it to empty string, leave as None.  Same for rc.location and rc.title

regolith/helpers/l_abstracthelper.py
```diff
@@ -0,0 +1,105 @@
...
+            rc.year = ''
+        if rc.location is None:
+            rc.location = ''
+        if rc.title is None:
+            rc.title = ''
+
+        for key in collection:
```
sbillinge, line 89: make it more readable.  Your coll is presentations, so `for presentation in presentations:` is much more intuitive....

regolith/helpers/l_abstracthelper.py
```diff
@@ -0,0 +1,105 @@
...
+            if rc.author in key[1]['authors']:
```
sbillinge, line 90: this is using AND logic which is probably not what you want.  e.g., you have the rc.title nested inside rc.author and so on. So if author is not specified this will print nothing.  Is that what you want?  Maybe we need more tests, e.g., a test for each of these:
```
> regolith helper l_abstract
> regolith helper l_abstract --author ascopatz
> regolith helper l_abstract --title tape
> regolith helper l_abstract --location upton
> regolith helper l_abstract --year 2018
> regolith helper l_abstract --location upton --year 2018
```
something like that.  What do you want to happen in each of these cases?

a trick you can use then for the OR logic is something like
```

**Review by sbillinge — COMMENTED** [review]
few comments on the code

File: tests/test_helpers.py (hunk 8 above)
Comment (priyankaanehra) [reply tests/test_helpers.py:147]: yes just looked into some standards, for this, do you mean in particular to have a space after each comma as there is not one in the last two for location?

**Comment by priyankaanehra** [comment]
Sorry I meant to warn you that I am working on integrating my functions with regolith tools already present so my code does not really work at all in regolith at all but slowly trying to use debugger to figure out how to do what I need. I wanted to keep a copy of this code however so I know the reasoning I did beforehand so I can look back when I am changing it all up.

Another meeting would be helpful for the helper. In the meantime, I am diving in. :)

File: regolith/helpers/l_abstracthelper.py (hunk 13 above)
Comment (priyankaanehra) [reply regolith/helpers/l_abstracthelper.py:90]: I was trying to bypass this problem via a maybe iffy way. If a certain command line argument such as "--title" and the keywords weren't entered, it would still enter the nested if loop because I wrote lines of code before the if loop, where if an argument wasn't entered, it would default to an empty string. This would let it still apply the author, location and year filters. I realize my code does fail when an entry in presentations doesn't contain the key "location" or "year" or "title" so that is something I also need to work on.

Comment (sbillinge) [reply regolith/helpers/l_abstracthelper.py:90]: yes, sounds iffy.

A pythonic way might be
```
pres_dates = get_dates(pres)
good_year = [pres for pres in presentations if pres_dates.get("year") == rc.year]
good_author = [pres for pres in presentations if rc.author in pres.get("authors")]
good presentation = set(good_year+good_author)
```
of sthg like that.

**Review by sbillinge — COMMENTED** [review]
pls see my comment

File: regolith/helpers/l_abstracthelper.py [hunk 14]
```diff
@@ -77,29 +72,17 @@ def construct_global_ctx(self):
...
+            if rc.title.casefold() not in presentation.get('title').casefold():
+                continue
+            print("---------------------------------------")
+            print(f"Title: {presentation.get('title')}\n")
+            author_list=[author
+                         if not get_person_contact(author, self.gtx["people"], self.gtx["contacts"])
+                         else get_person_contact(author, self.gtx["people"], self.gtx["contacts"]).get('name')
+                         for author in presentation.get('authors')]
+            print(", ".join(author_list))
+            print(f"\nAbstract: {presentation.get('abstract')}")
```
sbillinge, line 87: always finish code files with a line-feed (i.e., put cursor there and hit enter)

sbillinge (commented):
exciting!  please see one small comment inline

priyankaanehra:
All 7 tests pass when I reassign helper_map =[my 7 tests] at the bottom of the test_helpers.py as we went through together, however when I integrate it with all the tests, it doesn't. I think this is because there's a presentation being added that isn't in schemas for presentations. Not exactly sure yet. I also have it printing out all presentations if no command line arguments given but I will change that to the help panel soon.

regolith/helpers/l_abstracthelper.py
```diff
@@ -0,0 +1,104 @@
...
+        help="Increases the verbosity of the output.")
+    subpi.add_argument(
+        "-a",
+        "--author",
+        help='authors group ID(single argument only) to use to find presentation abstract.')
+    subpi.add_argument(
+        "-y",
+        "--year",
+        help='Start or end year of the presentation (single argument only) to use to find presentation.')
+    subpi.add_argument(
```
sbillinge, line 30: I think I might search based on institution if it is a seminar or location if it is a meeting.  The best would be if location could take either.  if pres['type'] is seminar or colloquium it should look in pres['institution'] and if it is anything else, look in location.

sbillinge (commented):
looks good.  by default maybe we should only look for future presentations (since this is the main UC) and have a --all that overrides this (change the help message to reflect what actually is happening)

sbillinge (commented):
pls see comments

regolith/helpers/l_abstracthelper.py
priyankaanehra: To double check, would it be preferential to have two command argument choice, one for '--location' and one for '--institution' or would it be easier to just use '--location' and from it, u can choose to enter an institution, city, state, etc. And either way, It will go into searching based on the type if its a seminar or colloquium or if not, look in location. I wasn't sure if it would confusing for the user to use '--location' for all types of locations such as institution since it's technically a location. In addition, I was thinking it would easier to have one that encompasses both types of locations so you don't have to remember to many arguments. I have figured out how to do it both ways, but let me know which may be more user-friendly and I will adjust it and push a test case for this condition. :)

sbillinge: yes, just one.  Maybe we should call it `--loc_inst` and the help message should be very clear about what is wanted.

regolith/helpers/l_abstracthelper.py
```diff
@@ -22,19 +23,26 @@ def subparser(subpi):
...
     subpi.add_argument(
         "-l",
-        "--location",
-        help='Location of presentation, either a country, city, state, or university.')
+        "--loc_inst",
+        help='Location of presentation, either a fragement of an institution, '
+             'country, city, state, or university. If an instiution is entered,'
+             'the search will be for seminars or colloquiums, otherwise the '
+             'search will be for meetings')
+
```
priyankaanehra, line 40: disregard the empty line at 40, it is actually removed (slipped my gaze initially) but I will push that in the next big change

regolith/helpers/l_abstracthelper.py
```diff
@@ -74,25 +82,37 @@ def sout(self):
...
         for presentation in presentations:
             if rc.title is not None:
                 if rc.title.casefold() not in presentation.get('title').casefold():
                     continue
-            if rc.location is not None and 'location' in presentation:
-                if rc.location.casefold() not in presentation.get('location').casefold():
-                    continue
-            if rc.location is not None and 'location' not in presentation:
-                continue
+            if presentation['type'] == 'seminar' or presentation['type'] == 'colloquium':
```
sbillinge, line 92: a better way to do this (less maintenance in general) is to make a global variable `SEMINAR_TYPES= ['seminar', colloquium]` at the top of the file and here you can type `if presentation.get('type') in SEMINAR_TYPES:` .  This way, if we decide later that "zoom-talk" is an allowed seminar type we just add it to that list and all the code just works.

Also, using `dict_object.get('item,on_fail)` method is a safer way to get a dict item.  If it can't find the dict item it will return whatever you put as the second argument, `on_fail` and the program won't crash.

regolith/helpers/l_abstracthelper.py
```diff
@@ -74,25 +82,37 @@ def sout(self):
...
+                if rc.loc_inst is not None:
```
sbillinge, line 93: make it more readable, don't use double negative!   `if rc.loc_inst:`

regolith/helpers/l_abstracthelper.py
```diff
@@ -74,25 +82,37 @@ def sout(self):
...
+                    if rc.loc_inst.casefold() not in presentation.get('institution').casefold():
+                        continue
+            else:
+                if rc.loc_inst is not None and 'location' in presentation:
```
sbillinge, line 97: ugly code here too.  If I were you, I would take a step back and try and think what is a more elegant way to do this whole activity of filtering for the location and start over.  I am not sure the logic, but when I see tortuous code like this I am almost sure there is a better way to do it.

sbillinge (commented):
some comments inline, please take a look.

sbillinge:
@priyankaanehra I thought some more and what you are essentially trying to do with the rc.loc_inst is a filter...from a set of object return the objects where some criterion is satisfied. Python has a filter function but a more pythonic way of doing it is considered to be a list comprehension.  Try
```
if rc.loc_inst:
    loc_satisfied_presentations = [
                  [talk for talk in presentations if rc.loc_inst in talk.get(key)] 
                  for key in SEMINAR_TYPES
                 ]
```
This may return a list of lists that you will have to flatten

regolith/helpers/l_abstracthelper.py
priyankaanehra: Thank you, I had no idea about dict_object.get('item,on_fail), this helped me cut down on some repetitive code. Sorry, java is what I know more formally so i'm slowly learning to be more formal programming in python as well since I am self-learning. ~work in progress

tests/test_helpers.py
sbillinge: yes.  If you are using PyCharm you can put the cursor right after that "bad" comma and you will see a yellow zigzag underscore.  A yellow light will pop up and you can click on it and it will say "reformat file" then you can do that.  It will make lots of changes in the file (formatting) so it is best to commit all your changes, then do this, then do another commit, or it becomes hard to find your actual edits alongside all the reformating edits

regolith/helpers/l_abstracthelper.py
```diff
@@ -1,15 +1,17 @@
-
 from regolith.helpers.basehelper import SoutHelperBase
 from regolith.fsclient import _id_key
 from regolith.tools import (
     all_docs_from_collection,
     get_pi_id,
     get_person_contact
 )
+import itertools
```
priyankaanehra, line 8: opps need to delete this, didn't use in the end

regolith/helpers/l_abstracthelper.py
```diff
@@ -1,15 +1,17 @@
...
+
 
 TARGET_COLL = "presentations"
 HELPER_TARGET = "l_abstract"
 
+
```
priyankaanehra, line 14: also extra line accidentally here not necessary anymore

sbillinge (commented):
thanks, it looks good.  Nearly ready to merge.  It seems from your comments you had a couple of things to fix, and I made a request in another comment.

regolith/helpers/l_abstracthelper.py
```diff
@@ -81,38 +82,37 @@ def construct_global_ctx(self):
...
+                                      or int(rc.year) == presentation.get('end_year', 'end_date')]
+        if rc.loc_inst:
+            filtered_inst = [presentation for presentation in presentations
+                             if presentation.get('type') in SEMINAR_TYPES and
+                             rc.loc_inst.casefold() in presentation.get('institution').casefold()]
+            filtered_loc = [presentation for presentation in presentations
+                            if rc.loc_inst.casefold() in presentation.get('location','institution').casefold()
+                            and rc.loc_inst.casefold() not in presentation.get('institution').casefold()]
+
+        list_of_sets = [filtered_inst, filtered_years, filtered_title,filtered_authors,filtered_loc]
```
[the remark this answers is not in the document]
priyankaanehra: Perfect, I have updated to easier variable names and other small changes. I had a small question about when there is no user input, what would be the best way (or if there is a command I am unaware of) to go to the help screen of commands? Does simply having the function "return" work for when there is no commands or is there more that needs to be done to do this functionality?

sbillinge:
@priyankaanehra this is a super job, fantastic.    It is basically ready to merge.

Regarding your question about the returns, the lister that has the best behavior in this regard is l_projecta.   We should maybe look at how that is structured and make this one the same.  The behavior there is that if no args are given it returns the help, but it is not done using a bs workaround of asking the user to type a required arg that is actually not required.    So I suggest to look at the pattern in there and we can maybe reuse that?

When you have done that I will merge, but you have done an amazing job here.....congrats!

priyankaanehra:
> @priyankaanehra this is a super job, fantastic. It is basically ready to merge.
> 
> Regarding your question about the returns, the lister that has the best behavior in this regard is l_projecta. We should maybe look at how that is structured and make this one the same. The behavior there is that if no args are given it returns the help, but it is not done using a bs workaround of asking the user to type a required arg that is actually not required. So I suggest to look at the pattern in there and we can maybe reuse that?
> 
> When you have done that I will merge, but you have done an amazing job here.....congrats!

Ahh okay. In that case, I think it's already got that function as if you type in "regolith helper l_abstract", it goes to the help messages screen. When I first coded in the case that no arguments were entered, I had followed the l_projecta template so that part is actually ready afterall. Yay! :)

On another note, I have been thinking about another case I wanted to bring up, specifically for the --loc_inst argument. There are several presentations that are located at columbia for 'institution'  but also for 'location'. The program currently would look through any seminars or colloquiums and if it finds matches, only output those types. However, this is problematic if you wanted the columbia matches in specifically 'location'. In this case, I don't know whether it is better to separate the user arguments for --location and --institution? I can also make it so that it searches through both 'location' and 'institution' for all entered places by the user so you get both that contain "columbia" for this instance. 

Let me know if this case isn't clear/ if it would be easier to set up a meeting to elaborate.

sbillinge:
fantastic job @priyankaanehra !   I am very impressed.   

The question about "both" a location and institution, this is just bad data in the database (the cause may be a bad schema of mine, but that is a story for another day).  So we should just fix up the database.  Actually, I guess we could put a couple of lines of code and throw an exception if the builder finds both in an entry.  Even  better would be something in the database validation that checks for this.  Database validation is done using Cerberus.  I think we should merge this, and make an issue on regolith (could you do that?) saying to fix this?  More or less a cut and paste of this convo into the issue.   

Then we find the next job for you.

sbillinge:
Hi @priyankaanehra 

OK so I tried it out and I have some comments.

1. it is not using the l_projecta pattern because it asks for the "run" argument.  We do want it to default to the help message if run without an y arguments, but it is more elegant if we don't use the "run" workaround and l_projecta doesn't
2. there appears to be no --date or --database keys.  These should (go at the bottom) but not be deleted.  The date is used for testing and database use used to override a default setting.  Please see the projecta for an example.
3. the help message nowhere tells the user what this helper does.
4. need a space in the help message for author
5. the behavior described in loc_inst help message I think is backwards.  I think the behavior should be that if the event is seminar/colloquium the abstract looks for this fragment in "institution" and otherwise it looks for it in "location".  
6. It's not exactly clear to me, but are the abstracts returned in chronological order (so latest at the bottom)?  This would be the best if possible.
7. The filter and keys logic seems to be gone too.  This should be preserved between listers.
8. I ran this `rh l_abstract run -y 2020 -l caltech` and it didn't find anything, but it should have....not sure what went wrong.

OK, I think that is it for now.  Great work....

priyankaanehra:
> Hi @priyankaanehra
> 
> OK so I tried it out and I have some comments.
> 
> 1. it is not using the l_projecta pattern because it asks for the "run" argument.  We do want it to default to the help message if run without an y arguments, but it is more elegant if we don't use the "run" workaround and l_projecta doesn't
> 2. there appears to be no --date or --database keys.  These should (go at the bottom) but not be deleted.  The date is used for testing and database use used to override a default setting.  Please see the projecta for an example.
> 3. the help message nowhere tells the user what this helper does.
> 4. need a space in the help message for author
> 5. the behavior described in loc_inst help message I think is backwards.  I think the behavior should be that if the event is seminar/colloquium the abstract looks for this fragment in "institution" and otherwise it looks for it in "location".
> 6. It's not exactly clear to me, but are the abstracts returned in chronological order (so latest at the bottom)?  This would be the best if possible.
> 7. The filter and keys logic seems to be gone too.  This should be preserved between listers.
> 8. I ran this `rh l_abstract run -y 2020 -l caltech` and it didn't find anything, but it should have....not sure what went wrong.
> 
> OK, I think that is it for now. Great work....

Comments 1,4,5,7,8 resolved on my local computer, to eventually push any changes, is this something I can do after creating the issue on regolith?  For #8, it wasn't working because there actually was an issue with the date as there was dates in many formats, specifically iso format caused the issues which would never evaluate as true for my boolean. This is updated. Once changes are pushed, this should be working!

For #6, they are printed out in the order in presentations.yaml file (which seems to be in order). I tried playing around to see if there was any easy fix in case they aren't in order in the yaml file. It's a little difficult since the months are sometimes strings instead of integers so it'll need some playing around to fit all the date types present to be in order.

PR #684, opened 2021-01-15
build bghtml

dragonyanglong:
Hi @sbillinge , I succeeded in passing the Google drive images to the html, and it can appear properly now.

I did some testing, lionmail google drive doesn't work, so we need to use a general gmail's google drive to save figures.

What I don't quite know is how to get values from yaml to the `bghtmlbuilder.py` level. So currently in this PR, I just write the texts and image id in the function to show you it works as you expected.

You can try running `regolith build bghtml` and click at the `html` file and see a figure (the PDFitc home page scrernshot) showing up.

### DIFF
--- a/regolith/builder.py
+++ b/regolith/builder.py
@@ -2,6 +2,7 @@
 
 from regolith.builders.cvbuilder import CVBuilder
 from regolith.builders.htmlbuilder import HtmlBuilder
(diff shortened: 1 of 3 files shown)


sbillinge:
thanks @dragonyanglong  that's cool.   It sucks that we can't do it from lionmail google drive though.  It is kind of a non-starter to have to set up a private Google for this.

PR #728, opened 2021-05-07
Make the --filter flag search recursively and added testing

### LINKED ISSUE #703: make filter function recurse
I am starting to use the `--filter target_key value` CL arg more and more.   rn it only searches for key at the top level though.  It would be good if it could recurse down to look for that key in a non-flat schema.  

There are a few possibilities...

1. if a key has a value that is type dict then look for `target_key` in the keys of that dict
2. if a key has a value that is a list, if the type of the list items is a dict then search through all the dicts in the list for `target_key`

not sure if this is going to result in infinite loops or sthg, but I think it should be straightforward.

mwinitch:
Closes #703

### DIFF
--- a/news/filter_recursive.rst
+++ b/news/filter_recursive.rst
@@ -1,0 +1,12 @@
+**Added:**
+    * The --filter flag can now recursively search through a top level target attribute, which can be an object or a list, in order to more effectively search a record
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


mwinitch:
I was thinking a lot on how to best approach this and decided the following was the best course:
Imagine that we have a record in a document that looks like the following
`
{
    '_id': 'abc',
    'name': 'sbillinge',
    'projects': [
        {'name': 'regolith helper', 'description': 'new helper for regolith'},
        {'name': 'Paper', 'description': 'Review student papers'}
     ],
    'institution': {
         'name': 'Columbia University',
         'location': 'NYC'
     },
     'status': 'active'
}
`

PR #737, opened 2021-05-21
Chattyclient

zthatch:
removed the annoying messages being sent that remind you to put a backend key in your regolithrc so that regolith no longer has to assume. This was meant to be the equivalent of a deprecation message, but because client.open gets called on init and by the context manager (for every database in a list), it gets printed excessively. Maybe I should instead just put a flag on it to say its been printed?

### DIFF
--- a/docs/commands/fs-to-mongo.rst
+++ b/docs/commands/fs-to-mongo.rst
@@ -15,7 +15,7 @@ https://www.mongodb.com/try/download/community
 .. code-block:: JSON
 {"groupname": "Example Group Name",
  "databases": [
-    "name": "<databasename>"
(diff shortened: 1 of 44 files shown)


sbillinge (commented):
one last comment.

sbillinge:
Thanks @zthatch 

Before merge, let's take this opportunity to remove all commented lines, unless they serve a purpose, and in case they are there, any unused imports (I didn't check).

sbillinge:
@zthatch gentle ping.  Please can we get this merged?  Thx.

sbillinge:
Thanks @zthatch .  That was a bigger clean than I intended, but all the better for that! :+1:

PR #740, opened 2021-05-28
Gui V0.1.0

yevgenyr:
@sbillinge, the first version of the GUI is ready to be used and merged. the development has it's own separate repo so it should not interfere with anything. 

- To run and play around with it - run the `rgui.py` file that is inside the `gui` repo.
- to set a default path to the local database - edit the entry in `gui/defaults/dbs_path.yml`  (for later versions I will make it to be the db-root repo with the regolithrc.json file... but later)


sbillinge:
moved to package in a seperate repo

PR #757, opened 2021-06-21
Added urllib.parse to handle password and id for mongo

zthatch:


zthatch:
I am finding a few issues with this line of programming logic. I can keep these in this PR, or start another. For the time being, I will keep them in this PR.

PR #760, opened 2021-06-24
Cleaning up the GUI and adding more widgets to helpers

PeterCheaterPumpkinEater:
I added more widgets and edited some text to make the GUI look prettier. I didn't know what to do with the "run" arguments under l_contacts and makeappointments. One thing to note is that running regolith helper from cmd will no longer work when I add widgets to the helpers, so I'm thinking that maybe we should have two seperate folders. The first folder will house the original helper files that regolith helper uses, and the second one will have the modified helper files that are imported into the helper_gui_main.py file.

### DIFF
--- a/news/TEMPLATE.rst
+++ b/news/TEMPLATE.rst
@@ -1,4 +1,5 @@
-**Added:** None
+**Added:**
+ * GUI for the helpers
 
 **Changed:** None
 
--- a/regolith/fsclient.py
+++ b/regolith/fsclient.py
@@ -154,7 +154,7 @@ class FileSystemClient:
             base, ext = os.path.splitext(collfilename)
             self._collexts[base] = ext
             self._collfiletypes[base] = "yaml"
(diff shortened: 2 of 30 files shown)


sbillinge:
Thanks Peter, this is great!
On the cli, I think it is likely a bad idea to separate into multiple files a gooey vs a cli version.  It makes it hard to maintain.  Much better would be some kind of design that tries to recognize if it is being run from gui-main or main-main and then wraps things in if statements where necessary.  We could use **kwargs for example, and the dict that is passed in to the parser depends on whether it is a gooey run or a cli one, something like that?  We may end up copy-pasting code, but finding a design that avoids this as much as possible generally pays back in the future.

sbillinge:
I am going to merge this into a branch called `gooey` on the main repo.  Please make additional PRs about the gui into that branch.  We can then make merge that into master when we are ready, but this way people can test it out more easily

PR #769, opened 2021-06-29
WIP: Gooey

sbillinge:
Hopefully CI will run on this PR.....

Please DO NOT MERGE yet.


sbillinge:
@pliu2781 @zthatch I am afraid I had to work on the parser of a_presentations (and maybe a_expenses?) and so your gooey work is now conflicted with master.    I think you will have to do a rebase onto master, but please make sure to keep all my changes and work around them!

Thanks so much, and sorry about that.  We are getting closer to merging gooey I think (can we do it before group meeting?) which will avoid this in the future.

S

PR #771, opened 2021-06-30
Branch off of original gooey branch

PeterCheaterPumpkinEater:
I made the code better by using isinstance() and **kwargs. Helpers should be able to run from both the command line and via the gui now. I am making this PR via the github website, but I still don't know why the CI isn't triggering. The weird thing is that the CI triggers for my PRs on to rg-db-group.

### DIFF
--- a/regolith/helpers/a_expensehelper.py
+++ b/regolith/helpers/a_expensehelper.py
@@ -12,21 +12,29 @@ from regolith.tools import (
     all_docs_from_collection,
     get_pi_id,
 )
+from gooey import GooeyParser
 
 TARGET_COLL = "expenses"
 ALLOWED_TYPES = ["business", "travel"] # need to check all expense types.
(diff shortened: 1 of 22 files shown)


sbillinge:
Thanks Peter, this is great.

The PR was made into master but we are making these PRs into the gooey branch now.   Don't worry, I changed it.    But please make new branches off of the gooey branch not off of master.

I am ok merging into gooey without CI running because we can run CI before merging gooey into master, but it is not great as if there are errors they will be harder to find there, so I will go ahead and merge.

Since you chose to just put the gooey kwargs into a dict and unpack it in the add_argument, the code would have been more readable if you had called all those variables `amount_gooey_kwargs` and so on.   If you can think of a quick way to change that it is probably still worth it, but it could be a lot of work.

sbillinge:
btw, the CI on the group repo is CircleCI and this is Travis, a different CI service, so it will be a different solution in each place.

PR #784, opened 2021-07-06
Updated tutorial docs

zthatch:
Documentation for mongo maintenance added as restructured text. Broker rst also moved to tutorial section


docs/tutorials/index.rst
```diff
@@ -0,0 +1,12 @@
...
+=================
+Tutorials
+=================
+The following contain tutorials for how to utilize regolith commands that require editing the regolith.rc file
+
+.. toctree::
+    :maxdepth: 1
+
+    broker
+    mongo_maintenance
```
sbillinge, line 12: I think this is the wrong place for this.  We want to separate "user-docs", "maintainer-docs" (a new category) and developer docs.    Developer docs don't appear here at all (docstrings etc. in the code).    I think maintainer docs could go here (why not, it is quite archival) but they should be well separated so we don't freak out users who wouldn't have to know about this stuff anyway.   Maybe we want to put "Maintaining production Regolith installations" or something somewhere way at the bottom and then put this tutorial down there?

docs/tutorials/mongo_maintenance.rst
```diff
@@ -0,0 +1,82 @@
+Mongo Maintenance
+=================
+
```
sbillinge, line 3: Start out explaining who this is for and when they may need to use it.  i.e., 

1.  if you are not using a mongo back-end, you can skip.   
2. Then give a roadmap from a high level.  What are the objectives that the person is trying to accomplish, so something like 
    1. set up a mongo instance, 
        1. use mongo atlas 
        1. run your own mongo locally
    2. move existing yml or json fs databases over
    3.  link regolith to the mongo db

and so on.  i.e, a brief roadmap of what the high-level objectives are, and then dive in to the step by step things to get you there.

sbillinge (commented):
This is really great!  I didn't proof read in detail.  Maybe that is actually easier after it is merged, or I can clone your branch and do it).  I left a few general comments though.

docs/tutorials/mongo_maintenance.rst
zthatch: Addressed

docs/tutorials/index.rst
zthatch: Addressed

docs/tutorials/index.rst
```diff
@@ -3,10 +3,9 @@
...
-The following contain tutorials for how to utilize regolith commands that require editing the regolith.rc file
+The following contain tutorials for how to utilize regolith commands that require a series of steps to perform
 
 .. toctree::
     :maxdepth: 1
 
-    broker
-    mongo_maintenance
\ No newline at end of file
+    broker
```
sbillinge, line 11: actually, do we want to move the broker instructions to the maintainance tutorials place?  Is the target audience users or maintainers (or developers)?   We could make the toc title less specific, so "Maintainer Docs"  "Developer Docs", something like that?

sbillinge (commented):
looks good.  another thought....

docs/tutorials/index.rst
zthatch: I would say that, even though the broker is a fairly involved tool, the intended audience are still typical users. Maintainers aren't going to have access to everyone's files such that they can create tokens etc. This is all up to the users. 

I'm thinking that database maintenance will ideally envelop how you setup mongo and how you set up your github fs infrastructure. Some broker info can go in the maintenance section, but just the info regarding how you set up the github storage location and distribute access. And I haven't written up anything about that.

sbillinge:
OK, I'm merging.  I will try and read the docs and make edits on another PR if it needs clarifying.

PR #785, opened 2021-07-07
Adding tests for test_filter_presentations

pliu2781:


tests/test_tools.py
```diff
@@ -2009,3 +2010,155 @@ def test_remove_duplicate_docs(inp, expected):
...
     actual = filter_employment_for_advisees(inp[0], inp[1], "ms", dt.date(2021,6,3))
     assert actual == expected
+
+person1 = {"_id":"tstark", "aka":"iron man", "name":"tony stark"}
+person2 = {"_id":"nromanov", "aka":"black widow", "name":"natasha romanov"}
+PEOPLE = [person1, person2]
+
+presentation1 = {"_id":"abc", "authors":["tstark"], "begin_year":2018, "begin_month":1, "begin_day":1,
+                 "department":"apam", "institution":"columbiau", "status":"accepted", "type":"award"}
+presentation2 = {"_id":"ghi", "authors":["tstark","nromanov"], "begin_year":2019, "begin_month":1, "begin_day":2,
```
sbillinge, line 2020: please can you enter as dates here `begin_date`, `end_date` to test that pattern

tests/test_tools.py
```diff
@@ -2009,3 +2010,155 @@ def test_remove_duplicate_docs(inp, expected):
...
+                 "end_year":2019, "end_month":1, "end_day":8, "department":"physics", "institution":"rutgersu",
+                 "status":"cancelled", "type":"poster"}
+PRESENTATIONS = [presentation1, presentation2]
+
+institution1 = {"_id":"columbiau", "city":"New York", "country":"USA", "name":"Columbia University", "state":"NY"}
+institution2 = {"_id":"rutgersu", "city":"New Brunswick", "country":"USA", "name":"Rutgers University", "state":"NJ"}
+INSTITUTIONS = [institution1, institution2]
+
+@pytest.mark.parametrize(
```
sbillinge, line 2029: please put this line right above the test with `@pytest.mark.parametrize("args, kwargs, expected", pres_map)` (see above other tests for the pattern)

tests/test_tools.py
```diff
@@ -2009,3 +2010,155 @@ def test_remove_duplicate_docs(inp, expected):
...
+    "status": "accepted",
+    "type": "award"}]
+    )
+  ]
+)
+
+
+def test_filter_presentations(args, kwargs, expected):
+    actual = filter_presentations(*args, **kwargs)
+    assert actual == expected
```
sbillinge, line 2164: need a line-feed here

tests/test_tools.py
```diff
@@ -2009,3 +2010,155 @@ def test_remove_duplicate_docs(inp, expected):
...
+    "department": {'name': 'apam'},
+    "institution": {'city': 'New York',
+                   'country': 'USA',
+                   'name': 'Columbia University',
+                   'state': 'NY'},
+    "status": "accepted",
+    "type": "award"}]
+    ),
+
+    #this tests 'statuses' and 'types' kwargs together
```
sbillinge, line 2091: it seems as if you are thinking of tests that test your testing structure (args, kwargs etc.) but we want to focus on testing the function.  Roughly speaking, look for all the `if` statements in the function and make sure you have a test that enters each of those.  Try and do it in as few tests as possible.

Just fyi, this is backwards from how we want to do things normally, so I am teaching you bad habits, but given the task I have given you, this is the strategy.  The RIGHT way to do it is to think what you want your function to do, write tests that test for that behavior, THEN write the function.....

sbillinge (commented):
Great job.  Few comments.  I checked codecov and actually there are only a few lines missing coverage, but maybe still worth seeing if they can be.

tests/test_tools.py
```diff
@@ -2015,17 +2015,15 @@ def test_filter_employment_for_advisees(inp, expected):
 person2 = {"_id":"nromanov", "aka":"black widow", "name":"natasha romanov"}
 PEOPLE = [person1, person2]
 
-presentation1 = {"_id":"abc", "authors":["tstark"], "begin_year":2018, "begin_month":1, "begin_day":1,
-                 "department":"apam", "institution":"columbiau", "status":"accepted", "type":"award"}
-presentation2 = {"_id":"ghi", "authors":["tstark","nromanov"], "begin_year":2019, "begin_month":1, "begin_day":2,
-                 "end_year":2019, "end_month":1, "end_day":8, "department":"physics", "institution":"rutgersu",
-                 "status":"cancelled", "type":"poster"}
+presentation1 = {"_id":"abc", "authors":"tstark", "date":"2018-01-01", "department":"apam",
```
sbillinge, line 2018: you changed them both, so now you are not testing the day, month year pattern...... :(

tests/test_tools.py
```diff
@@ -2015,17 +2015,15 @@ def test_filter_employment_for_advisees(inp, expected):
...
+                 "institution":"columbiau", "status":"accepted", "type":"award"}
+presentation2 = {"_id":"ghi", "authors":["tstark","nromanov"], "begin_date":"2019-01-02", "end_date":"2019-01-08",
+                 "department":"physics", "institution":"rutgersu", "status":"cancelled", "type":"poster"}
 PRESENTATIONS = [presentation1, presentation2]
 
 institution1 = {"_id":"columbiau", "city":"New York", "country":"USA", "name":"Columbia University", "state":"NY"}
 institution2 = {"_id":"rutgersu", "city":"New Brunswick", "country":"USA", "name":"Rutgers University", "state":"NJ"}
 INSTITUTIONS = [institution1, institution2]
-
```
sbillinge: this is not what I am asking for.  Did you check how it is done for the other tests, e.g., in test_helpers?  If it is not clear, please reach out.

tests/test_tools.py
```diff
@@ -2093,17 +2091,17 @@ def test_filter_employment_for_advisees(inp, expected):
     {"statuses" : ["all"], "types" : ["poster"]},
     [{"_id": "ghi",
     "authors": "tony stark, natasha romanov",
+    "begin_date": "2019-01-02",
```
sbillinge, line 2094: this is good.  btw, I think that good behavior for the function would be that it passes back date even if it was passed day, month, year.     Please use `get_dates` if it is not already being used.

sbillinge (commented):
few more comments.

sbillinge:
thanks @pliu2781  much better.  Still not quite the pattern I was looking for, but it is much cleaner and easier to follow, so I am merging

PR #787, opened 2021-07-08
Fixed mongo to fs

zthatch:
I also inserted code such that the existing subprocess calls print what went wrong before providing a traceback, which should be outstandingly useful for any future issues

### DIFF
--- a/news/mongo_to_fs.rst
+++ b/news/mongo_to_fs.rst
@@ -1,0 +1,13 @@
+**Added:**
+ * The "regolith mongo-to-fs" command can now be used to port local and remote mongo backends to the filesystem
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+ * Subprocess calls to mongo now printout the error message that would have gone to the stdout
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


sbillinge (approved):
Thanks, looks good.  Is it straightforward to write some tests for this?  Would be good if it is possible.  Also a News is needed.

zthatch:
> Thanks, looks good. Is it straightforward to write some tests for this? Would be good if it is possible. Also a News is needed.

@sbillinge Do you mean test the error printouts, or the mongo-to-fs?

The error printouts are really only invoked when something goes wrong to the point that the program fails. We can test these, but I don't see much benefit.

We currently test mongo-to-fs, just with local mongo servers, which is why this improper logic wasn't caught previously.

Most of the reason we even allow for localhost to be used is for testing, so this going wrong is kind of like I made an error in the test code. I had too much confidence in my tests passing and didn't actually run the code and seeing if it worked.

sbillinge:
thanks @zthatch this is great.  I am going to merge this now for further testing.  The develop branch is too far behind master and I pretty confident it is not breaking anything so let's merge and do more "field testing" from there.  You can close the prum.....congrats!

Did we decide the next step was to try and refactor todos helper to make direct queries?  I think so.  This will be hard if we maintain backwards compatibility with the fs client (which I think we must).  I think this will be VERY interesting to see how it works.  Please can you make yourself a prum for this but I want to stay closely involved.  We know that my current todo list (maybe I will tag it!) takes about 7-8 s to run from either fs or mongo....would love to get it to 0.7s on the mongo backend....

PR #793, opened 2021-07-13
getting dates to work in publist builder

sbillinge:


sbillinge:
this should have been handled better with more recent patches

PR #795, opened 2021-07-13
FIX: Mongo wasn't utilizing whitelist/blacklist feature

zthatch:
Mongo was loading all databases in the list every time. Tests pass locally, including mongo helpers.

### DIFF
--- a/regolith/builders/activitylogbuilder.py
+++ b/regolith/builders/activitylogbuilder.py
@@ -32,7 +32,7 @@ class ActivitylogBuilder(LatexBuilderBase):
     """Build CV from database entries"""
 
     btype = "annual-activity"
-    needed_dbs = ['groups', 'people', 'grants', 'proposals', 'institutions',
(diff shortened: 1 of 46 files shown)


regolith/builders/activitylogbuilder.py
```diff
@@ -32,7 +32,7 @@ class ActivitylogBuilder(LatexBuilderBase):
     """Build CV from database entries"""
 
     btype = "annual-activity"
-    needed_dbs = ['groups', 'people', 'grants', 'proposals', 'institutions',
+    needed_cols = ['groups', 'people', 'grants', 'proposals', 'institutions',
```
sbillinge, line 35: I think `colls` (short for `collections`) might be better but not sure.  What do you think?

sbillinge (commented):
one comment....I think I prefer `needed_colls` ....

regolith/builders/activitylogbuilder.py
zthatch: can do

sbillinge:
thanks @zthatch

Issue #696, opened 2021-02-03
Abstract lister helper Location vs Institution Filter


priyankaanehra:
The lister filters presentations via user entered arguments. There is an issue with the --loc_inst argument in particular. 

There are several presentations that are located at 'Columbia' for 'institution' but also for 'location'. The program currently would look through any seminars or colloquiums and if it finds matches, only output those types. However, this is problematic if you wanted the 'Columbia' matches that search in 'location' or if you wanted both that are in 'institution' and 'location'.

Potential Ideas from Simon:
-we can fix up the database
-we could put a couple of lines of code and throw an exception if the builder finds both in an entry or even better would be something in the database validation that checks for this.  Database validation is done using Cerberus.

connorjbracy:
@pliu2781's PR to close this issue has seemingly passed CI and been verified. Consider closing this issue.

sbillinge:
haha, I guess that this tested that pattern "closes issue#696" and it doesn't work.....  NP.,I will close.  The fix was in #796

PR #796, opened 2021-07-13
Fixes issue #696

pliu2781:
I edited the code so that, if l_abstract passes a --loc_inst to a presentation that has both 'location' and 'institution', then location will take priority. If a presentation only has 'institution', then the program will only lists presentations that are seminars/ colliquims

### DIFF
--- a/regolith/helpers/l_abstracthelper.py
+++ b/regolith/helpers/l_abstracthelper.py
@@ -98,10 +98,10 @@ class AbstractListerHelper(SoutHelperBase):
         if rc.loc_inst:
             filtered_inst = [presentation for presentation in presentations
                              if presentation.get('type') in SEMINAR_TYPES and
-                             rc.loc_inst.casefold() in presentation.get('institution',"").casefold()]
+                             rc.loc_inst.casefold() in presentation.get('institution',"").casefold()
+                             and rc.loc_inst.casefold() not in presentation.get('location',"")]
             filtered_loc = [presentation for presentation in presentations
-                            if rc.loc_inst.casefold() in presentation.get('location', 'institution').casefold()
-                            and rc.loc_inst.casefold() not in presentation.get('institution').casefold()]
+                            if rc.loc_inst.casefold() in presentation.get('location',"").casefold()]
 
         filtered_presentations_by_args = [filtered_inst, filtered_years, filtered_title,
                                           filtered_authors, filtered_loc]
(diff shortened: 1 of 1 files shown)


sbillinge:
I think we want it to throw an exception if it finds both.  We don't want to introduce magic.

btw, something is still wrong with travis...it is not triggering builds.

pliu2781:
I edited the code so that it displays a warning for presentations that have loc_inst in both location and institution. Then it prints those presentations.

pliu2781:
Is the use of warnings okay? Or should I have it throw an exception instead?

sbillinge:
exceptio

> 
> 
> Is the use of warnings okay? Or should I have it throw an exception instead?

exception.  Then it won't pass CI and the person will have to fix the problem.

regolith/helpers/l_abstracthelper.py
```diff
@@ -103,18 +102,15 @@ def sout(self):
...
+                            if rc.loc_inst.casefold() in presentation.get('location',"").casefold()]
             if exceptions:
-                warn("These presentations have %s in both 'location' and 'institution'" % (rc.loc_inst))
-            for presentation in exceptions:
-                print("---------------------------------------")
-                print(f"Title: {presentation.get('title')}")
-                print(f"location: {presentation.get('location')}")
-                print(f"institution: {presentation.get('institution')}")
+                for presentation in exceptions:
+                    raise Exception("This presentation has %s in both 'location' and 'institution'" % (rc.loc_inst),
```
sbillinge, line 110: - [ ] use an `fstring`
- [ ] raise ValueError
- [ ] tell the user the offending presentation, not just "This presentation" if you can
- [ ] move it up tp the top so it takes less time to crash if it is going to crash
- [ ] it is ok for this to be here (this is a lister), but it is more useful making sure the adder/updaters don't allow it in the first place.  They may not, but now is a great time to check while we are thinking about it.
- [ ] it needs a test

One more thought.  If we are even more clever we can do this check in the database validation.  google for cerberus and "conditional validation" or sthg like that.  Then we don't need any of htis code, which is kind of ugly anyway....

That is much more elegant a solution so worth looking into then we can just remove all this.

sbillinge (commented):
some thougthts, but at the end an idea that might be better....

regolith/helpers/l_abstracthelper.py
pliu2781: Should I implement the cerberus solution in l_abstracthelper.py or in schemas.py?

sbillinge: schemas.py.  It may also be necessary to tweak `regolith.validate`

pliu2781:
Closing this PR, as the cerberus fix makes this redundant

PR #798, opened 2021-07-14
Fixed issues regarding allowed doi entries and how different tags are handled

connorjbracy:
Updated allowed tags in the `doi` field of journal clubs. Now `na` is allowed and will appear as "N/A" in the html. Also, temporary journal article identifiers for arXiv papers are permitted (without being incorrectly cross-referenced through habanero) but will instead appear as simply the arXiv tag in the journal club entries rather than a full citation.


sbillinge:
nice,thanks

Issue #790, opened 2021-07-09
u_milestone bug


connorjbracy:
Attempting to run u_milestone with fields:
`$ helper_gui --ignore-gooey u_milestone -i 4 --notes "(x) Merge PR containing CI config files""(x) Meet with sbillinge to present/go through the setup process" -- cb_ci_for_getx3`

on `rg-db-group - commit db4217d9a8919728918e9134739e1969a552af10`

produces the following error

```
The milestone Write Setup Instructions for Setting Up CI on pdfgetx3 has been marked as finished in prum cb_ci_for_getx3
Traceback (most recent call last):
  File "~/anaconda3/envs/regolith_env/bin/helper_gui", line 4, in <module>
    __import__('pkg_resources').run_script('regolith==0.5.1', 'helper_gui')
  File "~/anaconda3/envs/regolith_env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 651, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "~/anaconda3/envs/regolith_env/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1448, in run_script
    exec(code, namespace, namespace)
  File "~/anaconda3/envs/regolith_env/lib/python3.7/site-packages/regolith-0.5.1-py3.7.egg/EGG-INFO/scripts/helper_gui", line 3, in <module>
    main()
  File "~/anaconda3/envs/regolith_env/lib/python3.7/site-packages/regolith-0.5.1-py3.7.egg/regolith/helper_gui_main.py", line 68, in main
    CONNECTED_COMMANDS[rc.cmd](rc)
  File "~/anaconda3/envs/regolith_env/lib/python3.7/site-packages/regolith-0.5.1-py3.7.egg/regolith/commands.py", line 150, in helper
    hlpr.hlp()
  File "~/anaconda3/envs/regolith_env/lib/python3.7/site-packages/regolith-0.5.1-py3.7.egg/regolith/helpers/basehelper.py", line 92, in hlp
    getattr(self, cmd)()
  File "~/anaconda3/envs/regolith_env/lib/python3.7/site-packages/regolith-0.5.1-py3.7.egg/regolith/helpers/u_milestonehelper.py", line 272, in db_updater
    pdoc[identifier].pop('identifier', None)
TypeError: pop() takes at most 1 argument (2 given)
```

sbillinge:
closed by #805

PR #802, opened 2021-07-15
Fixing issue breaking u_milestone

connorjbracy:
Closes [Issue #790](https://github.com/regro/regolith/issues/790#issue-940942427)

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -269,7 +269,8 @@ class MilestoneUpdaterHelper(DbHelperBase):
                     pdoc.update({'milestones': new_mil})
                 else:
                     pdoc.update({identifier: doc})
-            pdoc[identifier].pop('identifier', None)
+            for i in pdoc[identifier]:
+                i.pop('identifier', None)
             rc.client.update_one(rc.database, rc.coll, filterid, pdoc)
         for i in all_milestones:
             i.pop('identifier', None)
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -269,8 +269,8 @@ class MilestoneUpdaterHelper(DbHelperBase):
                     pdoc.update({'milestones': new_mil})
                 else:
                     pdoc.update({identifier: doc})
-            for i in pdoc[identifier]:
-                i.pop('identifier', None)
+            #for i in pdoc[identifier]:
+            #    i.pop('identifier', None)
             rc.client.update_one(rc.database, rc.coll, filterid, pdoc)
         for i in all_milestones:
             i.pop('identifier', None)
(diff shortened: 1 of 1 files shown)


sbillinge:
do you know what the "identifier" and pop are trying to accomplish?  It is hard to fix without knowing that.  git blame suggests that it was code inserted by zach.

sbillinge:
@connorjbracy I almost have a fix for this.  If it passes tests I will throw it into a PR and close this one.

connorjbracy:
> 
> 
> do you know what the "identifier" and pop are trying to accomplish? It is hard to fix without knowing that. git blame suggests that it was code inserted by zach.

The line that I put into a `for...loop` was indeed added by Zach, but he did not "invent" the line. Given that it is not clear in-line why we want to delete `'identifier'`, I had hopped into a call with Zach last night before making the edit to see if he knew but it was also not clear to him. Although he is not sure why that logic is there in the first place, his motivation for making the change was that, in setting up mongoDB, he noticed there were times when a `pdoc['milestone']` dictionary did not have an `'identifier'` key, thus he changed the previous deletion method to a more safe one (although he also added a new delete line)

The edit(s) he made which introduced the line causing the error that opened [Issue #790](https://github.com/regro/regolith/issues/790#issue-940942427) can be seen in [PR #788](https://github.com/regro/regolith/commit/9a142ed512dd82b2a5cbce2ce04fd25a63b11c0a) which shows the edit to a safer delete on line 274-275 in `regolith/helpers/u_milestonehelper.py`, as well as adding a corresponding line at line 272 (which is the line causing the error that opened [Issue #790](https://github.com/regro/regolith/issues/790#issue-940942427)). The problem with the erroring line is that it is attempting to do the same thing of deleting `'identifier'` from a `'milestone'` dictionary, but here it's targeting _milestones_ within in `pdoc` (i.e., a list of milestone dictionaries) which results in line 272 effectively doing `list({'milestone': ...}).pop('identifier', None)` so I simply wrapped this in a loop to properly delete 'identifier' from each 'milestone' dictionary.

connorjbracy:
It appears that deleting the additional (incorrect) `.pop(...)` line (272) altogether fixed the issue (at least, as far as the pytests are concerned), although I'm not sure if this line is indeed needed for the work in getting mongoDB to work or not.

connorjbracy:
For previous commit, diffed with `upstream/master` without fetching so I did not catch some edits that had been made since the initial branch which resulted in the previous commit's code coverage errors. This has been remedied by fetching, pulling, and merging `upstream/master` into this branch.

sbillinge:
ok, it took me way too long to figure this one out, but the problem was a missing `deepcopy ` when we first created the `all_milestones` list.  I then had to fix all the fixes scattered through the code coming from this error, some of which had been baked into the tests.  I then did a bit more cleaning and increased test coverage just a bit, so the code emerges better!!!! 

I will close this PR.  The issue is closed by the other merged PR.    

Thanks for your work on this.....

Issue #645, opened 2020-11-12
preslist builder dereferences authors from contacts collection as well as people


sbillinge:
rn it only looks in people to get the canonical name of the authors.   It needs also to be extended to look through contacts if it can't find a person in people, and only then default to printing the given name.

connorjbracy:
I think this might already be happening but the docstring wasn't updated to reflect this change. The lines I'll be referencing are within the current `master` branch but the "fix" that added this feature appears to have been [committed here](https://github.com/regro/regolith/commit/d8a3d585d3ad84412e78bdea1653e5b1d6d7bab9). Here's why I think this is already being handled (note that each link below is to the specific line that is being described, not the definitions of the function names that appear in the link):

- If I understand the issue correctly, we want the name that will appear in a citation to be searched for in both the `people` collection as well as the `contacts` collection. 
- The place where this search occurs is in [`regolith.tools.filter_presentations(...)`](https://github.com/regro/regolith/blob/f2edbfb0c8e4e9ba802ff98c5a4164ca515adb90/regolith/tools.py#L640) where each author's internal ID is converted to their citation name by using the result of [`fuzzy_retieval(...)`](https://github.com/regro/regolith/blob/f2edbfb0c8e4e9ba802ff98c5a4164ca515adb90/regolith/tools.py#L653) if the search was successful or defaulting to their [internal name](https://github.com/regro/regolith/blob/f2edbfb0c8e4e9ba802ff98c5a4164ca515adb90/regolith/tools.py#L645) if the search was not successful. 
- The `fuzzy_search(...)` [uses the `people` collection](https://github.com/regro/regolith/blob/f2edbfb0c8e4e9ba802ff98c5a4164ca515adb90/regolith/tools.py#L654) to get the citation name which is [passed in](https://github.com/regro/regolith/blob/f2edbfb0c8e4e9ba802ff98c5a4164ca515adb90/regolith/tools.py#L543) to the `regolith.tools.filter_presentation(...)` call via the first argument (referred to as `people`).
- In the context of writing the presentation lists, `regolith.tools.filter_presentation(...)` is called from [`regolith.builders.preslistbuilder.latex(...)`](https://github.com/regro/regolith/blob/f2edbfb0c8e4e9ba802ff98c5a4164ca515adb90/regolith/builders/preslistbuilder.py#L86) which passes a variable named `everybody` in for the `people` argument.
[...]

PR #808, opened 2021-07-16
Updated in-line documentation to reflect how an author name's is obtained for citation

connorjbracy:
Updating docstring to reflect changes in how a group member's ID is converted to the name that will appear in a publication citation as discussed in Issue #645


sbillinge:
Thanks for this.  Super!

PR #821, opened 2021-07-28
adding a whitespace

pliu2781:
Let's see if this passes on travis

### DIFF
--- a/regolith/helpers/l_progressreporthelper.py
+++ b/regolith/helpers/l_progressreporthelper.py
@@ -11,7 +11,7 @@ from regolith.tools import (
     get_pi_id,
     key_value_pair_filter,
 )
-from gooey import GooeyParser 
+from gooey import GooeyParser
 
 TARGET_COLL = "projecta"
 HELPER_TARGET = "l_progress"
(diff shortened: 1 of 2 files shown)


pliu2781:
This branch also has the same travis error

sbillinge:
ok, I will run the same travis test from my fork....

sbillinge:
this failed for me too.  We will have to track down why this is failing.  It seems that an assert is looking for "Errors found in " and gets back "".  We need to track down what module or package prints "Errors found in " and why it is not longer printing it, even though it seems to be there in the stdout that is captured in the test.

pliu2781:
Maybe Travis is using the wrong subprocess library? test_validate.py is using the xonsh.lib subprocess library.

sbillinge:
Hi @pliu2781 I am afraid we may need to get some help from Zach/Connor on this.  Can you reach out to them?   My best guess i that how the errors are passed back from the validators is changing with either a change in Cerberus (the json validation code we use) or a python version bump).  To recreate the bug we could try making a new local environment that is using the same version of python and cerberus as Travis is.   I am not sure to find out what Travis is using, but try making a new environment using the latest python and the latest cerberus.... That is my best suggestion at this point, but we really need it passing again....

PR #825, opened 2021-07-29
Fixing CI Bug

connorjbracy:
It seems two of the tests in  `tests/test_validate.py` were using an old syntax for obtaining/validating the output of a `subprocess.run(...)` call through the `xonsh` library


sbillinge:
thanks so much for finding that!

PR #826, opened 2021-08-02
cherry picking commits from other issue#814 PR

pliu2781:


pliu2781:
For some reason, travis isn't running

pliu2781:
TravisCI says these builds aren't running because: Owner regro does not have enough credits.

sbillinge:
OK, thanks.  I guess we just have to wait for credits to renew.

pliu2781:
When credits renew, I will try to do this cherry pick process again on a new PR

PR #832, opened 2021-08-05
Helperizing attestations, closes #827

### LINKED ISSUE #827: ENH: attestations -> helper
implement simon's attestations.py as a helper

pliu2781:
closes #827

### DIFF
--- a/regolith/helpers/attestationshelper.py
+++ b/regolith/helpers/attestationshelper.py
@@ -1,0 +1,259 @@
+# coding: utf-8
+# print([k for k,v in chained_db['people'].items()])
+
+# to use, set begin and end dates then "> python effortReporting"
+# fixme: make script for doing the appointments so I don't have to edit the db by hand
+import os
+
+from datetime import datetime, timedelta
+import dateutil.parser as date_parser
+import matplotlib.pyplot as plt
+import numpy as np
+
+from regolith.runcontrol import DEFAULT_RC, load_rcfile, filter_databases, \
+    connect_db
+from regolith.dates import month_to_int, get_dates
+from regolith.tools import (
+    fuzzy_retrieval,
+    all_docs_from_collection,
+    get_pi_id,
+    merge_collections_superior
+)
+from regolith.fsclient import _id_key
+from regolith.sorters import position_key
+from regolith.helpers.basehelper import DbHelperBase
+# print([k for k,v in chained_db['people'].items()])
+from pprint import pprint
+
+MONTH_COST = 3400
+
+# rc = DEFAULT_RC
+# if os.path.exists(rc.user_config):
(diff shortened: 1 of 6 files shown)


sbillinge:
@pliu2781 is there progress on this?

pliu2781:
> @pliu2781 is there progress on this?

Yes, I am still sorting things out with the tests, because I had to add to the exemplars in schemas.py

pliu2781:
closes #827

sbillinge:
> 
> 
> closes #827

for everything to work properly, I believe that this should be added to the very first comment/description.  You can click ... and hit edit and and copy-paste this there.

regolith/helpers/attestationshelper.py
```diff
@@ -123,21 +122,19 @@ def db_updater(self):
...
-                                ed = get_dates(p_appt).get("end_date")
-                                end = max(end, ed)
-                                months_on_grant = (ed - bd).days / 30.4 * loading
-                                appt = (person['_id'], bd, ed, loading, months_on_grant)
-                                appts.append(appt)
-            #                    on_grant = []
+                for _id, p_appt in person_appts.items():
+                    grantid = p_appt.get('grant')
+                    if grantid == grant_id:
+                        #accountnr = grant.get('account', grant['alias'])
```
sbillinge, line 128: delete things we don't use.  You can do this at the end when we are sure we don't need it.

regolith/schemas.py
```diff
@@ -231,6 +231,14 @@
             for i in range(1, 11)
         ],
         "payee": "scopatz",
+        "reimbursements": [
```
sbillinge, line 234: we should probably think about having a test for a non tbd reimbursement to cover everything, but decide if that is too much work and is worth it.

tests/test_helpers.py
```diff
@@ -7,6 +7,33 @@
 
 dash = "-"
 helper_map = [
+    (["helper", "attestations", "dmref15", "-b", "2019-09-01", "-e", "2019-11-01", "--no-plot"],
```
sbillinge, line 10: we will almost certainly have to pass in a date for testing, or the code will default to `today` on the day the test is run and the behavior will change over time.  Not sure, but that is usually the case.

sbillinge (commented):
this looks great!  Please see few comments

regolith/schemas.py
pliu2781: I added a non tbd reimbursement to test out

tests/test_helpers.py
pliu2781: If no begin/end_date is passed, the program just gets the begin/end_date from get_dates(grant)['begin/end_date']

sbillinge:
Thanks so much, this is great!  I will check it out and test!

PR #834, opened 2021-08-09
changed positions on webpage to title case

chloeann95:
I think the change is as simple as this. It is also labeled as .capitalize on the person's page instead of title case, but since it is in a heading font it overwrites the capitalization to make everything capital.


sbillinge:
Thanks @chloeann95 .  This should fix it.  I wondered why it still passed tests after the change but both of the tests were position = Professor where .capitalize and .titlecase would return the same!   Thanks for the fix.

PR #835, opened 2021-08-09
cherry picking 2nd commit from editing l_progress helper to fix issue#814

pliu2781:


pliu2781:
Something in this commit is causing the mongo-fs tests to fail.  I will try to find where it is

regolith/schemas.py
```diff
@@ -10,17 +11,20 @@
 
 ACTIVITIES_TYPES = ["teaching", "research"]
 AGENCIES = ["nsf", "doe", "other"]
+ALLOWED_STATI = ["proposed", "started", "finished", "back_burner", "paused", "cancelled"]
```
sbillinge, line 14: please remove.   These comments may have gotten lost on another PR perhaps, but this is not the right pattern.  You should be importing the `PROJECTA` stati into the progress report.  Don't duplicate them with new non-specific names.

sbillinge (commented):
thanks so much for adding these tests, it is great!  because it is alister, we can, in principle, add multiple tests to test all lines of code.  Codecov report can help to show which lines are not covered.

I also left a comment or two inline.

pliu2781:
I found the travis issue, it was a datetime object that I put in schemas that caused something to fail. I fixed it on the original PR, so I am closing this one

PR #850, opened 2021-11-03
Added __init__.py to make regolith.helpers a Python module

connorjbracy:


sbillinge:
wow, how did you figure that out!  Great  Catch!

PR #851, opened 2021-11-04
Cbracy revert regolith to v0.6.0

connorjbracy:
Series of commits to unwind previously made commits with undesired changes while still preserving integrity of the repo history

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,11 +4,6 @@ Regolith Change Log
 
 .. current developments
 
-v0.6.1
-====================
-
-
-
 v0.6.0
 ====================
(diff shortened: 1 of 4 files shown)


sbillinge:
Thanks @connorjbracy .  Are you doing it in two steps on purpose?  Fine if so, but we do also want to revert 0.6.0 back to the last commit before that.....just in case.

sbillinge:
Hi @connorjbracy what is the status on this.  Did you complete the work to roll this all the way back to 5.1?  I want to assign a regolith project to Michael aso we need this work to be finished before I can do that, so it is rather high priority I guess.

Issue #878, opened 2022-02-23
strip() id's before testing if they are valid in listers and adders etc.


sbillinge:
its annoying if we can't find a legit id because of some invisible whitespace.

PR #900, opened 2022-04-08
add clean_whitespace option to fuzzy retrieval function

ryboselm:
do not merge yet, i made this as a starting point for issue 878 (https://github.com/regro/regolith/issues/878), still figuring out what else needs to be modified outside of this one function

### DIFF
--- a/regolith/tools.py
+++ b/regolith/tools.py
@@ -916,7 +916,7 @@ def document_by_value(documents, address, value):
             return g_doc
 
 
-def fuzzy_retrieval(documents, sources, value, case_sensitive=True):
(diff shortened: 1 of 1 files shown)


sbillinge (commented):
before coding the solution,  code some tests.  You will find the tests for fuzzy_retrieval in `tests/test_tools.py`.  Then code the solution until the tests pass.  Add a test for each possible combination.

ryboselm:
I don't think the change I made in this commit actually solves anything related to issue #878. closing the PR because of that. will make a new one that actually fixes it.

Issue #914, opened 2022-06-06
u_milesone x's out all the unfinished todo items when it "finishes" a milestone


sbillinge:
u_milestone can be used to finish milestones.  It sets the status to finished and it adds an end_date.

These days we tend to put lists of todos in the Notes field and they look like:
```
Notes:
  - () do this
  - () do that
```
These can be checked off manually by the owner as they are completed, but more or less by definition, when the milestone is completed, all the tasks must have been done, so let's have the "finisher" to turn all the `()` into `(x)` automatically so it looks like
```
Notes:
  - (x) do this
  - (x) do that
 ```

mkwang2024:
kwang working on it

sbillinge:
closed by #915

PR #915, opened 2022-06-07
Fix issue #914

mkwang2024:
going to test regolith

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -225,12 +225,8 @@ class MilestoneUpdaterHelper(DbHelperBase):
                     rc.status = "finished"
                     doc.update({'end_date': now})
                     notes = doc.get("notes", [])
-                    for i in range(len(notes)): 
-                        note = notes[i]
-                        parenthesis_index = note.find('()')
-                        if parenthesis_index != -1:
-                            notes[i] = note[:parenthesis_index+1] + 'x' + note[parenthesis_index+1:]
-                    doc["notes"] = notes                 
+                    notes_with_closed_items = [note.replace('()', '(x)', 1) for note in notes]
+                    doc["notes"] = notes_with_closed_items              
                     if identifier == 'deliverable':
                         name = 'deliverable'
                     else:
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -224,6 +224,13 @@ class MilestoneUpdaterHelper(DbHelperBase):
                 if rc.finish:
                     rc.status = "finished"
                     doc.update({'end_date': now})
+                    notes = doc.get("notes", [])
+                    for i in range(len(notes)): 
+                        note = notes[i]
+                        parenthesis_index = note.find('()')
+                        if parenthesis_index != -1:
+                            notes[i] = note[:parenthesis_index+1] + 'x' + note[parenthesis_index+1:]
+                    doc["notes"] = notes                 
                     if identifier == 'deliverable':
                         name = 'deliverable'
                     else:
--- a/news/u_milestone_finish.rst
+++ b/news/u_milestone_finish.rst
@@ -1,0 +1,12 @@
+**Added:** 
+ * u_milesone now x's out all the unfinished todo items when it "finishes" a milestone
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None


regolith/helpers/u_milestonehelper.py
```diff
@@ -224,6 +224,13 @@ def db_updater(self):
                 if rc.finish:
                     rc.status = "finished"
                     doc.update({'end_date': now})
+                    notes = doc.get("notes", [])
+                    for i in range(len(notes)): 
+                        note = notes[i]
+                        parenthesis_index = note.find('()')
+                        if parenthesis_index != -1:
+                            notes[i] = note[:parenthesis_index+1] + 'x' + note[parenthesis_index+1:]
```
sbillinge, line 232: always prefer f-strings to concatenating strings with +

regolith/helpers/u_milestonehelper.py
```diff
@@ -224,6 +224,13 @@ def db_updater(self):
                 if rc.finish:
                     rc.status = "finished"
                     doc.update({'end_date': now})
+                    notes = doc.get("notes", [])
+                    for i in range(len(notes)):
```
sbillinge, line 228: This is not pythonic at all. 
```
for note in notes:
```
is much more readable.   Then maybe use the string replace method, so something like
```
note.replace('()', '(x)')
```
Finally, modern python loves list comprehensions, so something like
```
notes_with_closed_items = [note.replace('()', '(x)') for note in notes]
```
something like that (I didn't test it worked, but something along those lines will work)

sbillinge (commented):
nice job.  please see comments.

Please add a news and a test

regolith/helpers/u_milestonehelper.py
mkwang2024: Done

regolith/helpers/u_milestonehelper.py
mkwang2024: Corrected, so no longer need to concatenate.

mkwang2024:
I added the change to news. When I tried to add test, however, the tests/test_helpers.py is a large collection of tests for different helpers on the same database file, so I can’t create a test case for u_milestone with “—finish” without disturbing everything else.

mkwang2024:
In fact I think the test_helpers.py does an assert after every helper test in, but can only change the final output files

mkwang2024:
I tried to change the final output files to what I think it is, and when I run the tests, it will pass the later tests but fail the earlier ones. And if I change it back, it will fail the later tests. I tried to create my own test file/function, but I can’t get it to work without drastically altering schema.py and EXEMPLAR and creating new directories.

sbillinge:
The testing in regolith is tricky for the adders/updaters.  We never really solved it properly.  As you figured out, the way it is done is that we check if the change to the output document is what is expected, but the challenge is that we can only have one output document which makes it very difficult to test different command line args.   If `--finish` is already not being tested, it is probably because of this situation.  In principle, there is a workaround that will allow us to do it but it is a much wider regolith level issue that needs to addressed (how to do that in general) so in this case we probably  don't want to go there.  Thanks for trying though.  We can leave this untested.  If you find an elegant way to address this problem we should look into it.

I would like it to be passing tests before I merge it, but for this I think we need to fix the `xonsh` `CalledProcessError` issue.  The most valuable thing would be for you to look into this.   This just came up (was working, now broken, due to a breaking update of (presumably) the xonsh dependency.  Anything you find on that would be helpful.  I think @connorjbracy and @N-G-Asker are also looking into this, so keep each other updated.

N-G-Asker:
I have a note out to @connorjbracy with my hypothesis for what's going wrong. Long in short, `subprocess` is also a python standard library module, which does possess the CalledProcessError attribute, so it's possible the interpreter is getting confused as it were by the `import` of xonsh.lib.subprocess (which may have had this attribute before an update) and the solution may just be to replace it with the standard library subprocess module

sbillinge:
> I have a note out to @connorjbracy with my hypothesis for what's going wrong.
> 
> Long in short, `subprocess` is also a python standard library module, which does in fact have the CalledProcessError attribute, so it's possible the interpreter is getting confused as it were by the `import` of xonsh.lib.subprocess (which may have had this attribute before an update). The solution may just be to replace that import with the standard library `subprocess` module, but I think this idea needs careful vetting by someone more experienced with this system!
> 
> Looking forward to teaming up to get to the bottom of this!

This all sounds good.  This is maybe a point to reach out to the xonsh team to ask them.  I am not sure the best way but it may be to raise an issue on their GH site, but in a tentative way that asks what changed from 0.11 to 0.12 that would cause this to break and was it intentional....   If it is unintentional we found a bug for them, and if we change our code we would have to unchange it when they push a fix.....

sbillinge:
to get this to pass tests merge a clean master.  Then I can merge this PR.

mkwang2024:
Done

sbillinge:
great job.

PR #916, opened 2022-06-08
WIP: closes issue #908 (checking "gitlab" box in a_presentation automatic…

kaylabim:
…ally creates a gitlab repo in talks)

Comments: 
- for authentication code, I used a personal key but did not include in code! 
- tested with gitlab box checked and unchecked 
- assumed that date is in the correct format (i.e. did not add checks for correct yyyy-mm-dd format)

### DIFF
--- a/news/closes_908.rst
+++ b/news/closes_908.rst
@@ -1,9 +1,9 @@
 **Added:**
  * functionality to tools that checks that required repository request information (e.g. request parameters and url) are defined in rc 
- * functionality to tools that checks that authentication tokens for API requests are defined in rc 
+ * functionality to tools that checks that private authentication tokens for API requests are defined in rc 
  * functionality to tools that creates a repository based on given repo info 
- * no_talk_repo box added to presentation helper gui 
  * functionality to presentation helper that creates a repo by calling on create_repo function in tools when no_talk_repo box is not checked 
+ * no_repo box/option added to presentation helper gui 
 
 **Changed:** None
 
--- a/news/closes_908.rst
+++ b/news/closes_908.rst
@@ -1,6 +1,9 @@
 **Added:**
- * no_talk_repo box added to presentation helper
- * functionality to automatically create talk repo in gitlab for presentation
+ * functionality to tools that checks that required repository request information (e.g. request parameters and url) are defined in rc 
+ * functionality to tools that checks that authentication tokens for API requests are defined in rc 
+ * functionality to tools that creates a repository based on given repo info 
+ * no_talk_repo box added to presentation helper gui 
+ * functionality to presentation helper that creates a repo by calling on create_repo function in tools when no_talk_repo box is not checked 
 
 **Changed:** None
 
--- a/news/closes_908.rst
+++ b/news/closes_908.rst
@@ -1,0 +1,13 @@
+**Added:**
+ * create_talk_repo box to presentation in presentations
+ * functionality to presentation adder to create a gitlab repo in talks 
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/news/closes_908.rst
+++ b/news/closes_908.rst
@@ -1,9 +1,9 @@
 **Added:**
- * functionality to tools that checks that required repository request information (e.g. request parameters and url) are defined in rc 
- * functionality to tools that checks that private authentication tokens for API requests are defined in rc 
- * functionality to tools that creates a repository based on given repo info 
+ * repo_info_complete in tools to check the existence of remote repositories before operating on them 
+ * token_info_complete in tools to check for the existence of the user's private API authentication token 
+ * create_repo in tools to create a repository in a given remote repository
  * functionality to presentation helper that creates a repo by calling on create_repo function in tools when no_talk_repo box is not checked 
- * no_repo box/option added to presentation helper gui 
+ * no_repo box/option added to presentation helper gui
 
 **Changed:** None
 
--- a/news/closes_908.rst
+++ b/news/closes_908.rst
@@ -1,6 +1,6 @@
 **Added:**
- * create_talk_repo box to presentation in presentations
- * functionality to presentation adder to create a gitlab repo in talks 
+ * no_talk_repo box added to presentation helper
+ * functionality to automatically create talk repo in gitlab for presentation
 
 **Changed:** None
 
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -227,7 +227,7 @@ class PresentationAdderHelper(DbHelperBase):
             rc.client.insert_one(rc.database, EXPENSES_COLL, edoc)
             print(f"{key} has been added in {EXPENSES_COLL}")
 
-        rc.repos[0]['talk_repo'][0]['params']['name'] = key
+        rc.repos['talk_repo']['params']['name'] = key
         if not rc.no_talk_repo:
-            create_repo(key, rc)
+            create_repo(key, 'talk_repo', 'gitlab_private_token', rc)
         return
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -195,9 +195,8 @@ class PresentationAdderHelper(DbHelperBase):
                      'authors': authors,
                      'begin_date': begin_date,
                      'end_date': end_date,
+                     'notes': rc.notes,
                      })
-        if rc.notes:
-            pdoc.update({"notes": rc.notes})
         if rc.presentation_url:
             pdoc.update({"presentation_url": rc.presentation_url})
         if rc.webinar:
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -14,7 +14,7 @@ from regolith.tools import (
     get_pi_id,
     add_to_google_calendar,
     google_cal_auth_flow,
-    create_talk_repo
+    create_repo
 )
 from gooey import GooeyParser
 
@@ -227,6 +227,7 @@ class PresentationAdderHelper(DbHelperBase):
             rc.client.insert_one(rc.database, EXPENSES_COLL, edoc)
             print(f"{key} has been added in {EXPENSES_COLL}")
 
+        rc.repos[0]['talk_repo'][0]['params']['name'] = key
         if not rc.no_talk_repo:
-            create_talk_repo(key, rc)
+            create_repo(key, rc)
         return
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -100,7 +100,7 @@ def subparser(subpi):
     subpi.add_argument("--no-cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
-    subpi.add_argument("-r", "--no-talk-repo",
+    subpi.add_argument("-r", "--no-repo",
                        help=f"Do not add the presentation to gitlab repo under talks",
                        action="store_true")
     return subpi
@@ -228,6 +228,6 @@ class PresentationAdderHelper(DbHelperBase):
             print(f"{key} has been added in {EXPENSES_COLL}")
 
         rc.repos['talk_repo']['params']['name'] = key
-        if not rc.no_talk_repo:
+        if not rc.no_repo:
             create_repo(key, 'talk_repo', 'gitlab_private_token', rc)
         return
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -52,13 +52,16 @@ def subparser(subpi):
                        help="the person presenting the presentation, used for presentation name,"
                             " defaults to name in user.config",
                        )
-    subpi.add_argument("-u", "--authors", nargs="+",
+    subpi.add_argument("--authors", nargs="+",
                        help="specify the authors of this presentation, "
                             "defaults to person submitting the presentation",
                        )
     subpi.add_argument("-g", "--grants", nargs="+",
                        help="grant, or grants (separated by spaces), that support this presentation. Defaults to tbd",
                        default="tbd"
+                       )
+    subpi.add_argument("-u", "--presentation-url",
+                       help="the url to the presentation, whether it is on Google Drive, GitHub or wherever",
                        )
     subpi.add_argument("-n", "--notes", nargs="+",
                        help="note or notes to be inserted as a list into the notes field, "
@@ -188,6 +191,8 @@ class PresentationAdderHelper(DbHelperBase):
                      })
         if rc.notes:
             pdoc.update({"notes": rc.notes})
+        if rc.presentation_url:
+            pdoc.update({"presentation_url": rc.presentation_url})
         if rc.webinar:
             rc.no_expense = True
             pdoc.update({"webinar": True})
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -5,8 +5,6 @@ import time
 import dateutil.parser as date_parser
 import threading
 
-import requests 
-
 from regolith.helpers.a_expensehelper import expense_constructor
 from regolith.helpers.basehelper import DbHelperBase
 from regolith.fsclient import _id_key
@@ -15,7 +13,8 @@ from regolith.tools import (
     all_docs_from_collection,
     get_pi_id,
     add_to_google_calendar,
-    google_cal_auth_flow
+    google_cal_auth_flow,
+    add_to_gitlab
 )
 from gooey import GooeyParser
 
@@ -100,8 +99,8 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
-    subpi.add_argument("--create_talk_repo",
-                       help=f"add presentation to gitlab repo under talks",
+    subpi.add_argument("-r", "--no_talk_repo",
+                       help=f"Do not add the presentation to gitlab repo under talks",
                        action="store_true")
     return subpi
 
@@ -228,32 +227,7 @@ class PresentationAdderHelper(DbHelperBase):
             rc.client.insert_one(rc.database, EXPENSES_COLL, edoc)
             print(f"{key} has been added in {EXPENSES_COLL}")
 
-        # if gitlab box is checked, creates gitlab repo in talks group
-
-        if rc.create_talk_repo:
-            headers = {
-                # TODO : store this in environment variable somehow
-                'PRIVATE-TOKEN': '<api_authentication_key>'
-            }
-
-            # namespace_id = 35 always because talks group id is 35
-            # name derived by combining last 2 digits of the begin_date year, the begin_date month, first 2
-            # letters of person, underscore, place
-
-            params = {
-                'name': key,
-                'namespace_id': '35',
-                'initialize_with_readme': 'false'
-            }
-
-            try:
-                response = requests.post(
-                    'https://gitlab.thebillingegroup.com/api/v4/projects/', headers=headers, json=params
-                )
-                print(f"repo {key} has been created in talks: https://gitlab.thebillingegroup.com/talks/{key}")
-            except requests.exceptions.RequestException:
-                print("ERROR: Issue with GitLab API call")
+        # creates gitlab repo under talks by default (when box is not checked)
+        if not rc.no_talk_repo:
+            add_to_gitlab(key)
         return
-
-
-            
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -5,6 +5,8 @@ import time
 import dateutil.parser as date_parser
 import threading
 
+import requests 
+
 from regolith.helpers.a_expensehelper import expense_constructor
 from regolith.helpers.basehelper import DbHelperBase
 from regolith.fsclient import _id_key
@@ -98,6 +100,9 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
+    subpi.add_argument("--gitlab", 
+                       help=f"add presentation to gitlab repo", 
+                       action="store_true")
     return subpi
 
 
@@ -129,6 +134,7 @@ class PresentationAdderHelper(DbHelperBase):
     def db_updater(self):
         gtx = self.gtx
         rc = self.rc
+
         if not rc.no_cal:
             event = {
                         'summary': rc.name,
@@ -145,7 +151,35 @@ class PresentationAdderHelper(DbHelperBase):
                     wait_bool = add_to_google_calendar(event)
                     jump += 1
 
+        # if gitlab box is checked, creates gitlab repo in talks group
+
+        if rc.gitlab:
+            headers = {
+                # TODO : store this in environment variable somehow
+                'PRIVATE-TOKEN': '<api_authentication_key>'
+            }
+
+            # namespace_id = 35 always because talks group id is 35
+            # name derived by combining last 2 digits of the begin_date year, the begin_date month, first 2
+            # letters of person, underscore, place
+
+            # TODO : assuming begin_date is in correct format and follows yyyy-mm-dd
+            params = {
+                'name': rc.begin_date.strip()[2:4] + rc.begin_date.strip()[5:7] + rc.person.strip()[0:2]
+                        + '_' + rc.place.strip(),
+                'namespace_id': '35',
+                'initialize_with_readme': 'false'
+            }
+
+            try:
+                response = requests.post(
+                    'https://gitlab.thebillingegroup.com/api/v4/projects/', headers=headers, json=params
+                )
+            except requests.exceptions.RequestException:
+                print("ERROR: Issue with GitLab API call")
+
         # dates
+        # TODO : add date format check?
         begin_date = date_parser.parse(rc.begin_date).date()
         end_date = date_parser.parse(rc.end_date).date()
 
@@ -171,6 +205,7 @@ class PresentationAdderHelper(DbHelperBase):
 
         coll = self.gtx[rc.coll]
         pdocl = list(filter(lambda doc: doc["_id"] == key, coll))
+        print(pdocl)
         if len(pdocl) > 0:
             raise RuntimeError(
                 "This entry appears to already exist in the collection")
@@ -220,5 +255,6 @@ class PresentationAdderHelper(DbHelperBase):
             edoc = expense_constructor(key, begin_date, end_date, rc)
             rc.client.insert_one(rc.database, EXPENSES_COLL, edoc)
             print(f"{key} has been added in {EXPENSES_COLL}")
-
         return
+    
+            
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -100,8 +100,8 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
-    subpi.add_argument("--gitlab", 
-                       help=f"add presentation to gitlab repo", 
+    subpi.add_argument("--create_talk_repo",
+                       help=f"add presentation to gitlab repo under talks",
                        action="store_true")
     return subpi
 
@@ -151,33 +151,6 @@ class PresentationAdderHelper(DbHelperBase):
                     wait_bool = add_to_google_calendar(event)
                     jump += 1
 
-        # if gitlab box is checked, creates gitlab repo in talks group
-
-        if rc.gitlab:
-            headers = {
-                # TODO : store this in environment variable somehow
-                'PRIVATE-TOKEN': '<api_authentication_key>'
-            }
-
-            # namespace_id = 35 always because talks group id is 35
-            # name derived by combining last 2 digits of the begin_date year, the begin_date month, first 2
-            # letters of person, underscore, place
-
-            # TODO : assuming begin_date is in correct format and follows yyyy-mm-dd
-            params = {
-                'name': rc.begin_date.strip()[2:4] + rc.begin_date.strip()[5:7] + rc.person.strip()[0:2]
-                        + '_' + rc.place.strip(),
-                'namespace_id': '35',
-                'initialize_with_readme': 'false'
-            }
-
-            try:
-                response = requests.post(
-                    'https://gitlab.thebillingegroup.com/api/v4/projects/', headers=headers, json=params
-                )
-            except requests.exceptions.RequestException:
-                print("ERROR: Issue with GitLab API call")
-
         # dates
         # TODO : add date format check?
         begin_date = date_parser.parse(rc.begin_date).date()
@@ -205,7 +178,6 @@ class PresentationAdderHelper(DbHelperBase):
 
         coll = self.gtx[rc.coll]
         pdocl = list(filter(lambda doc: doc["_id"] == key, coll))
-        print(pdocl)
         if len(pdocl) > 0:
             raise RuntimeError(
                 "This entry appears to already exist in the collection")
@@ -255,6 +227,33 @@ class PresentationAdderHelper(DbHelperBase):
             edoc = expense_constructor(key, begin_date, end_date, rc)
             rc.client.insert_one(rc.database, EXPENSES_COLL, edoc)
             print(f"{key} has been added in {EXPENSES_COLL}")
+
+        # if gitlab box is checked, creates gitlab repo in talks group
+
+        if rc.create_talk_repo:
+            headers = {
+                # TODO : store this in environment variable somehow
+                'PRIVATE-TOKEN': '<api_authentication_key>'
+            }
+
+            # namespace_id = 35 always because talks group id is 35
+            # name derived by combining last 2 digits of the begin_date year, the begin_date month, first 2
+            # letters of person, underscore, place
+
+            params = {
+                'name': key,
+                'namespace_id': '35',
+                'initialize_with_readme': 'false'
+            }
+
+            try:
+                response = requests.post(
+                    'https://gitlab.thebillingegroup.com/api/v4/projects/', headers=headers, json=params
+                )
+                print(f"repo {key} has been created in talks: https://gitlab.thebillingegroup.com/talks/{key}")
+            except requests.exceptions.RequestException:
+                print("ERROR: Issue with GitLab API call")
         return
-    
+
+
             
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -14,7 +14,7 @@ from regolith.tools import (
     get_pi_id,
     add_to_google_calendar,
     google_cal_auth_flow,
-    add_to_gitlab
+    create_talk_repo
 )
 from gooey import GooeyParser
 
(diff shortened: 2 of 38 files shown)


regolith/helpers/a_presentationhelper.py
```diff
@@ -145,7 +151,35 @@ def db_updater(self):
...
+                'PRIVATE-TOKEN': '<api_authentication_key>'
+            }
+
+            # namespace_id = 35 always because talks group id is 35
+            # name derived by combining last 2 digits of the begin_date year, the begin_date month, first 2
+            # letters of person, underscore, place
+
+            # TODO : assuming begin_date is in correct format and follows yyyy-mm-dd
+            params = {
+                'name': rc.begin_date.strip()[2:4] + rc.begin_date.strip()[5:7] + rc.person.strip()[0:2]
```
sbillinge, line 168: I think we create a key for the presentation.  Can we reuse the key for the repo-name?

regolith/helpers/a_presentationhelper.py
```diff
@@ -145,7 +151,35 @@ def db_updater(self):
                     wait_bool = add_to_google_calendar(event)
                     jump += 1
 
+        # if gitlab box is checked, creates gitlab repo in talks group
+
+        if rc.gitlab:
+            headers = {
+                # TODO : store this in environment variable somehow
+                'PRIVATE-TOKEN': '<api_authentication_key>'
```
sbillinge, line 159: The way we are doing this now is that each user has a `user.json` in `~/.config/regolith/`

These are typically lying unencrypted so we may want to revisit how we do this, but for now it is ok.

regolith/helpers/a_presentationhelper.py
```diff
@@ -145,7 +151,35 @@ def db_updater(self):
...
+
+            # namespace_id = 35 always because talks group id is 35
+            # name derived by combining last 2 digits of the begin_date year, the begin_date month, first 2
+            # letters of person, underscore, place
+
+            # TODO : assuming begin_date is in correct format and follows yyyy-mm-dd
+            params = {
+                'name': rc.begin_date.strip()[2:4] + rc.begin_date.strip()[5:7] + rc.person.strip()[0:2]
+                        + '_' + rc.place.strip(),
+                'namespace_id': '35',
```
sbillinge, line 170: this shouldn't be hard-coded.  Somehow all the information about the URL of the target repo and the `namespace_id` should be stored in the run-control json file, like the database addresses themselves.

regolith/helpers/a_presentationhelper.py
```diff
@@ -98,6 +100,9 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
+    subpi.add_argument("--gitlab",
```
sbillinge, line 103: this should probably be more general.  What if people store their talk files in github?  Something like `--create_talk_repo`? 

I think for me the best behavior is if I can set this value in my global config (user.json) so it does it by default and this button would turn it off, but others may want it off by default.   Not sure how to do that, so maybe think about it.

sbillinge (commented):
This is really great, nice work!  please see inline comments.  Please check that your PR title will automatically close the issue.  For example, I know that putting `closes #908` in the PR description will automatically close the issue on merging but I am not sure if `closing issue #908` will do so.

We will need a "news" item.  Copy the `template.rst` in the `news` folder to `<branch_name>.rst` and fill it in.   The changelog is built from this.

Is there a way to write a test for this?

sbillinge:
@kaylabim it is failing two tests.   One way the helpers at tested is that stdout from the code is compared to what is expected and you added a print statement.  it could be a debugging print so remove it, or perhaps we want the code to "chat" by alerting the user that a repo of a certain name has been created (and possibly printing the link that can be clicked?).  In this latter case, update the expected output in the test to reflect this.

regolith/helpers/a_presentationhelper.py
kaylabim: I am having some trouble figuring out how to do this! I haven't worked with json files and API requests.

regolith/helpers/a_presentationhelper.py
```diff
@@ -98,6 +100,9 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
+    subpi.add_argument("--create_talk_repo",
```
sbillinge, line 103: if you can, is there a single letter left that is intuitive to save typing?

regolith/helpers/a_presentationhelper.py
```diff
@@ -221,4 +228,32 @@ def db_updater(self):
...
+                'PRIVATE-TOKEN': '<api_authentication_key>'
+            }
+
+            # namespace_id = 35 always because talks group id is 35
+            # name derived by combining last 2 digits of the begin_date year, the begin_date month, first 2
+            # letters of person, underscore, place
+
+            params = {
+                'name': key,
+                'namespace_id': '35',
```
sbillinge, line 245: move to rc

regolith/helpers/a_presentationhelper.py
```diff
@@ -221,4 +228,32 @@ def db_updater(self):
...
+                'initialize_with_readme': 'false'
+            }
+
+            try:
+                response = requests.post(
+                    'https://gitlab.thebillingegroup.com/api/v4/projects/', headers=headers, json=params
```
sbillinge, line 251: move to rc

regolith/helpers/a_presentationhelper.py
```diff
@@ -221,4 +228,32 @@ def db_updater(self):
...
+                )
+                print(f"repo {key} has been created in talks: https://gitlab.thebillingegroup.com/talks/{key}")
```
sbillinge, line 253: update after move to rc

regolith/helpers/a_presentationhelper.py
```diff
@@ -221,4 +228,32 @@ def db_updater(self):
...
+            except requests.exceptions.RequestException:
+                print("ERROR: Issue with GitLab API call")
         return
+
+
```
sbillinge, line 258: clean up emtpy lines.

regolith/helpers/a_presentationhelper.py
```diff
@@ -98,6 +100,9 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
+    subpi.add_argument("--create_talk_repo",
+                       help=f"add presentation to gitlab repo under talks",
+                       action="store_true")
```
sbillinge, line 105: I would rather that the default be true, so turn off if you don't want a talk repo to be create.  Better yet, if the talk url is in the rc then do it, but if it isn't then don't but print an informational message?  What do you think?

sbillinge (commented):
please see comments.

kaylabim:
Stored the URL info in a rc json file: ~/.config/regolith/tokens/gitlab_api/token.json

{"headers": {"PRIVATE-TOKEN": "<gitlab-personal-access-token>"}, "params": {"namespace_id": "35", "initialize_with_readme": "false", "name": ""}, "url": "https://gitlab.thebillingegroup.com/api/v4/projects/"}

regolith/tools.py
```diff
@@ -2102,6 +2104,40 @@ def google_cal_auth_flow():
...
+    tokendir = os.path.expanduser("~/.config/regolith/tokens/gitlab_api")
+    os.makedirs(tokendir, exist_ok=True)
+    tokenfile = os.path.join(tokendir, 'token.json')
+    data = {}
+    # print(tokenfile)
+    if os.path.exists(tokenfile):
+        # if it exists, grab the personal access token
+        with open(tokenfile) as fp:
+            data = json.load(fp)
+
```
sbillinge, line 2126: `else` ?

Think about what behavior the user might want first. If you like, to save time, we can discuss that.

regolith/tools.py
```diff
@@ -2102,6 +2104,40 @@ def google_cal_auth_flow():
...
+        try:
+            response = requests.post(data['url'], headers=data['headers'], params=data['params'])
+            with open(tokenfile, 'w') as f:
+                json.dump(data, f)
+            if response.status_code in {200, 201}:
+                print(f"repo {name} has been created in talks: https://gitlab.thebillingegroup.com/talks/{name}")
+        except requests.exceptions.RequestException:
+            print("ERROR: Issue with GitLab API call")
+    else:
+        print("ERROR: Missing private access token (PRIVATE-TOKEN) in headers of token.json file for GitLab")
```
sbillinge, line 2139: only print ERROR messages on an error, i.e., when an exception raises.   See the general review about what behaviors I think would be good.

regolith/tools.py
```diff
@@ -2102,6 +2104,40 @@ def google_cal_auth_flow():
...
+            data = json.load(fp)
+
+    private_token = data.get('headers').get('PRIVATE-TOKEN')
+    if private_token:
+        data['params']['name'] = name
+        try:
+            response = requests.post(data['url'], headers=data['headers'], params=data['params'])
+            with open(tokenfile, 'w') as f:
+                json.dump(data, f)
+            if response.status_code in {200, 201}:
```
sbillinge, line 2134: should those be braces there?  It looks like a funny construction to me.

regolith/tools.py
```diff
@@ -2102,6 +2104,40 @@ def google_cal_auth_flow():
...
+                print(f"repo {name} has been created in talks: https://gitlab.thebillingegroup.com/talks/{name}")
```
sbillinge, line 2135: don't hard-code the path here

sbillinge (commented):
maybe we should have a quick convo about good behavior for the code.  One possibility is that the code checks for whether `talk_repo` is defined and valid in `rc`.  If it is, then it creates the talk repo, if it isn't then it doesn't.   We can then remove it from the parser.   If it isn't defined in rc we could generate an information message pointing to documentation how to set it up.  If it is defined but not valid generate a WARNING message but keep going?

I think everything except the token should be defined in regolithrc.json.  This way, for example, I could set it up once for `rg-db-group` for our group repo and everyone would automatically have it, just like the lists of databases.  But the token is individual to the person, so it should go in their tokens directory as you did.   

If the repo is valid but there is no token, trigger a message telling the person how to set it up.    If there is a token but it is bad, just have it fail with a helpful message.

Ideally the function can be made agnostic about whather the repo is in GH or GL, or in principle wherever.   If their api's are different we may need to have checks to see which it is.  To avoid this becoming a huge project, just implement the GL version, but still code it up with this in mind so everywhere refer to `repo_url` and so on rather than `gitlab_url`.

regolith/helpers/a_presentationhelper.py
```diff
@@ -98,6 +99,9 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
+    subpi.add_argument("-r", "--no_talk_repo",
```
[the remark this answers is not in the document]
kaylabim: Done

regolith/helpers/a_presentationhelper.py
```diff
@@ -221,4 +227,7 @@ def db_updater(self):
             rc.client.insert_one(rc.database, EXPENSES_COLL, edoc)
             print(f"{key} has been added in {EXPENSES_COLL}")
 
-        return
+        # creates gitlab repo under talks by default (when box is not checked)
```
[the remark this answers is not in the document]
kaylabim: Done

regolith/helpers/a_presentationhelper.py
```diff
@@ -221,4 +227,7 @@ def db_updater(self):
...
+        if not rc.no_talk_repo:
+            add_to_gitlab(key)
+        return
```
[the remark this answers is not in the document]
kaylabim: Done

regolith/tools.py
```diff
@@ -2102,6 +2104,40 @@ def google_cal_auth_flow():
...
+    Parameters:
+        name - the name/key of the event
+
+    Returns:
+        None
+    """
+
+    tokendir = os.path.expanduser("~/.config/regolith/tokens/gitlab_api")
+    os.makedirs(tokendir, exist_ok=True)
+    tokenfile = os.path.join(tokendir, 'token.json')
```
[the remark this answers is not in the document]
kaylabim: done

regolith/tools.py
kaylabim: Changed to []

kaylabim:
Moved url info to ~/dbs/regolith/local/regolithrc.json, stored private token in ~/.config/regolith/user.json, and updated the code's behavior based on feedback. 

I created test repos in talks (everything from/after klkim test 1) but I cannot delete because I lack permissions. Please help delete them!

news/closes_908.rst
```diff
@@ -1,6 +1,6 @@
 **Added:**
- * create_talk_repo box to presentation in presentations
- * functionality to presentation adder to create a gitlab repo in talks 
+ * no_talk_repo box added to presentation helper
```
sbillinge, line 2: I think this is gone with the new workflow, right?

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
         token.write(creds.to_json())
     # Save the credentials for the next run
 
-def add_to_gitlab(name):
-    """Takes a newly created event, and adds it to the user's google calendar
+def create_talk_repo(name, rc):
+    """Takes in the presentation name and creates a repo
```
sbillinge, line 2109: add full behavior. ...if repo information is specified in regolithrc.json.

Also, remove "presentation" from the docstring to make the function more reusable

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
 
     Parameters:
         name - the name/key of the event
```
sbillinge, line 2112: I think strictly you can pass any string here and it will be used, so I would write this maybe like this:
```
    name - string
      the name that will be used for the repository. e.g., the key of presentation/projectum

File: regolith/tools.py [hunk 22]
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
         token.write(creds.to_json())
     # Save the credentials for the next run
 
-def add_to_gitlab(name):
-    """Takes a newly created event, and adds it to the user's google calendar
+def create_talk_repo(name, rc):
```
sbillinge, line 2108: in principle this can be used to create any repo, not just a talk repo, right?  For example Jay has a project to create a repo for a prum.  I suggest change the name to `create_repo`

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+        rc - run control object
 
     Returns:
         None
     """
-
-    tokendir = os.path.expanduser("~/.config/regolith/tokens/gitlab_api")
-    os.makedirs(tokendir, exist_ok=True)
-    tokenfile = os.path.join(tokendir, 'token.json')
+    regorcdir = Path('~/dbs/rg-db-group/local')
```
sbillinge, line 2118: don't hard-code paths.

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
-        try:
-            response = requests.post(data['url'], headers=data['headers'], params=data['params'])
-            with open(tokenfile, 'w') as f:
-                json.dump(data, f)
-            if response.status_code in {200, 201}:
-                print(f"repo {name} has been created in talks: https://gitlab.thebillingegroup.com/talks/{name}")
-        except requests.exceptions.RequestException:
-            print("ERROR: Issue with GitLab API call")
+
+    if rc.talk_repo:
```
sbillinge, line 2122: we will have to think of a way to generalize this to make it reusable.  One way would be for it to be rc.repos and it can take a dict and in regolithrc.json you can specify more than one repo ({talk_repo: <repoinfo>, prum_repo: <repoinfo>})

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+        token = rc.gitlab_private_token
+        repo_url = rc.talk_repo[0]['url']
```
sbillinge: talk repo shouldn't be a list I think

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+        param = rc.talk_repo[0]['params']
+        if token:
+            with open(regorcfile) as fp:
+                data = json.load(fp)
```
sbillinge: don't load regolithrc again.  It is already loaded and passed as rc

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+            data['talk_repo'][0]['params']['name'] = name
+            try:
+                response = requests.post(repo_url, params=param, headers={'PRIVATE-TOKEN': '{}'.format(token)})
+                print(f"status_code: {response.status_code}")
+                with open(regorcfile, 'w') as fp:
```
sbillinge: why are you dumping?

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+        token = rc.gitlab_private_token
+        repo_url = rc.talk_repo[0]['url']
+        param = rc.talk_repo[0]['params']
+        if token:
+            with open(regorcfile) as fp:
+                data = json.load(fp)
+            data['talk_repo'][0]['params']['name'] = name
+            try:
+                response = requests.post(repo_url, params=param, headers={'PRIVATE-TOKEN': '{}'.format(token)})
+                print(f"status_code: {response.status_code}")
```
sbillinge: don't make the code too chatty.  Users generally don't like it.  What will a chemist in iowa do with this information for example.  If it is a debugging print that is ok, just remembe rto delete it before merging.

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+                with open(regorcfile, 'w') as fp:
+                    json.dump(data, fp)
+            except requests.exceptions.RequestException:
+                print("ERROR: Issue with GitLab API call. Check that your private token is valid."
+                      "\nTo create a private token, refer to these instructions: https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html")
```
sbillinge: this will not be available to people in general, only group members in the group, so don't put the wiki url in the code.

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+            finally:
```
sbillinge, line 2138: this construction is not so clear.  Why don't you just put the first print statement in the try clause and the second one in the except clause?

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+                if response.status_code in [200, 201]:
+                    print(f"repo {name} has been created in talks: {repo_url}{name}")
+                else:
+                    print("ERROR: Issue with GitLab API call. Check that your private token is valid.")
+        else:
+            print("Put your gitlab private token in user.json (in ~/.config/regolith)"
```
sbillinge: this is a bit too imperative (bossy).  Just reveal that this capability is present in a more kind way. "If you would like regolith to automatically create a repository in GitHub/GitLab please add repository information in regolithrc.json. See regolith documentation for details".  We will have to add it to the docs.

regolith/tools.py
```diff
@@ -2104,39 +2105,47 @@ def google_cal_auth_flow():
...
+                  "as the value for \"gitlab_private_token\". \nTo create a private "
+                  "token, refer to these instructions: https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html")
     else:
-        print("ERROR: Missing private access token (PRIVATE-TOKEN) in headers of token.json file for GitLab")
+        print("talk_repo")
```
sbillinge: I guess this will go...

sbillinge (commented):
please see inline comments.   You did really well and clearly got it going nicely.  My one request is for you to go even slower and think more before implementing.   It is more work to define the problem that we are solving than to solve it, and we can save a lot of time by thinking carefully before typing.  Who will use it, how will it be used? Can it be abstracted (used for more than one thing) at relatively low cost?  Does this make sense to do that?  I often discover the answers to these by typing the full docstring.  Your docstring is too short for you to have discovered anything by typing it.  This is not a criticism at all, you are doing really great, but if I can teach you this way of thinking and working now, early in the process, it will save me a log of typing during future reviews....and you will develop really dynamite habits that are completely transferrable.   When you work like this, questions will come up, feel free to throw them at me through slack, though don't get stuck because I can't respond immediately so it is a balance of taking the initiative an also seeking feedback early in the process.

regolith/tools.py
kaylabim: sounds good! done

kaylabim:
Thank you for your feedback and guidance! I will definitely start adopting a more detailed preplanning strategy/mindset before I start implementing code. I completely agree that starting with a clear goal in mind is super helpful (for me and the user) and a skill that I will work on developing for myself. 
I made changes based on the feedback (edited docstring, warning messages, only using rc to query info, etc.)

regolith/tools.py
```diff
@@ -2105,47 +2105,49 @@ def google_cal_auth_flow():
...
-                print("Issue with GitLab API call. Check that your private token is valid.")
-            finally:
+                response = requests.post(repo_url, params=param, headers={'PRIVATE-TOKEN': '{}'.format(token)})
+                response.raise_for_status()
                 if response.status_code in [200, 201]:
                     print(f"repo {name} has been created in talks")
-                else:
-                    print("Issue with GitLab API call. Check that your private token is valid.")
+            except requests.exceptions.HTTPError:
+                print(f"Issue with the API call (status code: {response.status_code}). "
```
sbillinge, line 2139: It is better for this to be passed to the exception, something like:
```
try:
    int("string") 
except ValueError:
    raise ValueError("Your custom message here.")
```
unless you want the code to skip over and keep running at this point.  In that case the print is ok but make the error message useful to the user (who may not be a programmer), something like: "WARNING: Unsuccessful attempt at making a GitHub/GitLab etc., repository.  Check the URL information in your config.json/regolithrc.json files to look for errors."  Printing the status code is also helpful information to give back.  

I guess the HttpError and the RequestException are raised for different issues.  In that case you can give more granular info in the message, like "can't find the an org" vs. "found the org but access token not working" kind of thing (but written better).

sbillinge (commented):
this is getting very close now.  There is still one hard-coded talk-repo.  I am not sure if there is an easy way to make it just repo?  One way would be to pass in the actual repo information.  e.g.,
```
repo = rc.repos["talk_repo"]
create_repo(name, repo, rc)
```
This would make this function more reusable.

Two other gripes.  You are not using test-driven development which is partly why we are iterating so much on this.   The preferred approach is:
1. Think what you think the function should do and create an empty function with what you think is the right name and the right inputs
2. generate the docstring (automatic when you type """ in PyCharm)
3. write the docstring in detail.  During this process you will almost certainly change what the function does because as you try and describe in detail what goes in and what the outcomes are you will realize that the function needs to do something a bit different than what you thought (this is the process we are going through now, but it is much more expensive because you have to change code instead of just change a docstring).
4. write tests that test test all the things
5. Only then write the function. so that it passes the tests
``
I would actually suggest that you create a new function from scratch this way.  change the name of this one temporarily so it isn't used anywhere (e.g., add `_old` on it).  You will reuse some things like the nice error messages.  Then start over using the workflow I described above.  When you get used to it you will find it very enabling and powerful.  To begin with you will think it is slowing you down, but later, it results in much better and more reusable code with fewer iterations later at the expensive end of the process.

tests/test_tools.py
```diff
@@ -2174,5 +2175,69 @@ def test_get_tags_invalid():
...
+        ([{"repos": {"talk_repo": {"params": {"namespace_id": "", "initialize_with_readme": "false", "name": ""},
+                         "url": "https://gitlab.thebillingegroup.com/api/v4/projects/"}, "prum_repo": {}}}], [0]),
+    ]
+)
+def test_validate_repo(inputs, expected):
+    from regolith.runcontrol import DEFAULT_RC, load_rcfile
+    import copy
+    # import sys
+    rc = copy.copy(DEFAULT_RC)
+    rc._update(inputs[0])
```
sbillinge, line 2204: inputs don't have to be a list because you only pass one input in to the function.  You can make the code more readable by giving it a more descriptive name too in this case.  So instead of inputs, maybe call it `repo_information`.  Also, since this is a test of an abstract function, calling your repo `talk_repo` is ok, but your target audience of a future developer may be scratching her head.  Think about something like:
```
{'repos': {
    'repo1':{
       'params': {
           'namespace_id': "",
           'initialize_with_readme': 'false',
           ...
                       }
                }
             }
  "repo2": {}
}}
```
Looking at this structure, I wonder what work `'params'` is doing.  Isn't it redundant?  Does `'repo1` have anything else than `params` or will it in the future.  What can you think of that repo1 has that is not a parameter?  OK, to leave it there if there is something, but better to remove it if not.

tests/test_tools.py
```diff
@@ -2174,5 +2175,69 @@ def test_get_tags_invalid():
...
+        # namespace_id is not defined
+        ([{"repos": {"talk_repo": {"params": {"namespace_id": "", "initialize_with_readme": "false", "name": "repo name"},
+                         "url": "https://gitlab.thebillingegroup.com/api/v4/projects/"}, "prum_repo": {}}}], [0]),
+        # name is not defined
+        ([{"repos": {"talk_repo": {"params": {"namespace_id": "", "initialize_with_readme": "false", "name": ""},
+                         "url": "https://gitlab.thebillingegroup.com/api/v4/projects/"}, "prum_repo": {}}}], [0]),
+    ]
+)
+def test_validate_repo(inputs, expected):
+    from regolith.runcontrol import DEFAULT_RC, load_rcfile
```
sbillinge, line 2200: place imports at the top of the module to avoid duplicated code.

tests/test_tools.py
```diff
@@ -2174,5 +2175,69 @@ def test_get_tags_invalid():
...
+        # everything is defined, return 1
+        ([{"repos": {"talk_repo": {"params": {"namespace_id": "35","initialize_with_readme": "false", "name": "repo name"},
+                         "url": "https://gitlab.thebillingegroup.com/api/v4/projects/"},"prum_repo": {}}}], [1]),
+        # repos is not defined, return 0
+        ([{"repos": {}}], [0]),
+        # talk_repo is not defined, return 0
+        ([{"repos": {"talk_repo": {}, "prum_repo": {}}}], [0]),
+        # parameters are not defined
+        ([{"repos": {"talk_repo": {"params": {}, "url": "https://gitlab.thebillingegroup.com/api/v4/projects/"},"prum_repo": {}}}], [0]),
+        # url is not defined
```
sbillinge, line 2189: Another reason to make code readable and get rid of comments is when they are incorrect, which adds confusion doesn't remove it.  Actually, the url *is* defined, it is defined as an empty string.  What do you want your function to check?  One thing it could check is if the 'url' field is missing altogether. Another thing, which is stronger, is that it is not a properly defined URL (has incorrect characters in it for example).  Another thing it can check is if the URL actually exists.  Which is your goal?  The last is the most useful (and the hardest to test....would have to use Mock).

tests/test_tools.py
```diff
@@ -2174,5 +2175,69 @@ def test_get_tags_invalid():
         get_tags(coll)
         assert e_info == 'ERROR: valid tags are comma or space separated strings of tag names'
 
+@pytest.mark.parametrize(
+    "inputs, expected", [
+        # everything is defined, return 1
```
sbillinge, line 2180: I wonder if we should change what is returned.  exit(0) usually means the function executed successfully.   It makes confusion here.  Maybe to get the intent clear we could call the function `repo_information_complete()` and have it return True or False, then when it is used in the code it will look like 
```
  if repo_information_complete():
     make my repo code
```

sbillinge (commented):
great progress.  Please see comments.

tests/test_tools.py
kaylabim: inputs changed to repo_information and no longer a list. 
repo1 also contains url, and params is passed as a dict in the request!

regolith/tools.py
```diff
@@ -2102,6 +2105,176 @@ def google_cal_auth_flow():
...
+    elif not token_valid:
+        print("WARNING: The authentication token may not be defined. If you would like regolith to "
+              "automatically create a repository in GitHub/GitLab, please add your private authentication "
+              "token in user.json. See regolith documentation for details.")
+    else:
+        print("If you would like regolith to automatically create a repository in GitHub/GitLab, "
+              "please add your repository information and private authentication token in reolgithrc.json"
+              "and user.json respectively. See regolith documentation for details.")
+
+def create_repo_old(name, repo_info, rc):
```
sbillinge, line 2223: delete this function so we can merge the PR.  save the code and paste it into a new branch where you will work on the url creation code.

regolith/tools.py
```diff
@@ -2102,6 +2105,176 @@ def google_cal_auth_flow():
...
+            the key in rc for target repo information (e.g. 'repo1')
+        token_info - string
+            the key in rc for token (e.g. 'priv_token')
+        rc - run control object
+
+    Returns:
+        Success message (repo target_repo has been created in talks) if repo is successfully created in target_repo
+        Warning/setup messages if unsuccessful (or if repo info or token are not valid)
+    """
+    repo_info_valid = repo_info_complete(target_repo, rc)
```
sbillinge, line 2195: this is not true.  You have not validated that the repo is valid.  Only that something is entered for the bits and pieces.

sbillinge (commented):
this is not quite what I had in mind, I was thinking to put the warning and error messages into the validation function, and the create repo, just creates the repo (so all those elifs would be at the appropriate places in teh validation function.)  Is there a reason you put them here?

regolith/tools.py
```diff
@@ -2102,6 +2105,170 @@ def google_cal_auth_flow():
...
+                                return False
+                        else:
+                            print(
+                                "WARNING: The request url may not be defined. "
+                                "If you would like regolith to automatically create a repository in GitHub/GitLab, "
+                                "please add repository information in regolithrc.json. See regolith documentation "
+                                "for details.")
+                            return False
+                    else:
+                        print(
```
sbillinge, line 2184: whenever cut and paste of code is happening it is time to define a variable
```
message_url_not_defined = ""WARNING: The request url may not be defined. "
                            "If you would like regolith to automatically create a repository in GitHub/GitLab, "
                            "please add repository information in regolithrc.json. See regolith documentation "
                            "for details.")
```
and so on.

regolith/tools.py
```diff
@@ -2102,6 +2105,170 @@ def google_cal_auth_flow():
         token.write(creds.to_json())
     # Save the credentials for the next run
 
+def repo_info_complete(target_repo, rc):
+    """checks if repo information is defined and valid in rc.repos (i.e. that
```
sbillinge, line 2109: this is too long.  First line should be less than 70 characters.  then a blank line, then fill in the details.  Check PEPE8 and/or numpy docstrings for details.

sbillinge (commented):
this is very good now.  Just a couple of tiny polishes (see inline) and also fix tests (it is not passing CI). Please make sure it is passing tests locally before pushing as we only have limited CI credits so don't use CI to debug....

sbillinge:
@kaylabim I am ready to merge this.  Please can you resolve the conflict and remove the draft status?  Tx

regolith/tools.py
```diff
@@ -2110,6 +2122,143 @@ def google_cal_auth_flow():
...
+    if repo_info_exists and token_exists:
+        try:
+            response = requests.post(rc.repos[target_repo]['url'], params=rc.repos[target_repo]['params'],
+                                     headers={'PRIVATE-TOKEN': '{}'.format(token_exists)})
+            response.raise_for_status()
+            return f"repo {name} has been created in talks"
+        except requests.exceptions.HTTPError:
+            raise HTTPError(f"WARNING: Unsuccessful attempt at making a GitHub/GitLab etc., repository "
+                            f"due to an issue with the API call (status code: {response.status_code}). "
+                            f"Check that your private token and repository information are valid in "
```
sbillinge, line 2254: this is confusing because private token should NEVER be in regolithrc.json but the wording here makes it sound to a user like it might be.....please update the wording so this is handled explicitly.  Best may be to link to a place where the documentation is rather than making a really long message here.

sbillinge (commented):
please see inline comment.  I would like to merge this if possible so the functions become available.

regolith/helpers/a_presentationhelper.py
```diff
@@ -221,4 +227,7 @@ def db_updater(self):
             rc.client.insert_one(rc.database, EXPENSES_COLL, edoc)
             print(f"{key} has been added in {EXPENSES_COLL}")
 
+        rc.repos['talk_repo']['params']['name'] = key
+        if not rc.no_repo:
```
sbillinge, line 231: nicely done!  3 lines of code!

tests/test_tools.py
```diff
@@ -2174,5 +2179,105 @@ def test_get_tags_invalid():
...
 
+@pytest.mark.parametrize(
+    "repo_information, expected", [
+        ({"repos": {
+            "repo1": {
+                "params": {
+                    "namespace_id": "35",
+                    "initialize_with_readme": "false",
+                    "name": "repo name"},
+                "url": "https://gitlab.thebillingegroup.com/api/v4/projects/"}}
```
sbillinge, line 2190: Let's use something less specific.  We don't really want to give away this much info in the open repo. Since you are mocking on the test it can be any valid url.  Same goes below.

tests/test_tools.py
```diff
@@ -2174,5 +2179,105 @@ def test_get_tags_invalid():
...
+            "repo1": {
+                "params": {
+                    "namespace_id": "35",
+                    "initialize_with_readme": "false",
+                    "name": "repo name"},
+                "url": "https://example.com"}}
+         }, False)
+    ]
+)
+def test_repo_info_complete(repo_information, expected):
```
sbillinge, line 2239: is this testing all the functionality?  for example that it is raising the right exception and giving the right error message in different situations?

sbillinge (commented):
please see inline comments.  Nearly there.

tests/test_helpers.py
```diff
@@ -63,8 +63,8 @@
       "--title", "On the roundness of the Earth", "--status", "in-prep",
       "--notes", "this is a sample added presentation",
       "--presentation-url", "http://drive.google.com/SEV356DV",
-      "--no-cal"],
-     "2006as_mars has been added in presentations\n2006as_mars has been added in expenses\n"),
+      "--no_cal"],
```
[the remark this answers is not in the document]
kaylabim: changed!

tests/test_tools.py
kaylabim: Yes, I just reviewed all the test cases and cross checked them with everything I wanted to test for in the functionality. I made some tweaks to the code in the process while refining the tests as well

tests/test_tools.py
kaylabim: Done! changed to example url

news/closes_908.rst
```diff
@@ -0,0 +1,16 @@
+**Added:**
+ * functionality to tools that checks that required repository request information (e.g. request parameters and url) are defined in rc 
+ * functionality to tools that checks that private authentication tokens for API requests are defined in rc 
+ * functionality to tools that creates a repository based on given repo info
```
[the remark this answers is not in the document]
kaylabim: For the first point for example, would something like "functionality to tools that returns True if all required repository request information (e.g. request parameters and url) is defined and valid in rc" make more sense to the user/developer?

PR #929, opened 2022-06-15
WIP: Regolith Fix #880 v4

### LINKED ISSUE #880: expense added without null notes in a_expense and a_presentation

dpeters001:
Closes #880 and assigns default behavior for the a_presentation helper --notes option. Default behavior is an empty list. If notes are added, the rc.notes will update the document with a list of strings.  
All tests pass for test_helpers.py as well.


dpeters001:
Thanks for pointing this out. I did not think of this. I will remove the if statement.

sbillinge:
bravissimo, the code just got BETTER!

PR #917, opened 2022-06-12
WIP: Closes Issue #910 (ENH: allow different databases for updating expenses and presentations in a_presentation)

N-G-Asker:
closes #910

PR #936, opened 2022-06-20
WIP: closes #910, enhancement for --expense-db flag in presentation adder

### LINKED ISSUE #910: ENH: allow different databases for updating expenses and presentations in a_presentation

N-G-Asker:
Added --expense-db flag to a_presentation adder. If specified, it adds the expense item generated for the presentation to the specified database instead of the default one (first database in the regolithrc.json file)

closes #910.

### DIFF
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -103,6 +103,9 @@ def subparser(subpi):
     subpi.add_argument("--no_cal",
                        help=f"Do not add the presentation to google calendar",
                        action="store_true")
+    subpi.add_argument("--force",
+                       help=f"force adding presentation expense data to a public database, by default the first entry under 'databases` in regolithrc.json. DANGER: This could reveal sensitive information to the public.",
+                       action="store_true")
     return subpi
 
 
@@ -123,8 +126,59 @@ class PresentationAdderHelper(DbHelperBase):
         rc.coll = f"{TARGET_COLL}"
         if not rc.database:
             rc.database = rc.databases[0]["name"]
+
+        # if an expense database is specified using the --expense-db option,
+        if rc.expense_db:
+
+            # check that it's known / find it in rc.databases. Store its index number so we can access its other keys inside rc.databases
+            index = 0
+            db_known = False
+            while index < len(rc.databases):
+                if rc.databases[index]["name"] == rc.expense_db:
+                    db_known = True
+                    break
+                index += 1
+            expense_db_index = index
+
+            # If the database is not known, exit with descriptive error message
+            if not db_known:
+                 raise RuntimeError(
+                    "WARNING: The expense database provided does not exist. "
+                    "Please rerun specifying a known private database, "
+                )
+
+            # if the specified expense database is public and the --force option wasn't passed, exit with a warning message
+            if (rc.databases[expense_db_index]["public"] == True) and (not rc.force):
+                raise RuntimeError(
+                    "WARNING: The expense database provided is not private. Please rerun specifying a known private database, "
+                    "or, at your own risk, use the --force option to add the presentation expense data to a public database"
+                )
+        
+        # if no expense database is specified, set it as the first private database listed in rc. 
+        # If no private database is found/known, exit with a warning message. If, however, the 
+        # --force option was passed, the expense database is set to be the first entry under databases
+        # in regolithrc.json file, even if that database is public. 
         if not rc.expense_db:
-            rc.expense_db = rc.databases[0]["name"]
+            if rc.force:
+                rc.expense_db = rc.databases[0]["name"] # defaults to first entry under databases in regolithrc.json file
+                if rc.databases[0]["public"]
+                    print(f"{key} has been added in {EXPENSES_COLL} in database {rc.expense_db}")
+            else:
+                for db in rc.databases:
+                    # if the db is private, set it and break out of the for loop
+                    if db["public"] != True:
+                        rc.expense_db = db["name"]
+                        break
+                    # else, continues to next database in the dictionary
+
+                # If we still have no value for the rc.expense_db, it means we didn't find a private database after exhaustively iterating through the known databases
+                # So we fail with warning "no default private db to enter expense data into"
+                if not rc.expense_db:
+                    raise RuntimeError(
+                    "WARNING: there is no known private db to enter expense data into. Please rerun after adding a private database to your regolithrc.json file"
+                    "or, at your own risk, use the --force option to add the presentation expense data to a public database"
+                )
+
         gtx[rc.coll] = sorted(
             all_docs_from_collection(rc.client, rc.coll), key=_id_key
         )
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -144,7 +144,7 @@ class PresentationAdderHelper(DbHelperBase):
             if not db_known:
                  raise RuntimeError(
                     "WARNING: The expense database provided does not exist. "
-                    "Please rerun specifying a known private database, "
+                    "Please rerun specifying a known private database."
                 )
 
             # if the specified expense database is public and the --force option wasn't passed, exit with a warning message
@@ -161,7 +161,7 @@ class PresentationAdderHelper(DbHelperBase):
         if not rc.expense_db:
             if rc.force:
                 rc.expense_db = rc.databases[0]["name"] # defaults to first entry under databases in regolithrc.json file
-                if rc.databases[0]["public"]
+                if rc.databases[0]["public"]:
                     print(f"{key} has been added in {EXPENSES_COLL} in database {rc.expense_db}")
             else:
                 for db in rc.databases:
--- a/regolith/helpers/a_presentationhelper.py
+++ b/regolith/helpers/a_presentationhelper.py
@@ -1,6 +1,7 @@
 """Helper for adding a presentation to the presentation collection.
 """
 import time
+from warnings import warn
 
 import dateutil.parser as date_parser
 import threading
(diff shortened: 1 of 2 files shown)


N-G-Asker:
This commit passes all tests locally (automatic and manual):
1. `python run_tests.py tests/test_helpers.py` results in 93 tests passed with no errors
2. `python run_tests.py` results in all tests passed, excepting two errors that it shares with synchronized master and for which I just opened an issue (#937).
3. My manual testing locally succeeded. Specifically, locally I modified the regolithrc.json file in rg-db-public/local so that regolith regolith recognizes the rg-db-group database (see our other thread for the dirty [details](https://github.com/regro/regolith/pull/917#issuecomment-1155623868)), and then, from within rg-db-public, I used the CLI to add a presentation that only updated the expense database in rg-db-group, exactly as intended. I did this by running the presentation adder and specifying the `--expense-db` option. Again, please see the other [thread](https://github.com/regro/regolith/pull/917#issuecomment-1154540883) for the specific command syntax.

sbillinge:
Don't update the `regolithrc.json` in `rg-db-public` (or if you did for testing that is ok, but don't push it).  The code should work successfully without the group database specified in there.   Doing it this way will change the behavior of everything in rg-db-public (all the collections loaded will be different).

I see that it does mean we will have to work harder to make the code change work (doesn't it always end up that way?  I thought it was simple but it never is).  Ask Kayla about test driven development, but here are some use cases (UCs) that we will need to handle:

1. no databases specified on the command line.  Code just puts everything in the default db
2. just --database (should we change it to presentation_database?) specified.  This db is present in the rc.json.  Code uses this db for the presentation, uses the default one for expenses (assume selected)
3. just `--expenses-db` specified and the db present in rc.json. as above but the other way around.
4. both --database and --expenses-db specified. db present in rc.json.  Code does as asked.
5. any or both databases specified, but any of the dbs not present in rc.json.  Fail with helpful message.  Message should tell user to run the command in a directory with a regolithrc.json that contains the db info.

In this case, you would have to run the command from rg-db-group and specify both databases to get the desired behavior, the default behavior won't work.    There is a flag in the rc.json about whether a db is public or not.  It is unlikely that someone wants to put expense information in a public repo and so this is likely a mistake.  Issuing a warning, or even better, having the code fail with a good message (though the behavior should be overridable with something like a --force), may be best to avoid unwanted errors.

N-G-Asker:
> Don't update the `regolithrc.json` in `rg-db-public` (or if you did for testing that is ok, but don't push it). **The code should work successfully without the group database specified in there**. Doing it this way will change the behavior of everything in rg-db-public (all the collections loaded will be different).

In the bolded portion, by "successfully" do you mean failing with a helpful error message (consistent with UC 5)? If `regolithrc.json` in `rg-db-public` lacks an entry for `rg-db-group`, the program won't be able to find it to drop the expense data in the expense collection.

N-G-Asker:
> **Don't update the `regolithrc.json` in `rg-db-public` (or if you did for testing that is ok, but don't push it)**. The code should work successfully without the group database specified in there. Doing it this way will change the behavior of everything in rg-db-public (all the collections loaded will be different).

I only did this for manual local testing and did not and will not push the changes.

N-G-Asker:
> I see that it does mean we will have to work harder to make the code change work (doesn't it always end up that way? I thought it was simple but it never is). Ask Kayla about test driven development, but here are some use cases (UCs) that we will need to handle:

Kayla just gave us a great overview of Test-Driven Development. I'll rework this enhancement following its goals and using the use cases you laid out.

regolith/helpers/a_presentationhelper.py
```diff
@@ -119,6 +126,61 @@ def construct_global_ctx(self):
...
+                    db_known = True
+                    break
+                index += 1
+            expense_db_index = index
+
+            # If the database is not known, exit with descriptive error message
+            if not db_known:
+                 raise RuntimeError(
+                    f"WARNING: The expense database provided, {rc.expense_db}, does not exist. "
+                    "Please rerun specifying a known private database. "
```
sbillinge, line 147: similarly, I would combine these I think. If am a user, I may not know what "specifying a known private database means".  This  sounds a bit like one of those frustrating error messages written by a developer that doesn't really help me as a user to know what I am supposed to do to fix the situation.  Come up with your own (or use this) but I think if I was a user a message something like `f"ERROR: You specified database {rc.expense_db} for the expenses collection when you ran this helper, but I don't find that database specified in your regolithrc.json file.  Please check the spelling when you ran the program and that the database is correctly specified in regolithrc.json."`

regolith/helpers/a_presentationhelper.py
```diff
@@ -119,6 +126,61 @@ def construct_global_ctx(self):
...
+                    "(Consider checking your spelling against the known databases listed in regolithrc.json.)"
+                )
+
+            # if the specified expense database is public and the --force option wasn't passed, exit with a warning message
```
sbillinge, line 151: delete blank line and comment (evident from the code)

regolith/helpers/a_presentationhelper.py
```diff
@@ -119,6 +126,61 @@ def construct_global_ctx(self):
...
+            if (rc.databases[expense_db_index]["public"] == True) and (not rc.force):
+                raise RuntimeError(
+                    "WARNING: The expense database provided is not private. Please rerun specifying a known private database, "
```
sbillinge, line 154: this is a good error message.  Here I am not altogether sure whether it should be WARNING or ERROR.  Strictly, since we are raising an exception it should be ERROR.  Another approach is to not raise an exception here but call `sys.exit` and leave it as a warning.   I think this may be more friendly to the user because it will look less like a crash.

sbillinge (commented):
This is looking great!  good progress. I left some comments.   Before I leave more comments I had some thoughts.  I think it is probably fine as is, if it is passing tests (where they are possible, i.e., the lines of code that don't update any collection.  Nice work on that!).  However, I have suggestions how to make i tmore readable.

General comments are these:

1. don't extensively comment the code.  Just write it in a way that when I read the code I know what you are doing.  Only add short clarifying comments where hings are slightly ambiguous or not clear just from reading the code.  If these comments are structuring for yourself, that is ok but delete them before merging.
2. avoid blank lines except where needed.  Please refer to PEP8 and follow their guidelines.  Blank lines convey information about the structure of the code, which becomes very diffucult to interpret when they are scattered through everywhere
3.  try and avoid `break` which makes for more "spaghetti"-like  code where the flow is hard for a human to follow, even if a program can.
4.  I wonder if we can make the code more concise.  Can we do the work of iterating through the rc.databases just once at the beginning and retrieving the relevant database.   Then we can just run a series of checks on the found database?  something like
```
if rc.expense_db:
    expense_db =  [db for db in rc.databses if db["name"] == rc.expense_db]
else:
    expense_db = rc.databases[0]

if not expense_db:   # check but I think an empty list resolves as False.  To be more explicit we could if len(expense_db) == 0
   "can't find the db" error logic

if expense_db['public']:
    "db is public logic"

etc....
```

N-G-Asker:
Thank you for the detailed feedback. I think I made big progress on improving the code. Now I am writing and running tests locally and adapting the code accordingly.

sbillinge (commented):
Since we can test all the error messages because they bump out before updating the database, we can use TDD here.   I am getting a bit confused about all the things so TDD is definitely helpfull here as we can devise and test all the possibilities.  I would suggest to make a matrix of all the possibilities and what we think the user might want to happen in that circumstance, then write a test for each one.  You may have some of them covered but this will be more systematic.   for example,
```
 - cli.no-expense.   no cli.expenses_db, User doesn't want expense created, run cleanly, don't make an expense (can't test this one)
 -  cli.no-expense.   cli.expenses_db, User probably wants an expense created but didn't formulate the command completely.  Program runs cleanly and but produces a warning asking user if this was their intention?  Or raises an error and asks to reformulate correctly?  Or do we even need both these switches.  We could just use the cli.expense_db as the only switch that triggers the build of an expense?  What do you think?
 - no cli.no-expense cli.expenses_db database found in rc.json.  User wants expense, no problem doing it, program runs cleanly and works (test for this one)
 - ...and so on
```
Once this table is agreed upon, write the tests that test for this.  Put the error messages in the test.  I strongly suggest to delete all the code you have written in the a_presentation.py and then start over with a clean page to design the best way to make all the tests pass and still have readable code.   Here putting comments in the tests will be helpful with the intent of each test, which is a bit hard to get from reading the code itself.  so this is a good time to break my "don't comment code" rule, but the comments will look like:
`# User wants an expense created but the database is missing from regolithrc.json` or whatever the situation is that you are testing.

sbillinge:
one more thing.  It is convention that command line switches are written with dashes (which resolve to underscores in the variable).  For consistency please can you make that change for everything.  example `--no_cal` -> `--no-cal`.  No other code needs to be edited in this case because `--no-cal` results in `rc.no_cal == True` internally.

N-G-Asker:
> one more thing. It is convention that command line switches are written with dashes (which resolve to underscores in the variable). For consistency please can you make that change for everything. example `--no_cal` -> `--no-cal`. No other code needs to be edited in this case because `--no-cal` results in `rc.no_cal == True` internally.

Will do -- please see issue #942 opened for tracking.

N-G-Asker:
> Since we can test all the error messages because they bump out before updating the database, we can use TDD here. I am getting a bit confused about all the things so TDD is definitely helpfull here as we can devise and test all the possibilities. I would suggest to make a matrix of all the possibilities and what we think the user might want to happen in that circumstance, then write a test for each one. You may have some of them covered but this will be more systematic. for example,

Sounds good. I will create a matrix of "user stories" including the examples you listed so we can iterate on them here. Once that's established, for the code I'll start on a clean slate, on a fresh branch off the synchronized master.

N-G-Asker:
| `no-expense`  | `expense-db` EXPENSE_DB | User Intention | Desired Program Behavior |
| ------------------ | --------------------------------------- | ------------------ | ------------------------------------ |
| &check; | &cross; | User doesn't want expense created | run cleanly, don't make an expense (can't test this one) |
| &check;| &check; | User probably wants an expense created but didn't formulate the command completely. | Program runs cleanly and but produces a warning asking user if this was their intention?  Or raises an error and asks to reformulate correctly?  Or do we even need both these switches.  We could just use the cli.expense_db as the only switch that triggers the build of an expense?  What do you think? |
| &cross;  | &cross; | User wants expense | program runs cleanly and works (test for this one) |

N-G-Asker:
> Or do we even need both these switches.  We could just use the cli.expense_db as the only switch that triggers the build of an expense?  What do you think?

I think this comes down to your preference, as the primary end-user. If we eliminated the `no-expense` flag, by default no expense would be created unless `--expense-db EXPENSE_DB` is input as part of the helper command. If you forgot to specify the `--expense-db` option when logging a presentation, you might lose track of an expense you needed to account for.

sbillinge:
> > Or do we even need both these switches.  We could just use the cli.expense_db as the only switch that triggers the build of an expense?  What do you think?
> 
> I think this comes down to your preference, as the primary end-user. If we eliminated the `no-expense` flag, by default no expense would be created unless `--expense-db EXPENSE_DB` is input as part of the helper command. If you forgot to specify the `--expense-db` option when logging a presentation, you might lose track of an expense you needed to account for.

Yah, on balance I agree.  I would probably want the thing to create an expense automatically if it can and just let me know if it can't, but with an option to override this.   How about this for an idea (may be too gimmicky)...if no `expense-db` (and no `no-expense`) is specified, iterate through the list of databases and use the first non-public one found.  On balance, I think this is getting "too clever" and starting to look like magic.  Its always bad to have magic in code (code is behaving strangely and the user has no idea why....).    On balance, let's stick with 

1. use expense-db
2. if no expense-db, try and use the first database in the rc.json.  
   1. if it is public don't use it, error with clear message explaining why (...rerun with --force or specify expense-db)
   2. if not public, use it but put WARNING message that we couldn't find a expense-db and so used <databasename>.  Rerun with expense-db if you want different behavior.

Table looks good.  The table plus the discussion here might complete all the possibilities?

sbillinge:
I might suggest to copy-paste the code you have for handling these things into a scratch file and making a new branch and doing the changes on a clean branch.  Reference to this PR so we can find all the discussion (also maybe copy over the table and the final discussion).

Then go ahead and write the tests.  Take a lot of care and thought so that they capture the behavior described above.   Not all the behavior can be tested.  We can only have one expense.yaml (and it should be the one that tests most lines of code, so the one that does update the expense collection.)  We can't easily test the default behavior of finding a non-public default database (and it probably isn't worth the work to figure out how to do that) but we can test all the cases that error and make sure they error correctly and with the right error message.

This is just an onboarding prum issue, but it is a good learning experience so worth doing as well as possible, but at some point we want to capture the behavior and get it merged and move on.

PR #945, opened 2022-06-27
WIP: ENH #910 `a_presentation` helper `--expense-db` option

### LINKED ISSUE #910: ENH: allow different databases for updating expenses and presentations in a_presentation

N-G-Asker:
Added `--expense-db` option, developed using test-driven development / user stories to think about the different cases of end-user intention and program behavior

closes #910.

See also PR #936  for historical development planning / discussion.


sbillinge (approved):
This is really great!  Nearly there.

- [ ] please remove _all_ the comments from the code (But not from the tests!)
- [ ] uncomment tests in such a way that we test the most lines of code that we can.  This means all the tests that don't result in a database being updated can be uncommented.  Of the other tests, uncomment the one that does the most work.   This is almost certainly C.2.
- [ ] add the error message to the test for tests that result in a fail, if at all possible (i.e., test that the right error message is generated as well as the right exception is raised.
- [ ] The code that implements everything is a bit hard to follow.  Can we think about something like:
```
if rc.no_expense:
    if rc.expense_db:
         fail gracefully
else:
    rc.expense_db = rc.database[0] if not rc.expense_db and not rc.database[0].get("public")
    rc.expense_db = rc.database[0] if not rc.expense_db and rc.database[0].get("public") and rc.force
    #code to handle edge cases here:
    if rc.expense_db not in rc.databases:
           raise RunTimeError("right message here")
    if not rc.expense_db and not rc.database[0].get("public")
            print(INFO: message...put the expense in such and such db by default
```
and so on.  This way I see exactly from the code what the intent of the programmer is and I don't have to interate through lots of deeply nested if-statements full of edge-casey stuff to find the rc.expense_db assignment.   To be completely clear our intent we could also add
`rc.expense_db = rc.expense_db if rc.expense_db` as the first option.  Then it is really explicit.

sbillinge:
@N-G-Asker please can we put in a shift on this so we can merge it?  THnaks so much.

N-G-Asker:
> * The code that implements everything is a bit hard to follow.  Can we think about something like

It is quite tricky to make this cleaner/more readable. I adopted your recommended approach above but could not avoid making some necessary adjustments to the logic that make it a bit harder to read.

Late tonight/tomorrow morning, I will be working the testing to-do items you listed. For your awareness, I am skeptical I'll be able to overcome the major roadblocks I've been facing on the test front -- I alluded to them on my last substantive [commit](https://github.com/regro/regolith/pull/945/commits/c8acb71cc18101d22516cac30f7a4027d509f222):
- Testing for deliberately raised errors is hairy in the existing testing architecture. I haven't been able to figure this out -- do you have any suggestions?
- Testing for multiple additional entries in the same collection using the same adder is not possible in the existing test architecture. I want to confirm with you -- is this true?

sbillinge:
>     * Testing for multiple additional entries in the same collection using the same adder is not possible in the existing test architecture. I want to confirm with you -- is this true?

Yes, as we know this is true.  Please read my comments carefully (slow and steady wins the race) and you will see that I am not asking for that.

sbillinge:
> 
>     * Testing for deliberately raised errors is hairy in the existing testing architecture. I haven't been able to figure this out -- do you have any suggestions?
> 

we generally test fails in a different function.  something like `test_helper_fail()` or sthg like that, and you capture `stderr`.  I am less concerned about this than testing the main functionality, but it is a habit that we should get into if we want to use try/except in our logic flow, which is really acting like an if/else which we do test when we are able to.  Leave this to the end and I will be ok merging without it, but it is not a great practice and you can model this for later coders.

sbillinge:
> > * The code that implements everything is a bit hard to follow.  Can we think about something like
> 
> It is quite tricky to make this cleaner/more readable. I adopted your recommended approach above but could not avoid making some necessary adjustments to the logic that make it a bit harder to read.

OK, np.  I look forward to seeing how it went.

sbillinge:
nearly there.  because you changed the presentation that was added from the flat mars example you will have to change the url that is mocked in the test for creating a presentation database in gitlab in the hmr tests in test_helpers.

sbillinge:
great job, thanks @N-G-Asker !

sbillinge:
btw, I merged this, but I notice that there is no news item.  Please can you make a new PR with the news of what functionality was added for the changelog?  When that is merged we can close the onboarding prum (or whatever the prum is that had this in it).  Thanks.

PR #948, opened 2022-06-30
not_in_cv should be boolean not bool

sbillinge:


sbillinge:
moving these edits to pr with all the changes #949

PR #962, opened 2022-07-25
notes for mongodb projecta

dpeters001:
- adds a notes subdirectory to the regolith repository
- adds a markdown file with detailed notes about the cod mods used for testing in the dp_regolith_prums_to_mongodb projectum


sbillinge (commented):
please change `notes` to `developer_notes` and put it under `docs`, then I can merge.

I did intend it to be with the other mongo stuff, which would mean that you would build the branch off a clean pymongo branch not a clean master, but it is ok to merge it here.  Why not.

dpeters001:
Made the changes.
Also, thanks for explaining. I see my mistake. I could have pulled from pymongo branch and built my new branch from there.

sbillinge:
great, nice work!  Thanks, and yes, you nailed it with your comment above.  

We will have to deal with those merge issues when we merge pymongo into main, but they should be dealt with at that time, not now.

PR #983, opened 2022-11-30
WIP : add lead + remove projectum_id from milestone-updater required args

dpeters001:
1. lead id has been added as a required argument for the milestone-updater.
2. before I change projectum_id from a required arg to an optional arg, we need a query function to iterate through all projecta ids associated with the lead. the query will return a dictionary or list of key value pairs that combines all projecta documents associated with the lead in the correct syntax (deals with line 120-121)
3. once a dictionary or list of key value pairs is assembled, lines 133-149 copy, identify and list the milestones

- regarding list item 2, this part of the code uses the `find_one` function which returns the first doc that has the projectum_id listed. this is fine when the projectum_id is the key, however, if the functions searches for the lead there will be multiple docs in the projecta collection associated with a lead id and it will only return the first key value match that it finds (only one doc will be returned). I spent some time trying to create code that identified all the projecta associated with a lead and that combined these projecta into a single dictionary. I have been unsuccessful so far. I think the solution is to create a new function like `find_one` but call it `find_many` that returns a dictionary of N zipped projecta dictionaries associated with the lead. this way, line 133-149 will not need to be modified.

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -26,6 +26,9 @@ def subparser(subpi):
     if isinstance(subpi, GooeyParser):
         date_kwargs['widget'] = 'DateChooser'
 
+    subpi.add_argument("lead",
+                       help="Specify the project lead of the prum "
+                            "by id, e.g., sbillinge")
     subpi.add_argument("projectum_id", help="The id of the projectum.  If you just "
                                             "specify this the program will return a "
                                             "numbered list of milestones to select for "


sbillinge:
I think our new design is that we will use l_milestones to list the milestones.  l_milesteones will show the uuid for each milestone.  then the job of u_milestones is much easier.  It just has to get the milestone with the given uuid, and have the logic that allows it to be updated in whatever way then updates the db.  So we don't really need any of the logic you are discussing here, no?

sbillinge:
Hi Dom,

The best thing here is to close this PR and delete the branch and make a new  branch on a fresh main.  We just to capture the conversation somewhere, presumably into a new issue.

dpeters001:
Ok. Sounds good. I can add this conversation to the use cases for the prum, or I can create a new issue on the regro/regolith branch. Which do you prefer?

sbillinge:
I am not sure there is much we need on here actually.  It might be best if we take all the things we have to implement and make them as issues and make a release milestone and try and close all the issues.  I can make a start on that.

dpeters001:
Ok. That works. Closing PR.

dpeters001:
I thought about this workflow a little bit more, and it will probably work well for updating milestones that have already been created. However, creating a new milestone using the milestone-updater will still require the user to enter the projectum_id before adding the milestone to the projecta and todo collections. It would be ideal to do away with the projectum_id requirement for creating new milestones, but I think it is best to keep it for now.

PR #996, opened 2022-12-28
WIP: adds two new tests for the milestone-lister 

### LINKED ISSUE #985: Have milestone lister show the milestone uuid
Have the lister list the milestone uuid. 

In non-verbose mode have it look like:
```
2022-01-06 (aedes1) lead: dpeters, dp_take_projecta_to_mongo_in_production, create the rg-db-group equivalent db on the mongo atlas 
```
where `aedes1` are the first 6 symbols of the milestone uuid.

In verbose mode have it look like:
```
2022-01-06 (aedes1): lead: dpeters, dp_take_projecta_to_mongo_in_production, status: started
    Type: other
    Title: create the rg-db-group equivalent db on the mongo atlas cluster
    log url: 
    Purpose: create the official mongo backend db for the projecta collection
    Audience: dpeters, sbillinge, nasker
    Notes:
      - () (Simon) create a db on the billinge group atlas cluster that will contain the projecta collection
      - () (Simon) assign read and write privileges on new db to dpeters
      - () upload the entire projecta.yml file to the new atlas db using fs-to-mongo rego command
     uuid: aedes1dsrtesgdw4cdatesd
```
where `aedes1dsrtesgdw4cdatesd` would be the full uuid, whatever it is.

dpeters001:
Closes #985 

couple questions:
- I have added a test for the non-verbose entry. Is this necessary? 
- do I also need to modify the other l_milestone tests? (I believe this is necessary, but I want to make sure)

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -158,6 +158,12 @@ helper_map = [
      "2020-05-27: lead: ascopatz, sb_firstprojectum, status: proposed\n    Type: mergedpr\n    Title: planning meeting\n    log url: https://docs.google.com/document/d/1YC_wtW5Q\n    Purpose: develop a detailed plan with dates\n    Audience: ascopatz, scopatz, ascopatz\n"
      "2020-05-20: lead: ascopatz, sb_firstprojectum, status: proposed\n    Type: meeting\n    Title: Project lead presentation\n    log url: https://docs.google.com/document/d/1YC_wtW5Q\n    Purpose: lead presents background reading and initial project plan\n    Audience: ascopatz, scopatz, ascopatz\n    Notes:\n      - do background reading\n      - understand math\n"
      ),
+    (["helper", "l_milestones", "--lead", "pliu"],
+     "2021-08-03 (milest): lead: pliu, pl_secondprojectum, Milestone, status: converged\n"
+     ),
+    (["helper", "l_milestones", "--lead", "pliu", "--verbose"],
+     "2021-08-03 (milest): lead: pliu, pl_secondprojectum, status: converged\n    Type: \n    Title: Milestone\n    log url: None\n    Purpose: None\n    Audience: \n    uuid: milestone_uuid_pl2\n"
+    ),
     (["helper", "l_projecta", "--verbose", "--orphan"],
      "ab_inactive\n    status: backburner, begin_date: 2020-05-03, due_date: 2021-05-03, grant: dmref15\n    description: a prum that has various inactive states in milestones and overall\n    team:\n        lead: abeing\n        group_members: None\n        collaborators: None\n"
      ),


tests/test_helpers.py
```diff
@@ -158,6 +158,12 @@
      "2020-05-27: lead: ascopatz, sb_firstprojectum, status: proposed\n    Type: mergedpr\n    Title: planning meeting\n    log url: https://docs.google.com/document/d/1YC_wtW5Q\n    Purpose: develop a detailed plan with dates\n    Audience: ascopatz, scopatz, ascopatz\n"
      "2020-05-20: lead: ascopatz, sb_firstprojectum, status: proposed\n    Type: meeting\n    Title: Project lead presentation\n    log url: https://docs.google.com/document/d/1YC_wtW5Q\n    Purpose: lead presents background reading and initial project plan\n    Audience: ascopatz, scopatz, ascopatz\n    Notes:\n      - do background reading\n      - understand math\n"
      ),
+    (["helper", "l_milestones", "--lead", "pliu"],
```
sbillinge, line 161: these look like new tests (since I don't see any lines that are in red that will be deleted).  Since you are updating an existing helper there will almost certainly (should) be existing tests.  So your job is to go and find all the `l_milestones` tests and _update_ them so they exhibit the desired behavior.  You may also want/need to add additional new tests if there is new functionality that needs to be tested, but in this case we are not changing functionality, only what we want printed on the output line, so I guess we won't need any new tests, just updated existing tests.

tests/test_helpers.py
```diff
@@ -158,6 +158,12 @@
...
+     "2021-08-03 (milest): lead: pliu, pl_secondprojectum, Milestone, status: converged\n"
+     ),
+    (["helper", "l_milestones", "--lead", "pliu", "--verbose"],
+     "2021-08-03 (milest): lead: pliu, pl_secondprojectum, status: converged\n    Type: \n    Title: Milestone\n    log url: None\n    Purpose: None\n    Audience: \n    uuid: milestone_uuid_pl2\n"
```
sbillinge, line 165: This is exactly correct as we want it, except `milest` should be the uuid of the actual milestone that is printed by the test.  See if you can figure out where the code gets it from, and what it is (hint: you may want to open `schemas.py` and look around).  Reach out if you can't.

sbillinge (commented):
Thanks @dpeters001 this is great!

First point, you built your branch on a non-synchronized main.  ALWAYS check out main, then do `git pull upstream main`, make sure it was successful, then make the new branch off of that.  I suggest you just delete this branch and start over as the edits are quite modest.  Really get this workflow in your DNA so it more or less happens automatically.

Second, along the same lines, double-check that you have the commit-hook set so that you can't commit to main by mistake.  I expect you do and have no reason to think you haven't, but now is the time to double check.  Look in the group wiki to see how.  This will save a lot of heartache in the future.....

Please see my comments inline.

This is a great job.   You will be up to speed in no time and it will go quickly....

tests/test_helpers.py
dpeters001: Yes. I wrote two new tests to get a better understanding of the testing structure. I will get rid of these and update the existing tests. 

Also, I have a rookie question about my `issue_#985` branch. My main is synchronized as far as I can see, but how can I tell if the branch of my PR is synced to the latest version of the main branch? I am also having an issue with my gitlab login. I will follow up with you on Slack about that.

tests/test_helpers.py
dpeters001: In `schemas.py`, the uuid for all milestones look something like this: `milestone_uuid_pl1`.  That is why I have the `(milest)` bit. If there are exemplar uuids for milestones that look something like `aedes1dsrtesgdw4cdatesd`, I am not seeing those entries. Am I looking at the correct exemplars?

sbillinge:
OK, I think I have been a bit of an idiot, not taking enough time, you are way ahead of the curve already.   I saw the exclamation point in a triangle and thought this was indicating it couldn't be merged because of a conflict with main, but actually it is because you have it listed as a draft, my bad!  So I don't think there is any synchronization issue, just me being an idiot.

On the uuid's I am also being an idiot.  Later in the process I realized that I needed to edit the first six positions to make them unique, but I didn't go back and fix the original ones I made in the schema.  We have two options.  One is to make new uuid's in the schema so they behave better, or just leave it as the tests will work just as well.  Changing the uuids will trigger a bunch of changes that need to be made to outputs so is not that quick of a fix.  let's just leave them for now.

It seems like you are understanding everything rather well.  Therefore just delete these new tests and update the old tests and we can get on with the coding!   Sorry for creating confusion.

dpeters001:
No worries!

PR #1001, opened 2023-01-10
changes u_milestone helper tests

### LINKED ISSUE #987: milestone updater uses uuid to select which milestone to update
nothing else changes as to how it updates.  In each case, write tests first and get the reviewed, then write code.

- [ ] make uuid fragment an optional cl arg
- [ ] make prum _id an optional requirement (change from currently it being required)
- [ ] if no uuid, assume it wants to add a milestone and look for the prum_id
- [ ] if no prum_id assume it will update an existing milestone and follow that flow
- [ ] if both exit gracefully (with info to user about what went wrong and what they should do to put it right)
- [ ] if neither exit gracefully
- [ ] update the updating logic so the tests pass.  It will mostly be messing with existing flow logic (iterating over prums and things).  We are not yet working on making the actual updating logic easier to use.  That comes later.
- [ ] add rever news
- [ ] remove the listing capability (we will list using milestone lister now)
- [ ] make sure all tests pass

dpeters001:
Closes #987 

First attempt at making tests for new milestone updater functionality.

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -206,7 +206,7 @@ class MilestoneUpdaterHelper(DbHelperBase):
                             id.append(pid)
                 if target_ko:
                     for prum in target_ko:
-                        if prum.get('_id')[0:len(uuid)] == uuid[2:]:
+                        if prum.get('_id')[0:len(uuid)-2] == uuid[2:]:
                             kickoff = prum['kickoff']
                             upd_mil.append(kickoff)
                         if upd_mil:
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -32,7 +32,7 @@ def subparser(subpi):
                             "Takes a full or partial uuid. "
                             "Multiple uuids may be entered.",
                        nargs='+')
-    subpi.add_argument("--projectum_id", help="The id of the projectum. If you "
+    subpi.add_argument("-p", "--projectum_id", help="The id of the projectum. If you "
                                             "opt for this the program will assume "
                                             "you are adding a new milestone "
                                             "to the specified projectum.")
@@ -121,7 +121,7 @@ class MilestoneUpdaterHelper(DbHelperBase):
             if not target_prum:
                 pra = fragment_retrieval(self.gtx["projecta"], ["_id"],
                                          rc.projectum_id)
-                print("Projecta not found. Projecta with similar names: ")
+                print("Projectum not found. Projecta with similar names: ")
                 for i in range(len(pra)):
                     print(f"{pra[i].get('_id')}")
                 print("Please rerun the helper specifying the complete ID.\n"
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -27,27 +27,16 @@ def subparser(subpi):
     if isinstance(subpi, GooeyParser):
         date_kwargs['widget'] = 'DateChooser'
 
-    subpi.add_argument("-u", "--milestone_uuid",
+    subpi.add_argument("-i", "--milestone_uuid",
                        help="The uuid of a milestone. "
                             "Takes a full or partial uuid. "
-                            "Multiple uuids may be entered.")
-    subpi.add_argument("--projectum_id", help="The id of the projectum.  If you "
+                            "Multiple uuids may be entered.",
+                       nargs='+')
+    subpi.add_argument("--projectum_id", help="The id of the projectum. If you "
                                             "opt for this the program will assume "
                                             "you are adding a new milestone "
                                             "to the specified projectum.")
-#    subpi.add_argument("-c", "--current", action="store_true",
-#                       help="only list current (unfinished, unpaused) milestones",
-#                       )
-#    subpi.add_argument("-i", "--index",
-#                       help="Index of the item in the enumerated list to update. "
-#                            "To update multiple milestones with the same edits "
-#                            "(often used for finishing checklist items), "
-#                            "please enter in the "
-#                            "format of 2,5,7 or 3-7 for multiple indices.",
-#                       type=str)
-#    subpi.add_argument("-v", "--verbose", action="store_true",
-#                       help="Increases the verbosity of the output.")
-    subpi.add_argument("-d", "--due-date",
+    subpi.add_argument("-u", "--due-date",
                        help="New due date of the milestone. "
                             "Required for a new milestone.",
                        **date_kwargs)
@@ -119,7 +108,6 @@ class MilestoneUpdaterHelper(DbHelperBase):
             now = dt.date.today()
         new_mil = []
         pdoc = {}
-        all_milestones = []
         if rc.projectum_id and rc.milestone_uuid:
             raise RuntimeError("A uuid fragment and projectum id have been entered.\n"
                                 "You may enter a milestone uuid or a projectum id but not both.\n"
@@ -127,32 +115,21 @@ class MilestoneUpdaterHelper(DbHelperBase):
         if not rc.projectum_id and not rc.milestone_uuid:
             raise RuntimeError("No milestone uuid or projectum id was entered.\n"
                                 "Enter a milestone uuid to update an existing milestone, or a projectum id to add a new milestone to that projectum.\n")
-
         if rc.projectum_id:
-            key = rc.projectum_id
-            filterid = {'_id': key}
-            target_prum = rc.client.find_one(rc.database, rc.coll, filterid)
+            rc.projectum_id = rc.projectum_id.strip()
+            target_prum = rc.client.find_one(rc.database, rc.coll, {'_id': rc.projectum_id})
             if not target_prum:
                 pra = fragment_retrieval(self.gtx["projecta"], ["_id"],
                                          rc.projectum_id)
-                if len(pra) == 0:
-                    raise RuntimeError(
-                        "Please input a valid projectum id or a valid fragment of a projectum id")
                 print("Projecta not found. Projecta with similar names: ")
                 for i in range(len(pra)):
                     print(f"{pra[i].get('_id')}")
-                print("Please rerun the helper specifying the complete ID.")
+                print("Please rerun the helper specifying the complete ID.\n"
+                      "If your prum id looks correct, check that this id is in the collection "
+                      "in the database that regolith is looking in (i.e., {rc.database}).")
                 return
             milestones = deepcopy(target_prum.get('milestones'))
-            deliverable = deepcopy(target_prum.get('deliverable'))
-            kickoff = deepcopy(target_prum.get('kickoff'))
-            deliverable['identifier'] = 'deliverable'
-            all_milestones = [deliverable]
-            if kickoff:
-                kickoff['identifier'] = 'kickoff'
-                if not kickoff.get('type'):
-                    kickoff['type'] = 'meeting'
-                all_milestones = [deliverable, kickoff]
+            all_milestones = []
             for item in milestones:
                 item['identifier'] = 'milestones'
             all_milestones.extend(milestones)
@@ -171,16 +148,22 @@ class MilestoneUpdaterHelper(DbHelperBase):
                 mil.update({'audience': rc.audience})
             else:
                 mil.update({'audience': ['lead', 'pi', 'group_members']})
-            if rc.type:
-                mil.update({'type': rc.type})
-            else:
-                mil.update({'type': 'meeting'})
             if rc.status:
                 mil.update({'status': rc.status})
             else:
                 mil.update({'status': 'proposed'})
             if rc.notes:
                 mil.update({'notes': rc.notes})
+            if rc.type:
+                if rc.type in MILESTONE_TYPES:
+                    mil.update({'type': rc.type})
+                else:
+                    raise ValueError(
+                        "ERROR: The type you have specified is not recognized. "
+                        "Please rerun your command adding '--type' "
+                        f"and giving a type from this list: {MILESTONE_TYPES}")
+            else:
+                mil.update({'type': 'meeting'})
             mil.update({'identifier': "milestones"})
             new_mil.append(mil)
             pdoc = {'milestones': mil}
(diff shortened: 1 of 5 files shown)


tests/test_helpers.py
```diff
@@ -474,62 +474,44 @@
       "--date", "2020-04-29", "--collaborators", "afriend", "--description", "more work",
       "--group-members", "ascopatz", "--grants", "SymPy-1.1", "--due-date", "2021-01-01", '--notes', 'new note'],
      "ly_newprojectum has been added in projecta\n"),
-    (["helper", "u_milestone", "sb_firstprojectum", "--index", "4,5",
-      "--status", "converged", "--due-date", "2020-06-01", "--notes", "do this",
-      "do that", "--type", "meeting"],
-     "sb_firstprojectum has been updated in projecta\n"
-     ),
-    (["helper", "u_milestone", "sb"],
+    (["helper", "u_milestone", "--uuid_fragment", "sb_fir", "--status", "finished",
```
sbillinge, line 477: Let's make the variable called `milestone_uuid` so it can be the full uuid, or a fragment of it.

tests/test_helpers.py
```diff
@@ -474,62 +474,44 @@
...
+      "--due-date", "2023-01-01", "--notes", "do this", "do that", "--type", "mergedpr"],
+     "sb_firstprojectum has been updated in projecta.\n"
+    ),
+    (["helper", "u_milestone", "--uuid_fragment", "kosb_f", "--status", "finished",
```
sbillinge, line 481: Which lines of code/behavior is this testing that is different from the test above?  Also, fyi, we can usually only have one test that updates the db file, so we simply try and update as many things as possible to test as many lines of code as possible.  It is not great, but ok.

tests/test_helpers.py
```diff
@@ -474,62 +474,44 @@
...
-     "3. Project lead presentation    due date: 2020-05-20    status: proposed\n"
-     "4. deliverable    due date: 2020-06-01    status: converged\n"
-     "5. planning meeting    due date: 2020-06-01    status: converged\n"
-     ),
+    (["helper", "u_milestone", "--uuid_fragment", "sb_fir", "--projectum_id", "sb_firstprojectum",
+      "--due-date", "2023-01-01", "--notes", "do this", "do that", "--type", "mergedpr"],
+     "A uuid fragment and projectum id have been entered.\n"
+     "You may enter a uuid fragment or a projectum id but not both.\n"
+     "Please enter the uuid fragment of the milestone you would like to update.\n"
+     "If you would like to add a new milestone to a projectum, enter the projectum id.\n"
```
sbillinge, line 507: This is excellent.   I think I would remove the line `"Please enter the uuid fragment of the milestone you would like to update.\n"` which seems to be confusing and change the last line to something like `"enter a milestone uuid to update an existing milestone, or a projectum id to add a new milestone to that projectum\n"`

tests/test_helpers.py
```diff
@@ -474,62 +474,44 @@
...
+     ),
+     (["helper", "u_milestone", "--due-date", "2023-01-01",
+       "--notes", "do this", "do that", "--type", "mergedpr"],
+      "No uuid fragment or projectum id has been entered.\n"
+      "Please enter the uuid fragment of the milestone you would like to update.\n"
```
sbillinge, line 512: please see the comment above and make this help message something comparable so it is clear what the user should do.

sbillinge (commented):
This is great! nice work.   

I wonder if we want an easy way to finish a milestone.  something like
`["helper", "u_milestone", "--uuid_fragment", "kosb_f", "--finish_milestone"]`
We may not be able to test this (we can only test one update) but it is still nice to make the test but comment it out, in case we are able to test it in the future.  It is also a good memory aid, and we can also uncomment it and comment the other one when building the code to do it.  This behavior may already be in the code, I think it is.

One more thing...we don't test abbreviations for the parser, but is `-u` available for `--milestone_uuid`, and `-f` available for `--finish_milestone`?

tests/test_helpers.py
dpeters001: Will do. I like the idea of the user having the option to enter the full uuid, or the uuid fragment.

dpeters001:
> This is great! nice work.
> 
> I wonder if we want an easy way to finish a milestone. something like `["helper", "u_milestone", "--uuid_fragment", "kosb_f", "--finish_milestone"]` We may not be able to test this (we can only test one update) but it is still nice to make the test but comment it out, in case we are able to test it in the future. It is also a good memory aid, and we can also uncomment it and comment the other one when building the code to do it. This behavior may already be in the code, I think it is.
> 
> One more thing...we don't test abbreviations for the parser, but is `-u` available for `--milestone_uuid`, and `-f` available for `--finish_milestone`?

It looks like there already is a `--finish` option using the `-f` abbreviation that finishes a milestone. So the good news is that it's already in the code! I am going to add a test for the `-finish` option in the event that it is possible to run this test in the future. 

The bad news is that `-u` is currently being used for `due-date` so we will need to use a different abbreviation for `milestone_uuid`. How do you feel about changing the abbreviation for `--due-date` to `-d` to allow for `-u` to be used for `milestone_uuid`?

sbillinge:
> > This is great! nice work.
> > I wonder if we want an easy way to finish a milestone. something like `["helper", "u_milestone", "--uuid_fragment", "kosb_f", "--finish_milestone"]` We may not be able to test this (we can only test one update) but it is still nice to make the test but comment it out, in case we are able to test it in the future. It is also a good memory aid, and we can also uncomment it and comment the other one when building the code to do it. This behavior may already be in the code, I think it is.
> > One more thing...we don't test abbreviations for the parser, but is `-u` available for `--milestone_uuid`, and `-f` available for `--finish_milestone`?
> 
> It looks like there already is a `--finish` option using the `-f` abbreviation that finishes a milestone. So the good news is that it's already in the code! I am going to add a test for the `-finish` option in the event that it is possible to run this test in the future.
> 
> The bad news is that `-u` is currently being used for `due-date` so we will need to use a different abbreviation for `milestone_uuid`. How do you feel about changing the abbreviation for `--due-date` to `-d` to allow for `-u` to be used for `milestone_uuid`?

Yes, I am all good with this.

One clarification, my suggestion was that we just have one option for the uuid (`-u`) but the user can enter as much of the uuid as they have (but maybe starting from the beginning) and we compare to `milestone.get(`uuid`)[0:len(rc.uuid)]`, sthg like that.

We may want to have a check that `milestone.get(`uuid`)[0:len(rc.uuid)]` is locally unique and have logic what to do if it is not unique.

sbillinge:
> > This is great! nice work.
> > I wonder if we want an easy way to finish a milestone. something like `["helper", "u_milestone", "--uuid_fragment", "kosb_f", "--finish_milestone"]` We may not be able to test this (we can only test one update) but it is still nice to make the test but comment it out, in case we are able to test it in the future. It is also a good memory aid, and we can also uncomment it and comment the other one when building the code to do it. This behavior may already be in the code, I think it is.
> > One more thing...we don't test abbreviations for the parser, but is `-u` available for `--milestone_uuid`, and `-f` available for `--finish_milestone`?
> 
> It looks like there already is a `--finish` option using the `-f` abbreviation that finishes a milestone. So the good news is that it's already in the code! I am going to add a test for the `-finish` option in the event that it is possible to run this test in the future.
> 
> The bad news is that `-u` is currently being used for `due-date` so we will need to use a different abbreviation for `milestone_uuid`. How do you feel about changing the abbreviation for `--due-date` to `-d` to allow for `-u` to be used for `milestone_uuid`?

:+1:

dpeters001:
> > > This is great! nice work.
> > > I wonder if we want an easy way to finish a milestone. something like `["helper", "u_milestone", "--uuid_fragment", "kosb_f", "--finish_milestone"]` We may not be able to test this (we can only test one update) but it is still nice to make the test but comment it out, in case we are able to test it in the future. It is also a good memory aid, and we can also uncomment it and comment the other one when building the code to do it. This behavior may already be in the code, I think it is.
> > > One more thing...we don't test abbreviations for the parser, but is `-u` available for `--milestone_uuid`, and `-f` available for `--finish_milestone`?
> > 
> > 
> > It looks like there already is a `--finish` option using the `-f` abbreviation that finishes a milestone. So the good news is that it's already in the code! I am going to add a test for the `-finish` option in the event that it is possible to run this test in the future.
> > The bad news is that `-u` is currently being used for `due-date` so we will need to use a different abbreviation for `milestone_uuid`. How do you feel about changing the abbreviation for `--due-date` to `-d` to allow for `-u` to be used for `milestone_uuid`?
> 
> Yes, I am all good with this.
> 
> One clarification, my suggestion was that we just have one option for the uuid (`-u`/`--milestone-uuid`) but the user can enter as much of the uuid as they want (but maybe starting from the beginning) and we compare to `milestone.get('milestone_uuid')[0:len(rc.uuid)]`, sthg like that.
> 
> We may want to have a check that `milestone.get('milestone_uuid')[0:len(rc.uuid)]` is locally unique and have logic what to do if it is not unique.

Awesome idea! Having the flexibility to enter as much of the uuid as the user wants will make this easier to use.

tests/test_helpers.py
sbillinge: The general rule is to make one test that covers the most lines of code.  We have started to make tests for the other functionality but commented out.  Codecov will show which lines of code remain uncovered by the test, but they will be blocks that look like
```
    if rc.something:
        do something
```
and this if is not satisfied in any live test.  Just add as much as you can into one test and we have to hold our fingers for the other things that are not tested.

sbillinge:
@dpeters001 this is good to go when you finalize the best update test.

tests/test_helpers.py
dpeters001: Ok. One more question. 

The `--index` option will no longer be useful when we remove the listing capabilities of `u_milestone`. Do we still want the ability to update multiple milestones at the same time? For example, updating the deliverable and kickoff milestone by entering a list of the milestone uuids? Doing this would also allow us to validate two different parts of the code in one helper test.

dpeters001:
> @dpeters001 this is good to go when you finalize the best update test.

:+1:

tests/test_helpers.py
sbillinge: > Ok. One more question.
> 
> The `--index` option will no longer be useful when we remove the listing capabilities of `u_milestone`. Do we still want the ability to update multiple milestones at the same time? For example, updating the deliverable and kickoff milestone by entering a list of the milestone uuids? Doing this would also allow us to validate two different parts of the code in one helper test.

yes, get rid of --index

for updating multiple ones, I guess why not (especially finishing milestones this is very useful).  So let's have `-u` take a list.

dpeters001: Awesome. Will do.

dpeters001:
Changes:
- projectum_id is not an optional arg that can be specified to add a new milestone (rn it requires the that the complete id is specified)
- milestone_uuid option has been added (rn coded to require full uuid. still working on this. i plan to use the comparison we talked about previous sth like comparing `rc.milestone_uuid` to `milestone.get('milestone_uuid')[0:len(rc.uuid)]`)
- I commented out the `-c` , `-i` and `-v` options as these options were previously being used for listing milestones
- the logic code for adding a new milestone is mostly the same as the code that was there before. I think we will be able to remove the lines of code that get copy deliverables and kickoff 'lines 132-140`
- Right now I am assuming that the full uuid is being given, so I have not added runtime errors for instances in which a partial uuid is given and multiple milestones are identified 
- I still need to create code for the ability of the user to update multiple milestones simultaneously
- I still need to finish coding after `line 258` to update the projecta collection with an updated `milestones`, `kickoff`, or `deliverable` field

regolith/helpers/u_milestonehelper.py
```diff
@@ -27,24 +27,27 @@ def subparser(subpi):
...
-                       help="Index of the item in the enumerated list to update. "
-                            "To update multiple milestones with the same edits "
-                            "(often used for finishing checklist items), "
-                            "please enter in the "
-                            "format of 2,5,7 or 3-7 for multiple indices.",
-                       type=str)
-    subpi.add_argument("-v", "--verbose", action="store_true",
-                       help="Increases the verbosity of the output.")
-    subpi.add_argument("-u", "--due-date",
+    subpi.add_argument("-u", "--milestone_uuid",
```
sbillinge, line 30: let's use `-i` for this.  I use `-u` elsewhere for due date so it is good to keep it uniform.  Also, i for "id" seems to be more intuitive, and also we were using it for index before, so it all points towards that being a better idea.

regolith/helpers/u_milestonehelper.py
```diff
@@ -27,24 +27,27 @@ def subparser(subpi):
...
+                       help="The uuid of a milestone. "
+                            "Takes a full or partial uuid. "
+                            "Multiple uuids may be entered.")
+    subpi.add_argument("--projectum_id", help="The id of the projectum.  If you "
+                                            "opt for this the program will assume "
+                                            "you are adding a new milestone "
+                                            "to the specified projectum.")
+#    subpi.add_argument("-c", "--current", action="store_true",
```
sbillinge, line 38: go ahead and delete all these.

regolith/helpers/u_milestonehelper.py
```diff
@@ -27,24 +27,27 @@ def subparser(subpi):
...
+#    subpi.add_argument("-i", "--index",
+#                       help="Index of the item in the enumerated list to update. "
+#                            "To update multiple milestones with the same edits "
+#                            "(often used for finishing checklist items), "
+#                            "please enter in the "
+#                            "format of 2,5,7 or 3-7 for multiple indices.",
+#                       type=str)
+#    subpi.add_argument("-v", "--verbose", action="store_true",
+#                       help="Increases the verbosity of the output.")
+    subpi.add_argument("-d", "--due-date",
```
sbillinge, line 50: put back to `-u`

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +113,177 @@ def construct_global_ctx(self):
...
-                new_mil.append(mil)
-                pdoc = {'milestones': mil}
-            if idx > 1:
-                doc = deepcopy(all_milestones[idx - 2])
-                identifier = doc['identifier']
-                if not doc.get(
-                        'type') and not rc.type and identifier == 'milestones':
+        pdoc = {}
+        all_milestones = []
+        if rc.projectum_id and rc.milestone_uuid:
```
sbillinge, line 123: we will need tests for these if we don't have them.  put them into a separate test function called `test_helpers_bad()`.  You can mostly copy and paste from the existing test_helpers to get everything set up, but it won't need too much in teh way of outputs so some of the stuff can be deleted.  I can help if you need it. 

You usse this construction to to test that errors are getting raised properly:
```
        with pytest.raises(RuntimeError) as excinfo:
            actual = collect_appts(people, filter_key=key, filter_value=value, begin_date=start, end_date=end)
        assert str(excinfo.value) == expected
```
This checks that RuntimeError is raised and that the error message is as expected.   Then you can test bad inputs using `pytest.mark.parametrize` similar to the successful test.

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +113,177 @@ def construct_global_ctx(self):
...
+            raise RuntimeError("A uuid fragment and projectum id have been entered.\n"
+                                "You may enter a milestone uuid or a projectum id but not both.\n"
+                                "Enter a milestone uuid to update an existing milestone, or a projectum id to add a new milestone to that projectum.\n")
+        if not rc.projectum_id and not rc.milestone_uuid:
+            raise RuntimeError("No milestone uuid or projectum id was entered.\n"
+                                "Enter a milestone uuid to update an existing milestone, or a projectum id to add a new milestone to that projectum.\n")
+
+        if rc.projectum_id:
+            key = rc.projectum_id
```
sbillinge, line 132: you may have inherited this from before, but let's use this opportunity to clean it up, so just
```
target_prum = rc.client.find_one(rc.database, rc.coll, {'_id': rc.rojectum_id})
```
which I think is easier to read and to maintain.

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +113,177 @@ def construct_global_ctx(self):
...
-            if j['identifier'] == 'milestones' and i not in idx_parsed:
-                new_mil.append(j)
-        for mile in new_mil:
-            del mile['identifier']
-        new_mil.sort(key=lambda x: x['due_date'], reverse=False)
-        pdoc.update({'milestones': new_mil})
-        rc.client.update_one(rc.database, rc.coll, filterid, pdoc)
-        print("{} has been updated in projecta".format(key))
+                        "Please input a valid projectum id or a valid fragment of a projectum id")
+                print("Projecta not found. Projecta with similar names: ")
```
sbillinge, line 141: this should probably be "projectum not found, Projecta with similar names....".  I don't think we are ever looking for more than one prum in this workflow.

This also sometimes fails because it is looking in the wrong database, so it is a valid id but the valid id is in say `billingegroup-public` database but `rc.database` is set to `billingegroup-group`.  For now just mention that as a possible reason why this error was raised and later we can think if we can come up with a better fix (could you post that as an issue actually?).   maybe "if you prum id looks correct, check that this id is in the collection in the database that regolith is looking in (i.e., {rc.database})" or sthg like that.

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +113,177 @@ def construct_global_ctx(self):
...
+            raise RuntimeError("No milestone uuid or projectum id was entered.\n"
+                                "Enter a milestone uuid to update an existing milestone, or a projectum id to add a new milestone to that projectum.\n")
+
+        if rc.projectum_id:
+            key = rc.projectum_id
+            filterid = {'_id': key}
+            target_prum = rc.client.find_one(rc.database, rc.coll, filterid)
+            if not target_prum:
+                pra = fragment_retrieval(self.gtx["projecta"], ["_id"],
+                                         rc.projectum_id)
```
sbillinge, line 137: Please can you check that the rc.projectum_id gets stripped somewhere.   It wasn't before so leading and trailing whitespaces would cause it to fail but they are common with copy-=paste so better behavior is that it can recover from that.

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +113,177 @@ def construct_global_ctx(self):
...
+            for i, j in zip(index_list, new_all):
+                if j['identifier'] == 'milestones':
+                    new_mil.append(j)
+            for mile in new_mil:
+                del mile['identifier']
+            new_mil.sort(key=lambda x: x['due_date'], reverse=False)
+            pdoc.update({'milestones': new_mil})
+            rc.client.update_one(rc.database, rc.coll, filterid, pdoc)
+            print("{} has been updated in projecta".format(key))
+        elif rc.milestone_uuid:
```
sbillinge, line 197: I think this should be an `else`

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +113,177 @@ def construct_global_ctx(self):
...
+            if rc.objective and not target_del:
+                pdoc.update({'objective': rc.objective})
+            if rc.status:
+                pdoc.update({'status': rc.status})
+            if rc.audience:
+                pdoc.update({'audience': rc.audience})
+            if rc.notes:
+                pdoc.update({'notes': rc.notes})
+
+#        rc.index = rc.index.replace(" ", "")
```
sbillinge, line 260: delete all this (you can get it back later with a git checkout if you need it, but you won't...)

sbillinge (commented):
Looks great, nearly there!

Please see my inline comments.

Do you have a good workflow for investigating why the tests are failing and getting them to work?  A 30 min call could be very beneficial if not... Just let me know.

regolith/helpers/u_milestonehelper.py
dpeters001: `-i` for the uuid is a great idea. changing.

regolith/helpers/u_milestonehelper.py
dpeters001: deleted

regolith/helpers/u_milestonehelper.py
dpeters001: changed abbrev back to `-u`

regolith/helpers/u_milestonehelper.py
dpeters001: updated.

regolith/helpers/u_milestonehelper.py
dpeters001: fixed

regolith/helpers/u_milestonehelper.py
dpeters001: :+1:

regolith/helpers/u_milestonehelper.py
dpeters001: Done.

dpeters001:
> Looks great, nearly there!
> 
> Please see my inline comments.
> 
> Do you have a good workflow for investigating why the tests are failing and getting them to work? A 30 min call could be very beneficial if not... Just let me know.

For the tests that fail because of comparison errors, I use the comparison results to determine if the code or the helper test output file needs to be changed. A 30 min call would be great though so we can talk about some of the issues that the test are not capturing. I will probably need help creating the code that tests helper errors.

regolith/helpers/u_milestonehelper.py
dpeters001: I added the additional info to the error statement. Right now the milestone-updater includes an optional argument that allows the user to change `rc.database`. Is that a potential fix? Or were you thinking of a fix that scans multiple databases?

dpeters001:
I just pushed a new update to the code. Couple things:
- The two tests that do not pass are because they are using normal helper tests to test for runtime errors. I am going to start working on the code that will replace these tests.
- For the case in which a user enters multiple uuids, a for loop is used to update the database in each instance. I am not sure if this is the way you wanted to go about this. A print statement will be issued for each update to the projecta db.  If this is the logic you want, maybe we can have the print statement list the uuid or name of each milestone that is updated. Otherwise, I am sure we can figure out a way to update the database once.
- I still need to add tests (in addition to the runtime error tests) that test for errors raise by the deliverable and kickoff ids. Im sure there is other functionality not being tested as well. We can look at that to be sure we can test for as much as possible.

regolith/helpers/u_milestonehelper.py
```diff
@@ -27,23 +27,15 @@ def subparser(subpi):
...
-                            "format of 2,5,7 or 3-7 for multiple indices.",
-                       type=str)
-    subpi.add_argument("-v", "--verbose", action="store_true",
-                       help="Increases the verbosity of the output.")
+    subpi.add_argument("-i", "--milestone_uuid",
+                       help="The uuid of a milestone. "
+                            "Takes a full or partial uuid. "
+                            "Multiple uuids may be entered.",
+                       nargs='+')
+    subpi.add_argument("--projectum_id", help="The id of the projectum. If you "
```
sbillinge, line 35: shall we give this an abbreviation `-p`?

regolith/helpers/u_milestonehelper.py
sbillinge: please change "projecta" to "projectum" in the first word of the error message.  Yes, this was the fix I was asking for.  Again, to be kinder to the user, it is always better to tell them not so much what is wrong but what to do differently, let's add a statement at the end "...(i.e., {rc.database}).  In this case rerun with --database set to the database where the item is located."   Sthg like that.

regolith/helpers/u_milestonehelper.py
sbillinge: I set up a meeting for tomorrow.  We can discuss.

tests/outputs/u_milestone/projecta.yaml
```diff
@@ -187,7 +187,7 @@ sb_firstprojectum:
   deliverable:
     audience:
       - beginning grad in chemistry
-    due_date: 2020-06-01
+    due_date: '2021-05-05'
```
sbillinge, line 190: you are changing the test outputs here.   This will have a lot of knock on effects.  That you have to change it maybe suggests something needs to be changed elsewhere?  We can discuss.

sbillinge (commented):
please see comments.  Let's talk tomorrow.

PR #1003, opened 2023-01-24
U milestone UUID new

### LINKED ISSUE #987: milestone updater uses uuid to select which milestone to update

dpeters001:
Closes #987

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -206,7 +206,7 @@ class MilestoneUpdaterHelper(DbHelperBase):
                             id.append(pid)
                 if target_ko:
                     for prum in target_ko:
-                        if prum.get('_id')[0:len(uuid)] == uuid[2:]:
+                        if prum.get('_id')[0:len(uuid)-2] == uuid[2:]:
                             kickoff = prum['kickoff']
                             upd_mil.append(kickoff)
                         if upd_mil:
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -126,7 +126,8 @@ class MilestoneUpdaterHelper(DbHelperBase):
                     print(f"{pra[i].get('_id')}")
                 print("Please rerun the helper specifying the complete ID.\n"
                       "If your prum id looks correct, check that this id is in the collection "
-                      "in the database that regolith is looking in (i.e., {rc.database}).")
+                      "in the database that regolith is looking in (i.e., {rc.database}).\n"
+                      "If this is the case, rerun with --database set to the database where the item is located.")
                 return
             milestones = deepcopy(target_prum.get('milestones'))
             all_milestones = []
@@ -179,13 +180,14 @@ class MilestoneUpdaterHelper(DbHelperBase):
             print("{} has been updated in projecta".format(rc.projectum_id))
         else:
             for uuid in rc.milestone_uuid:
+                pdoc = {}
                 upd_mil = []
                 all_miles = []
                 id = []
                 target_mil = fragment_retrieval(self.gtx["projecta"], ["milestones"], uuid)
                 target_del = fragment_retrieval(self.gtx["projecta"], ["_id"], uuid)
                 target_ko = fragment_retrieval(self.gtx["projecta"], ["_id"], uuid[2:])
-                if target_mil:
+                if target_mil and not target_del and not target_ko:
                     for prum in target_mil:
                         milestones = prum['milestones']
                         for milestone in milestones:
@@ -204,7 +206,7 @@ class MilestoneUpdaterHelper(DbHelperBase):
                         if upd_mil:
                             pid = prum.get('_id')
                             id.append(pid)
-                if target_ko:
+                if target_ko and uuid[:2] == 'ko':
                     for prum in target_ko:
                         if prum.get('_id')[0:len(uuid)-2] == uuid[2:]:
                             kickoff = prum['kickoff']
@@ -212,7 +214,11 @@ class MilestoneUpdaterHelper(DbHelperBase):
                         if upd_mil:
                             pid = prum.get('_id')
                             id.append(pid)
-                if len(upd_mil) == 1:
+                if len (upd_mil) == 0:
+                    print(f"No ids were found that match your entry ({uuid}).\n"
+                          "Make sure you have entered the correct uuid or uuid fragment and rerun the helper.")
+                    return
+                elif len(upd_mil) == 1:
                     for dict in upd_mil:
                         pdoc.update(dict)
                     for i in all_miles:
@@ -254,15 +260,15 @@ class MilestoneUpdaterHelper(DbHelperBase):
                     pdoc['due_date'] = get_due_date(pdoc)
                     all_miles.append(pdoc)
                     all_miles.sort(key=lambda x: x['due_date'], reverse=False)
-                    if target_mil:
+                    if target_mil and not target_del and not target_ko:
                         doc.update({'milestones': all_miles})
                     if target_del:
                         doc.update({'deliverable': pdoc})
-                    if target_ko:
+                    if target_ko and uuid[:2] == 'ko':
                         doc.update({'kickoff': pdoc})
                     rc.client.update_one(rc.database, rc.coll, {'_id': id[0]}, doc)
-                    print("{} has been updated in projecta.".format(id[0]))
+                    print("The milestone uuid {} in {} has been updated in projecta.".format(uuid, id[0]))
                 else:
-                    print("Multiple ids match your entry(s).\n"
-                          "Rerun the helper and include more characters of each id.")
+                    print(f"Multiple ids match your entry ({uuid}).\n"
+                          "Try entering six or more characters of the uuid and rerun the helper.")
         return
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -32,7 +32,7 @@ def subparser(subpi):
                             "Takes a full or partial uuid. "
                             "Multiple uuids may be entered.",
                        nargs='+')
-    subpi.add_argument("--projectum_id", help="The id of the projectum. If you "
+    subpi.add_argument("-p", "--projectum_id", help="The id of the projectum. If you "
                                             "opt for this the program will assume "
                                             "you are adding a new milestone "
                                             "to the specified projectum.")
@@ -121,7 +121,7 @@ class MilestoneUpdaterHelper(DbHelperBase):
             if not target_prum:
                 pra = fragment_retrieval(self.gtx["projecta"], ["_id"],
                                          rc.projectum_id)
-                print("Projecta not found. Projecta with similar names: ")
+                print("Projectum not found. Projecta with similar names: ")
                 for i in range(len(pra)):
                     print(f"{pra[i].get('_id')}")
                 print("Please rerun the helper specifying the complete ID.\n"
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -27,27 +27,16 @@ def subparser(subpi):
     if isinstance(subpi, GooeyParser):
         date_kwargs['widget'] = 'DateChooser'
 
-    subpi.add_argument("-u", "--milestone_uuid",
+    subpi.add_argument("-i", "--milestone_uuid",
                        help="The uuid of a milestone. "
                             "Takes a full or partial uuid. "
-                            "Multiple uuids may be entered.")
-    subpi.add_argument("--projectum_id", help="The id of the projectum.  If you "
+                            "Multiple uuids may be entered.",
+                       nargs='+')
+    subpi.add_argument("--projectum_id", help="The id of the projectum. If you "
                                             "opt for this the program will assume "
                                             "you are adding a new milestone "
                                             "to the specified projectum.")
-#    subpi.add_argument("-c", "--current", action="store_true",
-#                       help="only list current (unfinished, unpaused) milestones",
-#                       )
-#    subpi.add_argument("-i", "--index",
-#                       help="Index of the item in the enumerated list to update. "
-#                            "To update multiple milestones with the same edits "
-#                            "(often used for finishing checklist items), "
-#                            "please enter in the "
-#                            "format of 2,5,7 or 3-7 for multiple indices.",
-#                       type=str)
-#    subpi.add_argument("-v", "--verbose", action="store_true",
-#                       help="Increases the verbosity of the output.")
-    subpi.add_argument("-d", "--due-date",
+    subpi.add_argument("-u", "--due-date",
                        help="New due date of the milestone. "
                             "Required for a new milestone.",
                        **date_kwargs)
@@ -119,7 +108,6 @@ class MilestoneUpdaterHelper(DbHelperBase):
             now = dt.date.today()
         new_mil = []
         pdoc = {}
-        all_milestones = []
         if rc.projectum_id and rc.milestone_uuid:
             raise RuntimeError("A uuid fragment and projectum id have been entered.\n"
                                 "You may enter a milestone uuid or a projectum id but not both.\n"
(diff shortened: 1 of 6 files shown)


regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +102,173 @@ def construct_global_ctx(self):
...
+            new_all = deepcopy(all_milestones)
+            for i, j in zip(index_list, new_all):
+                if j['identifier'] == 'milestones':
+                    new_mil.append(j)
+            for mile in new_mil:
+                del mile['identifier']
+            new_mil.sort(key=lambda x: x['due_date'], reverse=False)
+            pdoc.update({'milestones': new_mil})
+            rc.client.update_one(rc.database, rc.coll, {'_id': rc.projectum_id}, pdoc)
+            print("{} has been updated in projecta".format(rc.projectum_id))
```
sbillinge, line 180: please use f-strings

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +102,173 @@ def construct_global_ctx(self):
...
-                new_mil.append(mil)
-                pdoc = {'milestones': mil}
-            if idx > 1:
-                doc = deepcopy(all_milestones[idx - 2])
-                identifier = doc['identifier']
-                if not doc.get(
-                        'type') and not rc.type and identifier == 'milestones':
-                    raise RuntimeError(
-                        "ERROR: This milestone does not have a type set and this is required. "
+                    raise ValueError(
```
sbillinge, line 162: I am not sure I am following the logic, but with `if rc.type in MILESTONE_TYPES:` as the if statement, doesn't that  mean that this block has a VALID type and not an invalid type?

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +102,173 @@ def construct_global_ctx(self):
...
+                    new_mil.append(j)
+            for mile in new_mil:
+                del mile['identifier']
+            new_mil.sort(key=lambda x: x['due_date'], reverse=False)
+            pdoc.update({'milestones': new_mil})
+            rc.client.update_one(rc.database, rc.coll, {'_id': rc.projectum_id}, pdoc)
+            print("{} has been updated in projecta".format(rc.projectum_id))
+        else:
+            for uuid in rc.milestone_uuid:
+                pdoc = {}
```
sbillinge, line 183: to keep the code a bit cleaner, when initalizing use
```
pdoc, upd_mil, all_miles, id = {}, [], [], []

File: tests/test_helpers.py [hunk 4]
```diff
@@ -473,62 +473,35 @@
...
-     "       - do this\n"
-     "       - do that\n"
-     ),
-    (["helper", "u_milestone", "sb_firstprojectum", "--current"],
-     "Please choose from one of the following to update/add:\n"
-     "1. new milestone\n"
-     "3. Project lead presentation    due date: 2020-05-20    status: proposed\n"
-     "4. deliverable    due date: 2020-06-01    status: converged\n"
-     "5. planning meeting    due date: 2020-06-01    status: converged\n"
+    (["helper", "u_milestone", "--milestone_uuid", "milestone_uuid_sb1_", "sb_firstpro", "kosb_firstpro",
```
sbillinge, line 478: Make sure your test does as much work as it can.  Is this updating everything that it can?  I see a bunch of updates not covered by tests above and I think they could be?  In principle, we can use `u_milestone` to update every item in a milestone, so the test can cover everything (but not creating a new milestone because of our clunky testing structure)

tests/test_helpers.py
```diff
@@ -473,62 +473,35 @@
...
+      "--status", "converged", "--due-date", "2020-06-01", "--notes", "do this", "do that", "--type", "meeting"],
+     "The milestone uuid milestone_uuid_sb1_ in sb_firstprojectum has been updated in projecta.\n"
+     "The milestone uuid sb_firstpro in sb_firstprojectum has been updated in projecta.\n"
+     "The milestone uuid kosb_firstpro in sb_firstprojectum has been updated in projecta.\n"
+    ),
+    (["helper", "u_milestone", "--milestone_uuid", "kosb_firtpro",
```
sbillinge, line 484: to make this clearer calll it sthg like `"bad_id"`

tests/test_helpers.py
```diff
@@ -473,62 +473,35 @@
...
+    (["helper", "u_milestone", "--projectum_id", "pl", "--name", "new milestone",
+      "--due_date", "2023-01-01", "--objective", "do all the things to complete this milestone",
+      "--notes", "do this", "do that", "--type", "mergedpr"],
+     "Projectum not found. Projecta with similar names: \n"
+     "pl_firstprojectum\n"
+     "pl_secondprojectum\n"
+     "pl_thirdprojectum\n"
+     "Please rerun the helper specifying the complete ID.\n"
+     "If your prum id looks correct, check that this id is in the collection "
+     "in the database that regolith is looking in (i.e., {rc.database}).\n"
```
sbillinge, line 503: this should not be `{rc.database}` but should resolve to the name of the database in the test.  It is a concern that this test is passing!  In test driven development, always give more thought to your tests than your code.....

tests/test_helpers.py
```diff
@@ -860,6 +833,26 @@
 #      "List of all tags in citations collection:\n['nomonth', 'pdf']\nbuilding lists for all tags in the citation collection\nnomonth has been added/updated in reading_lists\npdf has been added/updated in reading_lists\n"),
 #     ]
 
+helper_map_bad = [
+ (["helper", "u_milestone", "--milestone_uuid", "sb_fir", "--projectum_id", "sb_firstprojectum",
+   "--due-date", "2023-01-01", "--notes", "do this", "do that", "--type", "mergedpr"],
+  "A uuid fragment and projectum id have been entered.\n"
```
sbillinge, line 839: To help the user,
`Detected both a uuid fragment and a projectum id.`

Then below `You may enter either a milestone uuid....`

tests/test_helpers.py
```diff
@@ -860,6 +833,26 @@
...
+  "You may enter a milestone uuid or a projectum id but not both.\n"
+  "Enter a milestone uuid to update an existing milestone, or a projectum id to add a new milestone to that projectum.\n"
+  ),
+ (["helper", "u_milestone", "--due-date", "2023-01-01",
+   "--notes", "do this", "do that", "--type", "mergedpr"],
```
sbillinge, line 844: remove these.  Just put enough to test the thing you are testing.  probably, `["helper", "u_milestone"]`

This makes it easier for someone to understand your intent with fewer brain cells burning

tests/test_helpers.py
```diff
@@ -860,6 +833,26 @@
...
+  "No milestone uuid or projectum id was entered.\n"
+  "Enter a milestone uuid to update an existing milestone, or a projectum id to add a new milestone to that projectum.\n"
+  )]
```
sbillinge, line 847: I think there are a  bunch more of these you can test.  Look above to see the blocks that Codecov show are not covered.  Any of those that result in an exception being thrown can be tested here.

You may have to put the type of error into the paramatrize options (i.e., pass in whether it is `RuntimeError` or `ValueError` or whatever as a variable as part of the test.

sbillinge (commented):
Great job @dpeters001 getting this to pass tests.  I have a couple of concerns with the tests (please scroll to the tests section) and a few comments.

I think when you have addressed these we can merge.  It may be a good idea to make an issue to refactor this a bit for easier reuse.  I think that we will be using this uuid retrieval elsewhere in the future (e.g., in tasks) and so having as much of the work as possible pulled out into functions in tools.py, the better.  But let's get this merged so we can start using it, celebrate, then start thinking about the refactor.  For now, make an issue for two refactors: (1) pulling uuid functionality into functions and (2) making those pesky deliverably and kickoff meeting items as valid milestones.  That will be more of a heavy lift....

regolith/helpers/u_milestonehelper.py
dpeters001: reformatted

regolith/helpers/u_milestonehelper.py
dpeters001: cleaned

regolith/helpers/u_milestonehelper.py
dpeters001: there is an `else:` statement before the `raise ValueError` on line 161.
the code block looks like:

```
            if rc.type:
                if rc.type in MILESTONE_TYPES:
                    mil.update({'type': rc.type})
                else:
                    raise ValueError(
                        "ERROR: The type you have specified is not recognized. "
                        "Please rerun your command adding '--type' "
                        f"and giving a type from this list: {MILESTONE_TYPES}")
```

tests/test_helpers.py
dpeters001: done

regolith/helpers/u_milestonehelper.py
```diff
@@ -110,161 +102,173 @@ def construct_global_ctx(self):
...
-                mil.update({'identifier': "milestones"})
-                new_mil.append(mil)
-                pdoc = {'milestones': mil}
-            if idx > 1:
-                doc = deepcopy(all_milestones[idx - 2])
-                identifier = doc['identifier']
-                if not doc.get(
-                        'type') and not rc.type and identifier == 'milestones':
-                    raise RuntimeError(
-                        "ERROR: This milestone does not have a type set and this is required. "
```
dpeters001, line 215: @sbillinge I removed the RuntimeError that was previously on lines 214-215. Are there any cases in which existing milestones would not have a meeting type?  I will put it back in the code for now because there are milestones on the schema without `type` specified.

tests/test_helpers.py
dpeters001: 👍

tests/test_helpers.py
dpeters001: :+1:

dpeters001: makes sense

regolith/helpers/u_milestonehelper.py
sbillinge: It should be there.  We want `type` to be required and selected from an `eallowed` list

regolith/helpers/u_milestonehelper.py
sbillinge: :+1: Thanks.  It was hard to see that, so better to check.

tests/test_helpers.py
dpeters001: right. the reason it was passing is that I had the entire phrase written as a part of the statement. I misunderstood your previous comment about providing database info to the user. will change the test to reflect the correct usage of `{rc.database}` information.

tests/test_helpers.py
```diff
@@ -860,6 +840,55 @@
...
+  "Please rerun your command adding '--type' \n"
+  "and giving a type from this list:\n"
+  "['mergedpr', 'meeting', 'other', 'paper', 'release', 'email', 'handin', 'purchase', 'approval', 'presentation', 'report', 'submission', 'decision', 'demo', 'skel']\n"),
+ ]
+
+@pytest.mark.parametrize("hmb", hm_val_err)
+def test_helpers_bad(hmb, make_db):
+ repo = Path(make_db)
+ os.chdir(repo)
+ with pytest.raises(ValueError) as excinfo:
```
sbillinge, line 888: I think we can catch everything in one test block if you make the inputs 
```
(["helper", "u_milestone", "--milestone_uuid", "milestone_uuid_sb1_", "--type", "bad_type"], <output string>, RunTimeError])
```
then your code block looks like:
```
def test_helpers_bad(hmb, make_db):
 repo = Path(make_db)
 os.chdir(repo)
 with pytest.raises(hm[2]) as excinfo:
  actual =  main(args=hmb[0])
 assert str(excinfo.value) == hmb[1]
```
and then you switch between `RunTimeError` and `ValueError` in the inputs as needed.  Can you give it a try?

tests/test_helpers.py
```diff
@@ -860,6 +840,55 @@
...
+   "-n", "new milestone", "-o", "complete ms", "--type", "bad_type"],
+  "The type you have specified is not recognized. \n"
+  "Please rerun your command adding '--type' \n"
+  "and giving a type from this list:\n"
+  "['mergedpr', 'meeting', 'other', 'paper', 'release', 'email', 'handin', 'purchase', 'approval', 'presentation', 'report', 'submission', 'decision', 'demo', 'skel']\n"),
+ (["helper", "u_milestone", "--milestone_uuid", "milestone_uuid_sb1_", "--type", "bad_type"],
+  "The type you have specified is not recognized. \n"
+  "Please rerun your command adding '--type' \n"
+  "and giving a type from this list:\n"
+  "['mergedpr', 'meeting', 'other', 'paper', 'release', 'email', 'handin', 'purchase', 'approval', 'presentation', 'report', 'submission', 'decision', 'demo', 'skel']\n"),
```
sbillinge, line 881: Are these being used?  I think they can be deleted.  What are they for?

sbillinge (commented):
Fantastic job!  Nearly there, just one last thing (see inline).  I think you misunderstood one of my requests.  Better to fix it now or there will be confusing junk left to confuse our future selves.

Amazing job!  I'm very impressed.

tests/test_helpers.py
dpeters001: Perfect! It is all working in one test block now. I tried this yesterday but there were issues. Probably a formatting error on my part. Thanks for writing it out.

tests/test_helpers.py
dpeters001: These test the value errors related to the `--type` option and `type` key. They have now been combined with the the other `helper_map_bad` tests

regolith/helpers/u_milestonehelper.py
dpeters001: Done.

tests/test_helpers.py
```diff
@@ -840,52 +840,49 @@
...
+  "Specify '--type' and rerun the helper to update this milestone.\n",
+  ValueError
   ),
  (["helper", "u_milestone", "--projectum_id", "sb_firstprojectum", "-u", "2020-06-01",
    "-n", "new milestone", "-o", "complete ms", "--type", "bad_type"],
   "The type you have specified is not recognized. \n"
   "Please rerun your command adding '--type' \n"
   "and giving a type from this list:\n"
-  "['mergedpr', 'meeting', 'other', 'paper', 'release', 'email', 'handin', 'purchase', 'approval', 'presentation', 'report', 'submission', 'decision', 'demo', 'skel']\n"),
+  "['mergedpr', 'meeting', 'other', 'paper', 'release', 'email', 'handin', 'purchase', 'approval', 'presentation', 'report', 'submission', 'decision', 'demo', 'skel']\n",
```
sbillinge, line 870: This would be better if it can be done with an f-string that expands this as a variable that is imported from schema.  Does it work if you do it that way?  It will be more robust because if we add something to list of allowed values in schema.py the test will automatically update.

sbillinge (commented):
please see my comment

tests/test_helpers.py
dpeters001: Yes. Nice catch. The test still works with an f-string.

sbillinge:
whoohoo, look at all that green!
image

dpeters001:
> whoohoo, look at all that green! image

Woo! 🙌 Onto the next!

sbillinge:
👍

PR #1007, opened 2023-01-27
fixes bug that raises valueerror after print statement when using --f…

dpeters001:
…inish on  ko and deliverable milestones. if a deliverable does not have a type, the code will finish the ms. if the ko does not have a type, a valueerror is raised that instructs user to rerun the helper with --type specified.

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -218,6 +218,18 @@ class MilestoneUpdaterHelper(DbHelperBase):
                 elif len(upd_mil) == 1:
                     for dict in upd_mil:
                         pdoc.update(dict)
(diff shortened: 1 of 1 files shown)


dpeters001:
Is it ok that deliverables do not require a value for `type:` ?

sbillinge (commented):
Thanks for the fix.  

Please can you post another issue and add it to the release milestone to fix the "success" messages.  This is printed at the point of the update, but I think the database itself is not updated until the dump at the end, so if, as in this case, the milestone was finished but then the code crashed before the dump, the database is actually not updated.  This will be confusing to people.  What we could do  is update a variable with a string in the code where the update happens, but then have the print statement _after_ the dump at the end and what is printed is this variable.  Something like that?  But let's make that an issue and fix it on a separate branch.

PR #1008, opened 2023-01-28
ko and deliverable milestones can be updated/finished without a type.…

dpeters001:
… fixes valueerror raised when trying to use --finish on ko milestones without types.


sbillinge:
Thanks @dpeters001 this looks like a better fix.  I was pulling my hair a bit about using pattern recognition of an input variable in if statements.  this of course is the horrible technical debt we have of the original design decision to make deliverables and kickoffs not valid milestones so that is the proper fix.  I htink that you solution now is a good intermediate level fix, though I am not sure if the current flow will break if I give a list of milestones that is a mixture of ko  - del and regular milestones.  Do you know?  Let's cross that bridge when we come to it.  It may be better to hold our nose and do the milestone refactor, though it will affect huge databases the already exist and will have to be migrated.....

PR #1010, opened 2023-01-30
a_todohelper.py: adds cl opt arg --milestone_uuid that adds todo to a...

### LINKED ISSUE #988: new milestones tasks logic in a_todos
modify a_todos so a todo item can be assigned to a milestone as an option (by giving a milestone uuid fragment).  Write tests for everything first, get them reviewed then write the code

- [x] add milestone uuid to cli parser
- [x] write a function in regolith tools module that does the adding (this will be reused in other places)
- [x] if rc.milestone_uuid then trigger logic that adds the uuid of this task to the task list of the given milestone and updates that prum (get that logic from u_milestones...`rc.client.update_one()`)
- [x] news

dpeters001:
...ms task list, logic added for what to do if a milestone_uuid is entered. 

regolith/tools.py: adds the fucntion add_todo_id_to_tasks that does the adding of the todo uuid to the tasks list of a milestone

f_prum/projecta.yaml: outputs changed to reflect new a_todo functionality 

test_helpers.py: two tests added for SystemExit instances; because this part of the code exits the a_todohelper, the print statements that notify the user of the reason for the exit are not captured by the tests but they will print in the cl and gui

closes #988

### DIFF
--- a/regolith/tools.py
+++ b/regolith/tools.py
@@ -20,7 +20,7 @@ from google.auth.transport.requests import Request
 from google.oauth2.credentials import Credentials
 from urllib.parse import urlparse
 
-from regolith.dates import month_to_int, date_to_float, get_dates, is_current
+from regolith.dates import month_to_int, date_to_float, get_dates, is_current, get_due_date
 from regolith.sorters import id_key, ene_date_key, \
     doc_date_key_high
 from regolith.schemas import APPOINTMENTS_TYPES, PRESENTATION_TYPES, \
@@ -2315,3 +2315,42 @@ def get_appointments(person, appointments, target_grant=None):
             appointments.append((person.get('_id'), bd, ed, appt.get('loading'),
                          round(weighted_duration, 2), appt.get('grant')))
     return appointments
+
+def add_todo_id_to_tasks(gtx, client, db, coll, milestone_id, task_id):
+    pdoc, upd_mil, all_miles, id, success = {}, [], [], [], []
+    target_mil = fragment_retrieval(gtx["projecta"], ["milestones"], milestone_id)
+    if target_mil:
+        for prum in target_mil:
+            milestones = prum['milestones']
+            for milestone in milestones:
+                if milestone.get('uuid')[0:len(milestone_id)] == milestone_id:
+                    upd_mil.append(milestone)
+                else:
+                    all_miles.append(milestone)
+            if upd_mil:
+                pid = prum.get('_id')
+                id.append(pid)
+    if len(upd_mil) == 0:
+        print(f"No ids were found that match your entry ({milestone_id}).\n"
+              "Make sure you have entered the correct milestone uuid or uuid fragment and rerun the helper.")
+        return
+    elif len(upd_mil) == 1:
+        doc = {}
+        for dict in upd_mil:
+            pdoc.update(dict)
+        for i in all_miles:
+            i['due_date'] = get_due_date(i)
+        pdoc.update({'tasks': task_id})
+        pdoc['due_date'] = get_due_date(pdoc)
+        all_miles.append(pdoc)
+        all_miles.sort(key=lambda x: x['due_date'], reverse=False)
+        doc.update({'milestones': all_miles})
+        client.update_one(db, coll, {'_id': id[0]}, doc)
+        success.append(f"The milestone uuid {milestone_id} in {id[0]} has been updated in projecta.")
+    else:
+        print(f"Multiple ids match your entry ({milestone_id}).\n"
+              "Try entering six or more characters of the uuid and rerunning the helper.")
+        return
+
+    return success[0]
+
--- a/regolith/tools.py
+++ b/regolith/tools.py
@@ -2317,6 +2317,30 @@ def get_appointments(person, appointments, target_grant=None):
     return appointments
 
 def add_todo_id_to_tasks(gtx, client, db, coll, milestone_id, task_id):
+    '''
+    add task_uuid to milestone tasks
+
+    Parameters
+    --------
+    gtx: dict
+        The global context from which todos and milestones will be queried and updated
+    client: client manager object
+        client database backend
+    db: string
+        database name where the milestone is located
+    coll: string
+        the collection of the db in which the milestone is located
+    milestone_id: string
+        the uuid of the milestone to which the todo uuid will be added
+    task_id: string
+        the uuid of the todo
+
+    Returns
+    -------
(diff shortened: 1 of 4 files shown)


regolith/tools.py
```diff
@@ -2315,3 +2315,42 @@ def get_appointments(person, appointments, target_grant=None):
             appointments.append((person.get('_id'), bd, ed, appt.get('loading'),
                          round(weighted_duration, 2), appt.get('grant')))
     return appointments
+
+def add_todo_id_to_tasks(gtx, client, db, coll, milestone_id, task_id):
```
sbillinge, line 2319: This needs a test and a docstring.
1) write the docstring first.  It will help you understand what the function is supposed to do
2) then write the tests for expected behavior under different circumstances.
3) then tweak the code so the tests pass.

In general, try and make functions do just one thing.  This function _may_ be doing too much (not sure actually, it may be ok).  If they are doing two things, consider splitting it to two functions where on function calls another.

regolith/tools.py
```diff
@@ -2315,3 +2315,42 @@ def get_appointments(person, appointments, target_grant=None):
...
+    pdoc, upd_mil, all_miles, id, success = {}, [], [], [], []
+    target_mil = fragment_retrieval(gtx["projecta"], ["milestones"], milestone_id)
+    if target_mil:
```
sbillinge, line 2322: this seems to be missing an `else:`

regolith/tools.py
```diff
@@ -2315,3 +2315,42 @@ def get_appointments(person, appointments, target_grant=None):
...
+        for prum in target_mil:
```
sbillinge, line 2323: this looks rather inefficient.  Is there not a `get_one` that can pull this directly from the db?

regolith/tools.py
```diff
@@ -2315,3 +2315,42 @@ def get_appointments(person, appointments, target_grant=None):
...
+                id.append(pid)
+    if len(upd_mil) == 0:
+        print(f"No ids were found that match your entry ({milestone_id}).\n"
+              "Make sure you have entered the correct milestone uuid or uuid fragment and rerun the helper.")
+        return
+    elif len(upd_mil) == 1:
+        doc = {}
+        for dict in upd_mil:
+            pdoc.update(dict)
+        for i in all_miles:
```
sbillinge, line 2341: try and use list comprehension wherever you can instead of for loops.

regolith/tools.py
```diff
@@ -2315,3 +2315,42 @@ def get_appointments(person, appointments, target_grant=None):
...
+            milestones = prum['milestones']
+            for milestone in milestones:
+                if milestone.get('uuid')[0:len(milestone_id)] == milestone_id:
+                    upd_mil.append(milestone)
+                else:
+                    all_miles.append(milestone)
+            if upd_mil:
+                pid = prum.get('_id')
+                id.append(pid)
+    if len(upd_mil) == 0:
```
sbillinge, line 2333: maybe handle this at the top?

tests/outputs/f_prum/projecta.yaml
```diff
@@ -261,6 +261,7 @@ sb_firstprojectum:
       notes:
         - do this
         - do that
+      tasks: test-uuid
```
sbillinge, line 264: I think tasks should be a list

sbillinge (commented):
great start. please see inline comments.

The adders and updaters all require a database as well as a collection.  This means that we can directly pull items from the datase, we don't have to iterate through collections.  But we do have to handle error messages where the user specified the wrong database (i.e., the id can't be found not because it is not there in the collection but it is in a different database). Reach out if any of this is confusing.

tests/outputs/f_prum/projecta.yaml
dpeters001: fixed.

regolith/tools.py
dpeters001: fixed.

regolith/tools.py
dpeters001: included in an else statement that completes the if statement: `if target_mil` so that it is handled at the top.

regolith/tools.py
sbillinge: per convo with Tom, let's leave this as is.  `fragment_retrieval` is maybe not hte best name for that function though. 
I wonder whether we want a function called something like `get_subcollection(collection, key)` that would return a list if key in collection  is a list, else returns some kind of failure key.  This seems like a general thing we might want to do.  Then `get_subcollection(gtx["projecta"], "milestones")` would return the list of milestones and we can do things with the milestones (including search it for an id, for example `[milestone for milestone in milestones if milestone.get('milestone_id') == milestone_id]`

btw, I think gtx["projecta"] is a generator so it empties when you iterate through it.  We may want to turn it into a list before passing it to a function,  though it will work this way if we only use it once.

sbillinge: back up a bit there, I am not sure what the functions are doing.  I guess the steps are:

1. figure out which prum contains the target milestone_id (this probably best done with an index when we figure out how to do that, but for now iterate through all the milestones in all the prums?). 
2. Get that prum
4. update tasks in that milestone by appending the task_uuid
5. update_one for that prum

It could then look like this.
```
target_prum = [prum for prum in prums if milestone.get("uuid") == target_uuid for milestone in prum.get('milestones') ]
if not target_prum:
    exception handling
else:
    for milestone in target_prum.get('milestones'):
        if milestone_id == target_id:
            milestone.get(tasks).append("uuid": task_id)
```
But I think if we had the get_subcollections it could look something like
```
for prum in prums:
    milestones = get_subcollection(prum, "milestones")
    milestone = [milestone for milestone in milestones if milestone.get("uuid") == target_id]
    tasks = get_subcollection(milestone[0], "tasks")
    tasks.append(task_id)
```
you will have to check if the append happens in place, in other words, does the whole prum object reflect this change (I am not sure), but if it does, then you would just have to do an `update_one` and push the updated prum to the db.

I am not sure about the syntax, but sthg like that.

sbillinge (commented):
please see comments

sbillinge:
I suggest to make a new branch for the get_subcollection() function, make tests then make the function and get that merged, then return to this branch and use it.

sbillinge:
@dpeters001 I got the ci working using GH Actions in the end.  To make it work
```
git checkout a_todo_ms_uuid
git pull upstream master -m "merging new ci"
git push
```
and it should start to build.   Do this also on your other PR.  

Also, I made PR that should fix the "too many databases problem".  Take a quick look.

PR #1011, opened 2023-01-31
u_milestonehelper.py: success and failure messages printed after db 

### LINKED ISSUE #1009: u_milestone "success" message printed after database update

dpeters001:
is updated. if multiple ids were entered, the user will see which milestones were successfully updated and which ids could not be found or which ids had multiple matches. test_helpers.py: u_milestone tests reflect the new printing sequence.

Closes #1009

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -21,7 +21,6 @@ TARGET_COLL = "projecta"
 MILESTONE_TYPES = MILESTONE_TYPES
 PROJECTUM_STATI = PROJECTUM_STATI
 
-
 def subparser(subpi):
     date_kwargs = {}
     if isinstance(subpi, GooeyParser):
@@ -29,7 +28,7 @@ def subparser(subpi):
 
     subpi.add_argument("-i", "--milestone_uuid",
                        help="The uuid of a milestone. "
-                            "Takes a full or partial uuid.",)
+                            "Takes a full or partial uuid.")
     subpi.add_argument("-p", "--projectum_id", help="The id of the projectum. If you "
                                             "opt for this the program will assume "
                                             "you are adding a new milestone "
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -7,6 +7,7 @@ from itertools import chain
 import datetime as dt
 import dateutil.parser as date_parser
 from gooey import GooeyParser
+from gooey.gui.components.widgets import textfield
 
 from regolith.helpers.basehelper import DbHelperBase
 from regolith.fsclient import _id_key
@@ -99,15 +100,13 @@ class MilestoneUpdaterHelper(DbHelperBase):
         gtx["float"] = float
         gtx["str"] = str
         gtx["zip"] = zip
-
     def db_updater(self):
         rc = self.rc
         if rc.date:
             now = date_parser.parse(rc.date).date()
         else:
             now = dt.date.today()
-        new_mil = []
-        pdoc = {}
+        new_mil, pdoc, success, fail = [],{},[],[]
         if rc.projectum_id and rc.milestone_uuid:
             raise RuntimeError("Detected both a uuid fragment and projectum id.\n"
                                 "You may enter either a milestone uuid or a projectum id but not both.\n"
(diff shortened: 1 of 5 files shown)


dpeters001:
Take a look at the new printing structure that prints success and failure messages at the end of the script. I am still not sure if this solves this issue. Hopefully you can point me in the right direction.

dpeters001:
@sbillinge I did find a way to fix the notes text box. We can add the following widget

`widget='Textarea', gooey_options={'height':100}` its looks great but it is making all of the test fail so I did not include it in this commit. I will try to get a better understanding as to the cause of the tests failing due to the additional widget argument.

regolith/helpers/u_milestonehelper.py
```diff
@@ -272,8 +270,16 @@ def db_updater(self):
...
                     if not rc.finish:
-                        print(f"The milestone uuid {uuid} in {id[0]} has been updated in projecta.")
+                        success.append(f"The milestone uuid {uuid} in {id[0]} has been updated in projecta.")
                 else:
-                    print(f"Multiple ids match your entry ({uuid}).\n"
-                          "Try entering six or more characters of the uuid and rerunning the helper.")
+                    fail.append(f"Multiple ids match your entry ({uuid}).\n"
+                          "Try entering six or more characters of the uuid and rerunning the helper "
+                          f"to update milestone ({uuid}).")
+        if success:
```
sbillinge, line 278: This is basically it but the logic breaks here because, in principle, you could have success and failure but that is not possible.

Maybe something like:
```
print(f"{*fail,}") if fail else print(f"{*success,}")
```
I am not sure if this will give the right behavior, but we can try.  

Also, I think there may be a logic problem higher up that it can still update a database that doesn't contain the prum under some circumstance (I found it when finishing a milestone).  This actually may be the main problem, that it succeeds when it should fail.

sbillinge (commented):
please see comment

sbillinge:
> @sbillinge I did find a way to fix the notes text box. We can add the following widget
> 
> `widget='Textarea', gooey_options={'height':100}` its looks great but it is making all of the tu_milestone tests fail so I did not include it in this commit. I have to figure out how to add this new widget argument to the allowable arguments.

yes, this should be on a separate branch.   Also, please check the behavior.  I think that we tried to add this before and it didn't give the behavior we wanted so we went back.  The key is to try and figure out from the gooey documentation how things typed in there are parsed into strings.  Ideally we would like each row to be a separate string item in a list, but I don't think it does that by default.

regolith/helpers/u_milestonehelper.py
dpeters001: I think the problem is the logic as well. I am not sure we can use the updater to update multiple milestones without an `update_many` function like the one in pymongo or mongo query language. We only have the `update_one` to work with right now. I think it will be best for me to remove the loop that allows the user to input multiple uuids. The way the code is written allows for success and failure in instances in which some milestone uuids are located in a database and others are not. Since most users are updating one milestone at a time, I do not think it is a major hindrance to take remove this functionality until we can utilize the `update_many` function. The `update_many` function will allow us to make sure there are no errors with any of the ids or milestones before the helper updates the database.

sbillinge:
Yes,  one thing at a time.  Please make an issue for multi milestone
updates but remove the functionality from the current pr

dpeters001:
@sbilligne do you know why the circleci tests are not running on this pull request?

sbillinge:
@dpeters001 we ran out of credits.  I will work on it but at travis-ci it normally takes a day or two because a human being over there has to approve it....

sbillinge:
we started migrating all our ci over to github actions....maybe time to finish that job! :)

sbillinge:
@dpeters001 please can you make a small change and push another commit to the PR to see if it triggers a build?  Thanks!

dpeters001:
> @dpeters001 we ran out of credits. I will work on it but at travis-ci it normally takes a day or two because a human being over there has to approve it....

ahh I see! I will make a change and push.

dpeters001:
> we started migrating all our ci over to github actions....maybe time to finish that job! :)

😆

dpeters001:
> @dpeters001 please can you make a small change and push another commit to the PR to see if it triggers a build? Thanks!

@sbillinge still no build.

dpeters001:
@sbillinge I was able to add the `Textarea` widget to the `--notes` option and have all tests pass. I ended up removing the widget, however. The widget does not allow for multi-line text that has `\n` delimiters. The only way you can have multi-line text is if the text wraps from the previous line. I looked through the git documentation and there are no additional option for this. Someone raised this issue previously: [https://github.com/chriskiehl/Gooey/issues/499].
The solution was to turn off the `requires_shell` global option. Out of curiosity, I set the global config `@gooey.Gooey(requires_shell=False)` and the program was not able to find the python program. So this does not seem to be the correct fix either.

sbillinge:
@dpeters001 I plan to close this PR and not merge it. I think most of the issues are better handled either by #1014 or can be more easily fixed by building off of that PR.  I didn't merge it because I wanted you to look at it. Did you look at it?  I think it is fundamentally a better approach.   Actually, I will go ahead and merge it as you can still find it as a closed PR, it is just harder to find, but please look at it and let me know what you think.  It will still need some thought if we want to allow users to update multiple milestones across multiple databases.  Maybe we should simply not allow that, at least for the time being.

PR #1017, opened 2023-02-09
u_milestonehelper.py: removed functionality that allowed the user to …

### LINKED ISSUE #1009: u_milestone "success" message printed after database update

dpeters001:
…input multuple ms uuids. Fixed the problem of having a mix of success and failure messages. test_helper.py: u_milestone tests updated to reflect diminished functionality. only one uuid is tested at a time.

Closes #1009
@dpeters001  if you add this closes statement here (see above, I did it) the relevant issue will automatically close when the  PR merges.

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -29,7 +29,11 @@ def subparser(subpi):
 
     subpi.add_argument("-i", "--milestone_uuid",
                        help="The uuid of a milestone. "
+<<<<<<< HEAD
                             "Takes a full or partial uuid. ")
+=======
+                            "Takes a full or partial uuid.")
+>>>>>>> 76214c70 (added test to test_helpers.py that tests the --finsih arg in u_milestonehelper.py)
     subpi.add_argument("-p", "--projectum_id", help="The id of the projectum. If you "
                                             "opt for this the program will assume "
                                             "you are adding a new milestone "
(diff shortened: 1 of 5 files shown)


dpeters001:
@sbillinge when you get a chance please check out my last two commits. I made commit (235f24b) without syncing my branch `u_milestone_rm_loop` to your commit (e5510a8). I rebased the branch in order to push my commits. Does that make sense? I have no problem starting over if this will be an issue.

sbillinge:
@dpeters001 please kill this and do a new branch, because the structure has changed high up in the code it is better to start over, but I think the changes may only be a few lines at this point.  We could even leave in the multiple uuids actually with this behavior:

> 1) try and find the uuids in the current db
> 2) if they are all found, go for it
> 3) if some are found, do the ones that are found and let the user know that `[these]` were done but `[these]` were not found, maybe they are in a different database than the one specified?
> 4) if none were found, message that none were found, maybe in a different db

I think this will work nicely actually.

dpeters001:
@sbillinge seems that some of the tests in test/test_validate.py are failing.

dpeters001:
> @dpeters001 please kill this and do a new branch, because the structure has changed high up in the code it is better to start over, but I think the changes may only be a few lines at this point. We could even leave in the multiple uuids actually with this behavior:
> 
> > 1. try and find the uuids in the current db
> > 2. if they are all found, go for it
> > 3. if some are found, do the ones that are found and let the user know that `[these]` were done but `[these]` were not found, maybe they are in a different database than the one specified?
> > 4. if none were found, message that none were found, maybe in a different db
> 
> I think this will work nicely actually.

Will do!

PR #1018, opened 2023-02-16
u_milestonehelper.py: only one milestone uuid input allowed now.

### LINKED ISSUE #1009: u_milestone "success" message printed after database update

dpeters001:
Removes loops for uuids. this has to be removed because the database is only updated after the code is executed...meaning only the last uuid entered would actually be updated. we do not want to confuse the use, so it is best to add the functionality to update multiple milestones later. success and fail messages have been added and their logic has been changed so that they are printed after a successful update or a failure. test helpers | output files : these were updated to test the the code that update a kickoff meeting and the --finish argument

closes #1009

### DIFF
--- a/regolith/helpers/u_milestonehelper.py
+++ b/regolith/helpers/u_milestonehelper.py
@@ -29,9 +29,7 @@ def subparser(subpi):
 
     subpi.add_argument("-i", "--milestone_uuid",
                        help="The uuid of a milestone. "
-                            "Takes a full or partial uuid. "
-                            "Multiple uuids may be entered.",
-                       nargs='+')
+                            "Takes a full or partial uuid.")
     subpi.add_argument("-p", "--projectum_id", help="The id of the projectum. If you "
                                             "opt for this the program will assume "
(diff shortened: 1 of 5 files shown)


sbillinge:
super @dpeters001 fantastic!

Good catch on the problem that all the databases are loaded up front.  We need to address this at some point (soon!) as it breaks all our current use-cases.  However, for now let's merge this, move on to the tasks activity and get that merged.  Then we can look at how to disentangle the bulk loading and dumping for the updater and adder helpers....

The build is ok.  I will remove Travis shortly.  But before I do, Travis started failing on some tests that weren't failing before.  These tests (and more) related to the mongo backend are all failing on the GitHub Actions and so I am skipping them currenlty in the GH actions.  This is why those pass but Travis fails.  Moving forward I would love to fix the tests so they pass in GH actions and I will turn them back on.   Many of them are redundant because they test the same functionality but going through the mongo client.  A better way of testing this is to test the mongo client fully but not have every other test  run through the client (though this is a good "sanity check")
Great job!

PR #1035, opened 2023-09-30
pinning gooey to the version that will load correctly.  It may have to be loaded from pip

sbillinge:


sbillinge:
fixed by #1056

PR #1037, opened 2023-10-03
added kwargs argument to builder subparser in main for builder comman…

davidkim1:
…d to accept kwargs as argument

### DIFF
--- a/news/kwargs.rst
+++ b/news/kwargs.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* new functionality that allows general kwargs to be passed to builders from the command line
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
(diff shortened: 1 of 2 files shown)


sbillinge:
This looks good.  I forgot to tell you to make a news

1. look in the directory called news
2. find a file called `TEMPLATE.rst`
3. copy it to a new name with the `<name of your branch>.rst`
4. This is the information that will go into the changelog when we release it.  Since you added the passing of kwargs in the builder, under "Added" add a bullet that says "new functionality that allows general kwargs to be passed to builders from the command line"
5. Add and commit this file and push to the PR.

Typically there is one news per PR.   Because people are not good at making good commit messages we use this to handle the public-facing release changelog rather than listing commit messages.

PR #1038, opened 2023-10-14
In test builders, added an elif case for publist that runs kwargs wit…

davidkim1:
Added an elif case in test_builder function to test builders to test when bm is publist and kwargs is passed. All theses passed.

### DIFF
--- a/tests/test_builders.py
+++ b/tests/test_builders.py
@@ -108,6 +108,9 @@ def test_builder(bm, db_src, make_db, make_mongodb, monkeypatch):
     elif bm == "grantreport":
         main(["build", bm, "--no-pdf", "--grant", "SymPy-1.1",
               "--from", "2017-04-01", "--to", "2018-03-31"])
+    elif bm == "publist":
+        subprocess.run(["regolith", "build", bm, "--no-pdf",
+                        "--kwargs", "pages-e103"], check=True, cwd=repo)
     else:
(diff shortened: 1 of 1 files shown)


tests/test_builders.py
```diff
@@ -108,6 +108,9 @@ def test_builder(bm, db_src, make_db, make_mongodb, monkeypatch):
     elif bm == "grantreport":
         main(["build", bm, "--no-pdf", "--grant", "SymPy-1.1",
               "--from", "2017-04-01", "--to", "2018-03-31"])
+    elif bm == "publist":
+        subprocess.run(["regolith", "build", bm, "--no-pdf",
+                        "--kwargs", "pages-e103"], check=True, cwd=repo)
```
sbillinge, line 113: I proposed a syntax of `key:value` but you used `key-value`.  I think mine may be safer and more intuitive but I am not sure.  What do you think?

This testing syntax is fine, but now it won't test without the kwargs.    To test both with and without we may want to run subprocess twice in that elif clause

sbillinge (commented):
pls see comment

tests/test_builders.py
davidkim1: Got it. Will add another subprocess.run without the kwargs and change it to a key:value syntax and commit a PR.

sbillinge:
thanks @davidkim1

PR #1039, opened 2023-10-16
added scopatz_page.bib and scoptatz.tex in test/outputs/publist for t…

davidkim1:
…he testing --kwargs


sbillinge (commented):
This looks good, but please could you look at what the output looks like for the cases where arguments such as --from --to and --grant are passed in and make this look similar?  Both the file-name and the heading inside the document are changed to include information about what was passed in, so this should also be.  The title will go from "Publication Listing" to "Publication Listing - e103 in pages" or something like that.   Then in our actual use-case it would read "Publication Listing - nslsii in facility".  The filename should also look something like this.

It would be nice to also have a test where `--from` and `--to` are passed in as well as facility and grants.  In principal we might want to test a mix and match.  

Maybe a more interesting test for the `--kwargs` might be `doi:10.10221` rather than pages:e309

PR #1042, opened 2023-10-19
Publist kwarg builder

davidkim1:
Changed the capital L's in Public Listing in the publist templates to follow the style you suggested.


regolith/builders/publistbuilder.py
```diff
@@ -76,6 +76,26 @@ def latex(self):
...
             filestub = filestub + "{}".format(cat_grants)
             qualifiers = qualifiers + " from Grant{} {}".format(pl, text_grants)
+        if self.rc.kwargs:
+            kw = True
+            kwargs = self.rc.kwargs
+            if isinstance(kwargs, str):
+                key, value = kwargs.split(':', 1)  # Split only at the first colon
+                if key == 'facilities':
+                    facilities = value
+                    filestub = filestub + "_facility_{}".format(facilities)
```
sbillinge, line 86: ok here, but please use f-strings for formatted strings in future

regolith/builders/publistbuilder.py
```diff
@@ -76,6 +76,26 @@ def latex(self):
...
+                    qualifiers = qualifiers + " from facility {}".format(facilities)
+                elif key == 'doi':
```
sbillinge, line 88: i don't think we want this, so remove it.  I guess you need it to make the tests pass, but it is polluting the code.  It presumably means we have to update the schema :(

I will merge this so I can use the code, but let's make an issue and fix this right away.

regolith/builders/publistbuilder.py
```diff
@@ -76,6 +76,26 @@ def latex(self):
...
+                    doi = value
+                    filestub = filestub + "_doi_{}".format(doi)
+                    qualifiers = qualifiers + " from facility {}".format(doi)
+                else:
+                    facilities, doi = None, None
```
sbillinge, line 93: we can probably turn 5 lines of code into one line of code if we set this at the top.  Fewer lines means less to maintain.

sbillinge (commented):
please see comments.

regolith/builders/publistbuilder.py
davidkim1: For sure. Just to make sure I understand, you mean about the key == doi part?

regolith/builders/publistbuilder.py
davidkim1: will do!

regolith/builders/publistbuilder.py
davidkim1: Yes Professor. I can fix that now as well as the string. Not sure if you would like me to delete the elif key=="doi" part as well.

sbillinge:
Yes, no mention of doi in the whole code.  It means changing the test

sbillinge:
Btw, sorry about that... Didn't think things through enough

regolith/builders/publistbuilder.py
```diff
@@ -74,8 +75,15 @@ def latex(self):
...
             for g in grants:
                 cat_grants = cat_grants + "_" + g
-            filestub = filestub + "{}".format(cat_grants)
-            qualifiers = qualifiers + " from Grant{} {}".format(pl, text_grants)
+            filestub = f"{filestub}{cat_grants}"
+            qualifiers = f"{qualifiers} from Grant{pl} {text_grants}"
+        if self.rc.kwargs:
+            kw = True
+            key, value = self.rc.kwargs[0].split(':', 1)
+            if key == "facilities":
```
sbillinge, line 83: this would be a better place for the `facilities = None` to make the code more readable.  Just above the if `key == 'facilities'`

sbillinge (commented):
There might be something wrong here as these should be failing tests.  The test indicates that there is a facility of tbd in the meurernomonth entry, but there isn't so the tests should fail.  I am not sure what is happening, but I for sure can't merge it like this.  It is important to get the tests to fail before writing to code to make them pass.  It might be easier if I dig into this and try and fix it if you want to move on to the  next issue.

sbillinge:
@davidkim1 I managed to get this merged finally.  Thanks for the hard work getting it started.  It ended up being a bit complicated to get the testing right, so it was ok that I finished it.

Looking forward to your next PR.....

davidkim1:
@sbillinge I see that, Professor. Thank you for the guidance throughout anyway! I've learned a lot from this issue. Looking forward to submitting a successful PR soon!"

PR #1052, opened 2023-11-07
F todo UUID update test

davidkim1:
Added a test for f_todo when uuid is passed. I did add a flag as you suggested.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -575,7 +575,7 @@ helper_map = [
     (["helper", "f_todo", "--index", "99100"],
      "WARNING: indices >= 9900 are used for milestones which should be finished using u_milestone and not f_todo\n"
      ),
-    (["helper", "f_todo", "--index", "1saefa"],
+    (["helper", "f_todo", "1saefa"],
      "The task \"(1saefa) test a_todo\" in test for sbillinge has been marked as finished.\n"
      ),
     (["helper", "u_todo", "--index", "3", "--assigned-to", "sbillinge",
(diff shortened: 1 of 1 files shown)


tests/test_helpers.py
```diff
@@ -575,6 +575,9 @@
     (["helper", "f_todo", "--index", "99100"],
```
sbillinge, line 575: the index and uuid should be handled in the same way as each other as we discussed, so we probably either need to change how index is handled (remove optional argument --index and take is a required argument) or have the uuid taken in --index.  I am not sure there is a way to test this unless we run the tests again from scratch.   Maybe it is time to set that up.

Given that choice I would be inclined to pass the uuid as index. Maybe we change that variable name to --index_uuid or something like that?  We will have to comment out the test.

sbillinge (commented):
please see inline comment

tests/test_helpers.py
davidkim1: I would be inclined to change variable name to index_uuid or adding a variable name uuid. Because it would be clearer to the user what to enter. If we change it to index_uuid it is not as clear because one can think it means entering the index and uuid. If we leave it as just index, I will change the definition of described in index to add that it is interchangeable with the first 6 characters of a todo's uuid.

sbillinge: My idea is that the user can enter either an index or a uuid.  Either would work in exactly the same way.  If I enter `regolith helper f_todo --index 123` or `regolith helper f_todo --index 8yt4q` will work the same way if todo with index 123 has uuid 8yt4q.

Let's just keep the arg name as index for now.  Late we can easily change it if we find a better name.

PR #1054, opened 2024-01-24
moving gooey from conda-forge to pip installation and updating gh wor…

sbillinge:
…kflow accordingly


sbillinge:
fixed by #1056

PR #1084, opened 2024-07-04
Fix import not used and E501 issues 

stevenhua0320:
I leave E722 for check, which is `do not leave bare except`. However, the code structure is like this:
截屏2024-07-05 上午1 16 34

### DIFF
--- a/regolith/tools.py
+++ b/regolith/tools.py
@@ -569,7 +569,7 @@ def filter_presentations(
       The id of the person you will build the list for
     types: list of strings.  Optional, default = all
       The types to filter for.  Allowed types are
-      {*PRESENTATION_TYPES,}
+      {*PRESENTATION_TYPES, }
     since: date.  Optional, default is None
         The begin date to filter from
     before: date. Optional, default is None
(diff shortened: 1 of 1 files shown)


sbillinge:
For this one, please can you try and run tests and make sure they pass?  Let me know here on the PR and when tests pass I can merge.

for the E722, I am not sure how that would fail and raise an exception.  I will try and play with it a bit.

sbillinge:
@stevenhua0320 I think we can just remove that try except clause, so it becomes
```
            if index:
                task["preamble"] = (
                    f"({task.get('importance')})({task.get('days_to_due')} days): ({task.get('running_index', 0)}) "
                )
            else:
                task["preamble"] = ""
```

stevenhua0320:
> @stevenhua0320 I think we can just remove that try except clause, so it becomes
> 
> ```
>             if index:
>                 task["preamble"] = (
>                     f"({task.get('importance')})({task.get('days_to_due')} days): ({task.get('running_index', 0)}) "
>                 )
>             else:
>                 task["preamble"] = ""
> ```

yeah, I think the code here should be an if statement, then it could pass. Let me try it and if it works, I would push it.

sbillinge:
the if statement was already there.  I just removed the try/except

sbillinge:
@stevenhua0320 please confirm that this is passing tests and I can merge it.

stevenhua0320:
Yes, I have verified that it is a passing test.

PR #1091, opened 2024-07-05
Fix E501 F541 F841 for main, mongoclient, conftest, schemas.py

stevenhua0320:


regolith/schemas.py
```diff
@@ -157,14 +157,6 @@ def load_exemplars():
     return exemplars
```
sbillinge, line 157: since we are cleaning this function, let's 

1. remove the `exemplars = raw_exemplars`
2. change
```
with open(exemplar_file, "r", encoding="utf-8") as exemplar_file:
        raw_exemplars = json.load(exemplar_file)
```
to
```
with open(exemplar_file, "r", encoding="utf-8") as efile:
        exemplars = json.load(efile)
return exemplars
```

sbillinge (commented):
please see inline comment

PR #1095, opened 2024-07-08
Fix E501 E266 E712 F821 in test_validate, test_validators, runcontrol, test_commands.py

stevenhua0320:


tests/test_commands.py
```diff
@@ -38,9 +38,6 @@ def test_mongo_to_fs_python(make_mongo_to_fs_backup_db, make_mongodb):
         main(["mongo-to-fs"])
     except Exception as e:
         print(e)
-        assert True == False
-    else:
-        assert True == True
```
sbillinge, line 43: I think by deleting these lines you are changing behavior so it is probably not a good idea to change it.

If you simply chage the first to `assert False` and the second to `assert True`, does that remove the flake8 error?  That would then not change the behavior.

Also, same for below.

sbillinge (commented):
please see inline comment.

PR #1099, opened 2024-07-09
Fix E731 error for sorters.py

stevenhua0320:


regolith/sorters.py
```diff
@@ -4,12 +4,89 @@
...
-category_val = lambda x: x.get("category", "<uncategorized>")
-level_val = lambda x: x.get("level", "<no-level>")
-id_key = lambda x: x.get("_id", "")
+
+def doc_date_key(x):
+    """
+    Convert a Datetime object of the doc to a float key.
+    Parameters
+    ----------
+    x
```
sbillinge, line 13: let's give this a more intuitive name like `date`

then in the numpy-style docstring it would look like
```
date:  dict
  the document that is expected to contain date-like information in "year" and/or "month"

```

regolith/sorters.py
```diff
@@ -4,12 +4,89 @@
...
-doc_date_key_high = lambda x: date_to_float(x.get("year", 1970), x.get("month", "dec"))
-ene_date_key = lambda x: date_to_float(x.get("end_year", 4242), x.get("end_month", "dec"))
-category_val = lambda x: x.get("category", "<uncategorized>")
-level_val = lambda x: x.get("level", "<no-level>")
-id_key = lambda x: x.get("_id", "")
+
+def doc_date_key(x):
+    """
+    Convert a Datetime object of the doc to a float key.
+    Parameters
```
sbillinge, line 11: we need a blank line here in all the docstrings, and the first line inline, so something like:
```
"""Convert a Datetime object of the doc to a float key.

Parameters
------------
```

regolith/sorters.py
```diff
@@ -4,12 +4,89 @@
...
+    ----------
+    x
+
+    Returns
+    -------
+    float representing the date key.
```
sbillinge, line 17: I think there is no key here,  maybe more descriptive would be `the float serialization of the date information in the document`

sbillinge (commented):
please see my inline comments and try and update all the functions using this as a template.

sbillinge (commented):
sorry for this, but please see my edited comment.  I read the code more carefully and I realize that the thing I called "date" is not a date but is a document, so let's call it a document.  My suggested docstring is in the edited comment.

regolith/sorters.py
```diff
@@ -4,12 +4,95 @@
...
+    return date_to_float(date.get("year", 1970), date.get("month", "dec"))
+
+
+def ene_date_key(date):
+    """Convert a dict of ene Datetime object to float serialization of date info.
+
+    Parameters
+    ----------
+    date: dict
+        the document that is expected to contain date-like information in "year" and/or "month"
```
sbillinge, line 44: this should be `"end_year"` and `"end_month"`

regolith/sorters.py
```diff
@@ -4,12 +4,95 @@
...
+    """
+    return date_to_float(date.get("end_year", 4242), date.get("end_month", "dec"))
+
+
+def category_val(category):
+    """Convert the category of a dict of object into serialization of string of category info.
+
+    Parameters
+    ----------
+    category : dict
```
sbillinge, line 58: I think the input should again be called document here.

regolith/sorters.py
```diff
@@ -4,12 +4,95 @@
...
+        A dict of all corresponding categories for objects
+
+    Returns
+    -------
+    serialization of string representing the category.
```
sbillinge, line 63: this is not a serialization, it is just returning the category item.  We probably don't need this function in fact.  Like elsewhere, we could just put the lambda inside the sorter, unless it is used in many places.  For now, just make changes to the docstring and let's go with that.

sbillinge (commented):
please see additional comments inline.

PR #1100, opened 2024-07-09
Fix E741 E722 for gradebuilder, test_broker, storage.py

stevenhua0320:


regolith/builders/gradebuilder.py
```diff
@@ -43,22 +43,23 @@ def render(self, tname, fname, **kwargs):
...
             if not course.get("active", True):
                 continue
             course_id = course["_id"]
             stats = self.makestats(course)
             asgn = filter((lambda x: course_id in x["courses"]), self.gtx["assignments"])
-            catfunc = lambda x: x["category"]
+
+            def catfunc(x):
+                return x["category"]
             asgn = sorted(asgn, key=catfunc)
```
sbillinge, line 63: I think this would be cleaner if you just move the lambda to this line rather than define it as a function that is not used elsewhere.  So
```
asgn = sorted(asgn, key=lambda x: x["category"])
```
and delete line 61.  Something like that should work, is cleaner and should pass ci.

sbillinge (commented):
one quick inline comment here.

PR #1102, opened 2024-07-09
Fix bare except, trailing space, import not used, lambda function def error for sorters, l_currentappointmentshelper, l_projectahelper, makeappointmentshelper, l_todohelper, runcontrol, publistbuilder, formalletterbuilder.py

stevenhua0320:
Also PR #1087 is ready for review.

### DIFF
--- a/regolith/helpers/l_currentappointmentshelper.py
+++ b/regolith/helpers/l_currentappointmentshelper.py
@@ -76,16 +76,16 @@ class CurrentAppointmentsListerHelper(SoutHelperBase):
         for person in people:
             p_appt = person.get("appointments", None)
             if p_appt:
-                for k, (_id, appt) in enumerate(p_appt.items()):
+                for _id, appt in p_appt.items():
                     grantid = appt.get("grant")
                     if not grantid:
-                        print("No grant found in {} appt {}".format(person.get("_id"), k))
+                        print(f"No grant found in {person.get("_id")} appt {person.get("appt")}")
                     grant = fuzzy_retrieval(grants, ["name", "_id", "alias"], grantid)
                     if not grant:
-                        print("No grant found for {}".format(grantid))
+                        print(f"No grant found for {grantid}")
                     try:
                         accountnr = grant.get("account", grant["alias"])
-                    except NotImplemented:
+                    except Exception:
                         accountnr = ""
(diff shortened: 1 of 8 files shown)


regolith/helpers/l_currentappointmentshelper.py
```diff
@@ -76,7 +76,7 @@ def sout(self):
         for person in people:
             p_appt = person.get("appointments", None)
             if p_appt:
-                for _id, appt in p_appt.items():
+                for k, (_id, appt) in enumerate(p_appt.items()):
```
sbillinge, line 79: I don't think this is the right fix.  That k should be returning the key of the appointment, something like `appt.get("_id")` but it could be something else.  We don't want just an index there.

While we work to fix this, please can you also replace with f-strings.

regolith/helpers/l_currentappointmentshelper.py
```diff
@@ -85,7 +85,7 @@ def sout(self):
                         print("No grant found for {}".format(grantid))
                     try:
                         accountnr = grant.get("account", grant["alias"])
-                    except:
+                    except NotImplemented:
```
sbillinge, line 88: This will be a NotImplemented error.  For now maybe just put `Exception` though we would actually like to know how it fails.

regolith/helpers/l_projectahelper.py
```diff
@@ -196,7 +196,7 @@ def sout(self):
                     elif not isinstance(projectum.get("end_date"), dt.date):
                         try:
                             dt.datetime(*[int(num) for num in projectum.get("end_date").split("-")])
-                        except:
+                        except ValueError:
```
sbillinge, line 199: This should also probably be `Exception` now too. I am not sure how it fails.

regolith/helpers/makeappointmentshelper.py
```diff
@@ -147,7 +145,7 @@ def construct_global_ctx(self):
         try:
             if not rc.database:
                 rc.database = rc.databases[0]["name"]
-        except:
+        except NotImplemented:
```
sbillinge, line 148: `Exception`

sbillinge (commented):
please see comments

sbillinge:
yippee

PR #1103, opened 2024-07-10
try to fix tools, test_tools.py with test_awards_grants_honors

stevenhua0320:


tests/test_tools.py
```diff
@@ -1748,8 +1748,46 @@ def test_is_fully_appointed(appts, start, end, expected):
...
+                "honors": [{"name": "CVPR best paper award", "year": 2024, "month": 5}]
+            },
+            [
+                {"description": "CVPR best paper award", "year": 2024, "_key": 2024.05},
+                {"description": "Omega Laser User's Group Travel Award (\\$1,100)", "year": 2013, "_key": 2013.0},
+            ],
+        ),
+    ],
+)
+def test_honors_grants_honors_t2(input, expected):
```
sbillinge, line 1789: rather than write a different test function for this, can we do it by varying the input?

I think the best struture for this would be to define an input that contains funding, honors and service etc., and pass it to the function but then give as input in the parameterize the allowed tag (`service`,`honors`, etc.) and have it return the right thing.

sbillinge (commented):
please see my comment.

sbillinge:
closed by #1105

PR #1114, opened 2024-07-16
build regolith in cookiecutter workflow, passed local CI test, add news in regolith_cookiecut.rst

stevenhua0320:


sbillinge (commented):
I think you used the wrong org name in the cookie cutter.  I suggest we close this and redo it.....  Please see inline comments.

stevenhua0320:
Yes, I think redo it from scratch is a good idea. The main difficulty in the cookiecut is that different from the diffpy packages, the original repo has the regolith dir name so I need to make edits of the name of the directory after the migration, and also the import modules dir I need to change... I would pay attention to these comments when I redo the cookiecut. Thanks prof!

PR #1115, opened 2024-07-17
Cookiecutter for regolith

stevenhua0320:


sbillinge (commented):
let's break this up into smaller PRs.  The first one can be to simply move the `regolith` directory to `src/regolith`.  This affects a lot of files.

I will create a repo called `cookie` and I can merge changes into that.

Let's close this PR again and go again.

PR #1116, opened 2024-07-17
change dir from regolith/ to src/regolith/

stevenhua0320:


sbillinge (commented):
you seem to have edited files so the imports are all broken.  Don't use refactor in pycharm or vscode to do this.  Use the command line and the `mv` command.

At the end, none of the files should be modified, but just moved.

Please close this PR and let's try again.

PR #1122, opened 2024-07-17
Untrack files

stevenhua0320:
Merge files in requirement dir and other rst files


AUTHORS.rst
```diff
@@ -0,0 +1,10 @@
+Authors
+=======
+
+Billinge Group and community contibutors.
```
sbillinge, line 4: Please look in the authors of the original and add all the names that are there on the line above, and then this line

LICENSE.rst
```diff
@@ -0,0 +1,30 @@
+BSD 3-Clause License
```
sbillinge, line 1: We will use the copyright from this file, `Copyright (c) 2024, The Trustees of Columbia University
in the City of New York.` but not the rest, so I suggest to copy this copyright statement nad put it in the other license file (don't delete anything in that file), then delete this whole file.

requirements/README.txt
```diff
@@ -0,0 +1,11 @@
+# YOU MAY DELETE THIS FILE AFTER SETTING UP DEPENDENCIES!
```
sbillinge, line 1: please delete this file.

sbillinge (commented):
please see comments

PR #1123, opened 2024-07-17
requirement, doc, and tests dir

stevenhua0320:
Merge requirement, doc, and tests dir


doc/source/license.rst
```diff
@@ -0,0 +1,39 @@
+:tocdepth: -1
```
sbillinge, line 1: THis file needs to be changed by copy-pasting the license text from the other license file in here to replace the license text here.

sbillinge (commented):
please see comments inline

PR #1126, opened 2024-07-18
move tests dir to src/regolith/tests dir

stevenhua0320:
Safe Merge these test files
Cover the old conftest.py to new one since old one is for regolith while new one is for diffpy functions


sbillinge:
closing.  Replaced  by #1130

PR #1127, opened 2024-07-18
Change docs dir to doc dir

stevenhua0320:
Merge these doc files


sbillinge:
@stevenhua0320 sorry about this, but please can you redo this on a new PR without the reverted commits.  WHen you make a commit and then revert it, it stores in the git database the full commit and then the un-commit, i.e., It stores all the changes to all the files, and then it stores the unchanges.  This is ok when the commit is small (just a few lines changed) but when the commit is big, like moving hundreds of files, it is not a good idea.

The work you did in the last two commits is fine and I can merge that.

stevenhua0320:
> @stevenhua0320 sorry about this, but please can you redo this on a new PR without the reverted commits. WHen you make a commit and then revert it, it stores in the git database the full commit and then the un-commit, i.e., It stores all the changes to all the files, and then it stores the unchanges. This is ok when the commit is small (just a few lines changed) but when the commit is big, like moving hundreds of files, it is not a good idea.
> 
> The work you did in the last two commits is fine and I can merge that.

OK, I would redo these right now. I would close this PR.

PR #1132, opened 2024-07-18
copy docs dir files to doc dir

stevenhua0320:
Merge doc dir files with no change in content
use old version of conf.py and index.rst to cover cookiecut generated conf.py and index.rst and set to doc dir


sbillinge:
@stevenhua0320 nice work!  Very exciting.  pre-commit and all tests are passing....

What is left to do now?

PR #1135, opened 2024-07-18
api files and doc source files pre-commit fix

stevenhua0320:
Merge these api files and doc source dir pre-commit check files


sbillinge:
I change this PR so it is into a different branch called `api`.    

Do you have an idea about how we can test this?  Can you build this and then compare the api docs side-by-side by with the online ones?

stevenhua0320:
> I change this PR so it is into a different branch called `api`.
> 
> Do you have an idea about how we can test this? Can you build this and then compare the api docs side-by-side by with the online ones?

Is that mean we need to check whether these built api contains all the info that currently the online ones have?

sbillinge:
> > I change this PR so it is into a different branch called `api`.
> > Do you have an idea about how we can test this? Can you build this and then compare the api docs side-by-side by with the online ones?
> 
> Is that mean we need to check whether these built api contains all the info that currently the online ones have?

yes, that's right.  It can be in a different order, it doesn't have to be identical, and there can be additional information (which would be good), but just check that none of the current information is lost.

PR #1138, opened 2024-07-19
README.rst and conf.py update

stevenhua0320:
Merge README.rst and conf.py (safe merge to api)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,120 @@
-Regolith
+|Icon| |title|_
+===============
+
+.. |title| replace:: regolith
+.. _title: https://regro.github.io/regolith
+
+.. |Icon| image:: https://avatars.githubusercontent.com/regro
+        :target: https://regro.github.io/regolith
+        :height: 100px
+
+|PyPi| |Forge| |PythonVersion| |PR|
+
+|CI| |Codecov| |Black| |Tracking|
+
+.. |Black| image:: https://img.shields.io/badge/code_style-black-black
+        :target: https://github.com/psf/black
+
+.. |CI| image:: https://github.com/regro/regolith/actions/workflows/main.yml/badge.svg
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -1,8 +1,120 @@
...
+
+Before contribuing, please read our `Code of Conduct <https://github.com/regro/regolith/blob/main/CODE_OF_CONDUCT.rst>`_.
+
+Contact
+-------
 
-Docs:
------
-https://regro.github.io/regolith-docs/
+For more information on regolith please visit the project `web-page <https://regro.github.io/>`_ or email Prof. Simon Billinge at sb2896@columbia.edu.
```
sbillinge, line 120: I think this link needs to be fixed.  It is going to the wrong place.

doc/source/conf.py
```diff
@@ -19,11 +19,16 @@
 
 import cloud_sptheme as csp
 
+import regolith
 from regolith import __version__ as REGOLITH_VERSION
 from regolith.fsclient import _id_key, dump_json, json_to_yaml
 from regolith.main import CONNECTED_COMMANDS, DISCONNECTED_COMMANDS
 from regolith.schemas import EXEMPLARS, SCHEMAS
 
+autodoc_mock_imports = [
```
sbillinge, line 28: please can you double check that this is needed?

sbillinge (commented):
please see inline comments.

doc/source/conf.py
stevenhua0320: Yes, otherwise it would run into issues like e.g. WARNING: autodoc: failed to import module '...' from module 'regolith'; the following exception was raised: No module named 'regolith'. Something like that. It is the solution that resolves this issue that gives in the api workflow.

PR #1139, opened 2024-07-19
add doc/source files, change regolith.rst to index.rst in doc/source/api

stevenhua0320:
Merge these files. 
Note that Merge PR #1138 first to make sure it would successfully work


doc/source/collections/contacts.rst
```diff
@@ -1,18 +1,24 @@
 Contacts
 ========
-a lighter version of people.  Fewer required fieldsfor capturing people who are less tightly coupled
+a lighter version of people.  Fewer required fields for capturing people who are less tightly coupled
```
sbillinge, line 3: I see a few typo's being fixed but presumably these typos were being made on the auto-build and will be overwritten next time the api workflow is run.   Did you also make edits in the docstrings of the relevant module to also correct the mistake there?

sbillinge (commented):
this is really great, fantastic to see.  Please can you check if there were any manual edits for typos that all of them are actually corrected in the original docstring and not just here?  Otherwise they will be overwritten the next time the api workflow is run.

Maybe put those corrections on a separate PR

doc/source/collections/contacts.rst
stevenhua0320: I have verified that the original typo comes from schemas.json, and the original typo has been fixed in previous lint check PR. I'm sure that in current version the api workflow follow the edited version and corrected the typo automatically here.

PR #1144, opened 2024-07-20
add cloud_sptheme to docs.txt

stevenhua0320:
Safe merge this file


sbillinge:
this is the wrong fix. we are moving away from cloud_sptheme.  fixed by #1143

PR #1159, opened 2024-07-21
edits for citation files

stevenhua0320:
Have made edits on those appropriate places
but might need edits in tests/some files to make it pass CI?


news/citation.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* article synopsis and professional summary added to grant-report citation list
+* acknowledgement statement added to grant-report citation list
+**Changed:**
```
sbillinge, line 5: not sure, but I think we need a blank line here

sbillinge (commented):
super, thanks!

Yes, this is why it was not merged before, because it needs tests.

PR #1160, opened 2024-07-24
some cleaning of init and rever

sbillinge:


sbillinge:
taken care of by cookiecutting

PR #1163, opened 2024-07-26
deprecation remove

stevenhua0320:
Safe Merge these files

### DIFF
--- a/src/regolith/builders/basebuilder.py
+++ b/src/regolith/builders/basebuilder.py
@@ -5,7 +5,7 @@ from glob import glob
 from itertools import groupby
 
 from jinja2 import Environment, FileSystemLoader
-from xonsh.lib import subprocess
+from xonsh.api import subprocess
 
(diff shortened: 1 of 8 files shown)


sbillinge:
I checked the CI output and it looks as if we may be missing this deprecation fix in `basebuilder.py`: `$ conda install -c conda-forge mongodb/usr/share/miniconda/envs/test/lib/python3.12/site-packages/regolith/builders/basebuilder.py:8: DeprecationWarning: Use `xonsh.api.subprocess` instead of `xonsh.lib.subprocess`.`

Also there is a problem with the CI script that it is not loading `mongod` and running the mongo client tests.  Please can you make an issue for that and we can fix it on a separate PR.

There are a bunch of other warning too, of issues with the exemplar datase (things not found).  Let's also make an issue to clean that up, but it is low priority.  Maybe a new person can do that under your guidance.

PR #1166, opened 2024-10-24
Cookierelease

tinatn29:
add and commit untracked files


requirements/README.txt
```diff
@@ -0,0 +1,10 @@
+# YOU MAY DELETE THIS FILE AFTER SETTING UP DEPENDENCIES!
```
sbillinge, line 1: this file should be deleted

sbillinge (commented):
looks good. There is one file that needs deleting. ALso, we need to merge the requirements files.   I think run.txt -> conda.txt, but we also may need a way to handle the pip.txt here too.  This will be a bit more complicated than usual because we were installing gooey from pip and the rest from conda-forge....this is easier to explain in person

PR #1169, opened 2024-10-25
Cookierelease 3

tinatn29:
more modified files (old vs cookiecutter versions)


doc/make.bat
tinatn29, line 1: These two are the same (with a line spacing difference)

src/regolith/__init__.py
```diff
@@ -1,14 +1,24 @@
...
+# See LICENSE.rst for license information.
+#
+##############################################################################
 
 from xonsh.main import setup
 
+# package version
 from regolith.version import __version__
 
-"""A Research group database management system"""
```
sbillinge, line 9: I believe we still want this

doc/source/license.rst
```diff
@@ -7,6 +7,11 @@ License
 
 OPEN SOURCE LICENSE AGREEMENT
 =============================
+BSD 3-Clause License
```
sbillinge, line 10: Please check this, I think this has a CC license. We are not allowed to unilaterally change the license.

sbillinge (commented):
Please see comment s

doc/source/license.rst
tinatn29: OK I found the text buried in LICENSE.rst. I'll change this line to "Creative Commons License"

tinatn29: do you know if it's the usual CC BY 4.0? I can't find this in the existing doc/source/license.rst or LICENSE.rst

tinatn29:
I've edited license.rst and __init__.py per your comments

doc/source/license.rst
sbillinge: just reproduce here the license text from the LICENSE.rst   It predates CC BY 4.0 I think

doc/source/license.rst
```diff
@@ -7,6 +7,11 @@ License
 
 OPEN SOURCE LICENSE AGREEMENT
 =============================
+Creative Commons license (CC BY 4.0)
```
sbillinge, line 10: please see my comment.  I think this needs to be the exact form of the license from LICENSE.rst.

sbillinge (commented):
please see comment

doc/source/license.rst
tinatn29: OK -- I looked more carefully and it's CC0 license (meaning no copyright?)

doc/source/license.rst
```diff
@@ -7,11 +7,7 @@ License
 
 OPEN SOURCE LICENSE AGREEMENT
 =============================
-Creative Commons license (CC BY 4.0)
-
-Copyright (c) 2024, The Trustees of Columbia University in
-the City of New York.
-All Rights Reserved.
+Creative Commons Zero v1.0 Universal license (CC0 1.0)
```
tinatn29, line 10: OK -- I looked more carefully in LICENSE.rst and it seems to be CC0 license (public domain, copyright waived?), so edited it and removed the copyright lines also... Let me know if this is okay.

Currently in LICENSE.rst, there's a legal explanation of CC0 but not specific text that says "this package is under CC0..." --> do we add anything there?

At the bottom of LICENSE.rst there's also a line that says "NOTE: ace and jsoneditor in the regolith/static/ dir are covered under their own licences."
^What are these licenses and do we need to add this line to doc/source/license.rst also?

sbillinge: I think that just leaving the text as is there is ok.

Also for the CC0 I think just leaving it as it is is fine.  I think it is understood that the license in license.rst is the license that applies to the package.    If you want, you could look around in other packages.  It also probably doesn't do harm to add "this package is covered by....".

sbillinge (commented):
please see inline.  I will merge this, but you will need those comments for your other license pr.....

PR #1174, opened 2024-11-06
cookierelease: add license type to first line

tinatn29:


tinatn29:
STILL A DRAFT! Will let you know when I'm done here. I got somewhat confused by the versions -- will restore the old version since we have CC0 not the cookiecutter default license

PR #1176, opened 2024-11-06
Remove gitattributes, rever, run_test files

bobleesj:
Files no longer needed using our current cookiecutting standard


sbillinge:
@bobleesj I am moving this cookie work to have a base branch of cookie to avoid disasters as usual.  I let Tina know but she may need help to get the new workflow set up.  I will merge this, but into cookie, so please checkout that branch and build new branches from there, as usual.  We will merge to main when tests are passing.

bobleesj:
@sbillinge okay, will do my best to prevent that disaster

PR #1203, opened 2024-11-26
Configure non-reusable Python 3.11 Test on PR GitHub CI

bobleesj:
Pytest CI passes:

Screenshot 2024-11-26 at 3 26 20 PM

This successfully builds wxPython from sdist like in labpdfproc. Hence, it will take 28 mins to build.

### DIFF
--- a/.github/workflows/tests-on-pr.yml
+++ b/.github/workflows/tests-on-pr.yml
@@ -16,7 +16,7 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - name: Check out diffpy.labpdfproc repository
+      - name: Check out regolith repository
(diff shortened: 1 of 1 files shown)


bobleesj:
Here, `Python 3.11` is configured.

```
        activate-environment: test
          auto-update-conda: true
          environment-file: environment.yml
          auto-activate-base: false
          python-version: 3.11
```

Once @tinatn29 resolves all warnings and skipped tests with Python 3.11, we could later update our CI Python env to 3.12, then 3.13.

Like in `labpdfproc`,  `wxpython` needs to be  built using `sdist` for Linux, hence taking about 30 minutes..

PR #1205, opened 2024-11-27
change position values in grants to remove warnings in tests

sbillinge:
remove some warnings and at the same time test the ci


bobleesj:
I think gooey supports up to 3.10 via conda-forge..

https://anaconda.org/conda-forge/gooey/files

PR #1207, opened 2024-12-11
fix matplotlib plot_date warnings

### LINKED ISSUE #1201: Fix pytest 40 warnings from matplotlib plot_date
### Problem
When running `pytest` on a Python 3.11 environment, the tests are passing but with 57 warnings (and skips). 40/57 warnings are matplotlib deprecation warnings for `plot_date`

### Proposed solution
Replace `plot_date` with `plot`

tinatn29:
fix #1201


tinatn29:
40/57 are fixed with this PR (just matplotlib) - ready to merge if this looks good. I'll create a separate issue for Bob so we can look at the rest

sbillinge:
> 40 (out of 57 warnings) are fixed with this PR (just matplotlib) - ready to merge if this looks good. I'll create a separate issue for Bob so we can look at the rest

haha, gotta love that!!!!!

PR #1211, opened 2024-12-12
try and suppress trailing whitespace after a line feed in dump

sbillinge:


sbillinge:
closing on the basis that this wasn't really the source of my problems and so maybe we don't need it, and we can revisith easily later anyway.

Issue #1218, opened 2024-12-13
Existence of two GitHub pages - deploy API docs


bobleesj:
### Problem

@sbillinge 

We have both  https://regro.github.io/regolith-docs/ and https://regro.github.io/regolith/ deployed. 

It appears they are similar. If so, should we keep one and update URLs in readme and everywhere.

sbillinge:
https://regro.github.io/regolith-docs/  was from our old builds.    https://regro.github.io/regolith/ is from our new builds.

If we can figure it out, it would be good to just use the new builds but make a line from the old name to the new name (in case anyone is still statically linking to the old name.)  Do we know how to do that/  If not, I don't actually think too many people are using this and we can just try and figure out how to delete the old page.

bobleesj:
> it would be good to just use the new builds but make a line from the old name to the new name (in case anyone is still statically linking to the old name)

I can't find an option for re-directiong domain via github pages below..

<img width="591" alt="Screenshot 2024-12-13 at 6 38 15 PM" src="https://github.com/user-attachments/assets/6846f504-1565-40a8-a3d9-3d51de8c4948" />

@sbillinge  I think we have a separate repo for `regolith-docs` - https://github.com/regro/regolith-docs I think the GitHub pages should be unpublished there if we decided to delete that domain.

@tinatn29 could you pls update the URL?

sbillinge:
Sounds good.  I don't really know how to resolve this.  In a way I would like to keep things standard as I know what to type to find the docs if it conforms to our group standard (the new one), so I would rather keep that URL if we can only have one.

bobleesj:
@sbillinge for confirmation, You'd rather have 'regolith-doc' in the URL? 

If that's the case, then we may have to sync the current regolith and regolith-doc repos synced side by aide. Not sure if this is what is desired

sbillinge:
I am ok deprecating the old url.

PR #1219, opened 2024-12-13
Build API documentation 

tinatn29:
built docs with `auto_api.py` per cookiecutting instructions (files in doc/source were modified) + fixed typos flagged by codespell. 

Besides fixing the typos, I didn't make any tweaks after `auto_api` did its thing, so let me know if any of the files modified should be different...

did this build one of the two docs (or both) that @bobleesj mentioned in #1218?


doc/source/collections/abstracts.rst
bobleesj, line 1: why is this file added?

doc/source/collections/projecta.rst
```diff
@@ -45,7 +45,7 @@ The following lists key names mapped to its type and meaning for each entry.
 	:notes: list, any notes about the kickoff, optional
 	:status: string, status of the kickoff. Allowed values are {', '.join(PROJECTUM_STATI)}, optional
 
-		Allowed values:
```
bobleesj, line 48: not sure the cause of these extra spaces provided throuhgout. I think this PR should only fix along `doc/source/api/...` right?

bobleesj (commented):
I had a preliminary look.  Please check my in-line comments. @tinatn29 

For this PR, since we are modifying a lot, I think it would be safer for us to just do api_doc only? Also since there is one commit, I can't really track what manual modifications have been made.

tinatn29:
will redo this PR to separate commits and make sure `auto_api` is doing the right thing

PR #1220, opened 2024-12-13
auto_api build doc

tinatn29:
ran auto_api per cookiecutting instructions


bobleesj:
Yeah, it's better! Thanks Tina.

PR #1221, opened 2024-12-13
add news item

### LINKED ISSUE #1214: news for the cookiecutting activity
### Problem
@sbillinge merged to main without news thinking there may already be a news there but isn't  @tinatn29 please could we have a news for the whole activity.  The wording is somewhere either in teh cookiecutting instructions, or maybe better, look in the changelog of one of the released packages.

tinatn29:
fix #1214. I added `cookie-news.rst` to the `news/` folder per cookiecutting instruction. Let me know if we need anything else here!

### DIFF
--- a/news/cookie-news.rst
+++ b/news/cookie-news.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Repo structure modified to the new diffpy standard
+
+**Security:**
+
+* <news item>
--- a/news/cookie-news.rst
+++ b/news/cookie-news.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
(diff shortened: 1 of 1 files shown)


news/cookie-news.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Repo structure modified to the new diffpy standard
```
bobleesj, line 19: Maybe need to have "cookiecutter" in the news? since we will be mostly use that term when we revisit?

tinatn29: "Repo structure modified to the new diffpy (cookiecutter) standard"? And we should update the instruction on here too? https://github.com/Billingegroup/cookiecutter/tree/main

bobleesj: Yes, if you could make an issue, i will address it

tinatn29: I created an issue in cookiecutter repo: https://github.com/Billingegroup/cookiecutter/issues/185

PR #1223, opened 2024-12-31
strip the prum finisher input before assigning to key

### LINKED ISSUE #878: strip() id's before testing if they are valid in listers and adders etc.

sbillinge:
closes #878


sbillinge:
closed by #1280

PR #1228, opened 2025-01-13
feat: cv and resume now differentiate current and past projects

### LINKED ISSUE #1227: fix: don't reproduce non-active projects in cv.
### Problem
All projects appear to be appearing in cv

### Proposed solution
better behavior would be that only active projects appear in the cv.  this  gives the user more control over what appears and what doesn't.

sbillinge:
closes #1227 

Chose different behavior.  For cv, show current and past projects, since cv is supposed to show the whole career, but separate them into current and past for easier interpretation by readers.


sbillinge:
@bobleesj do you know why the news check is failing?

bobleesj:
Screenshot 2025-01-13 at 3 37 13 PM

seems like TEMPLATE.rst has been modified 

@sbillinge

sbillinge:
ah, that could explain it. I will fix.

PR #1231, opened 2025-01-19
modify fsclient yaml dump in attempt to have it conform ot prettier standards

sbillinge:


sbillinge:
closing, need to think about this.

PR #1251, opened 2025-06-22

zmx27:


.github/workflows/tests-on-pr.yml
```diff
@@ -1,59 +1,15 @@
...
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
   workflow_dispatch:
 
 jobs:
-  validate:
```
zmx27, line 11: should this section be added back?

zmx27:
@sbillinge this is ready for review. Sorry about all the auto fixes that were applied by the pre-commit hooks...

sbillinge (commented):
please see my inline comments.  However, the main thing here is that I suggest we break this into a few different PRs. 

1. exclude all html/css/js/json files from prettier.  Make a PR.  This should have all the docformatter auto-edits on it and the other changes from the package updates.
2. Then let's work to get the tests passing.  Please can you share what errors you are getting?
3. then un-exclude json files (but not hte others) and make a PR.  Make sure the tests still pass
4. Finally, try a pr when we unexclude the html etc.  This last step will likely break all the tests as we test by making sure output html looks like some template.  We can then choose to ignore these if we can't get it to pass after linting

sbillinge:
Please could you do the "point 1" PR on a clean branch from main?  This one has everything on it, including the things we don't want to do yet.

For the pytest errors, most of those are because of changes in file structure.  We can work on that in step 2.  For now let's just do step one.  Don't forget to exclude the files from prettier I mentioned to exclude in step 1

PR #1253, opened 2025-06-22

zmx27:


.github/workflows/tests-on-pr.yml
```diff
@@ -1,58 +1,15 @@
...
-      - name: Install libgtk for Linux
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libgtk-3-dev
-
-      - name: Install regolith and requirements
-        run: |
-          conda install --file requirements/test.txt
-          conda install --file requirements/conda.txt
-          pip install gooey
```
zmx27, line 44: I think that the Github action is failing because it's not running pip install gooey when running the tests. This was something that I noticed it too when I was running it locally. How should I add this back?

.github/workflows/tests-on-pr.yml
```diff
@@ -1,58 +1,15 @@
...
-        with:
-          verbose: true
-          fail_ci_if_error: true
-          token: ${{ secrets.CODECOV_TOKEN }}
+  tests-on-pr:
+    uses: scikit-package/release-scripts/.github/workflows/_tests-on-pr.yml@v0
+    with:
+      project: regolith
+      c_extension: false
+      headless: true
```
sbillinge, line 13: I think this should be false, but not 100% sure.  I don't think there are any gui tests.

.github/workflows/tests-on-pr.yml
sbillinge: In principle we can try and fix this but I will aim to make a release of Gooey shortly that should fix this so I would rather not make a workaround fix.

This may come up in general in the future for people.  We could ask @bobleesj if it is possible to have an entry in the tests-on-pr that can insert a `run` statement into the GitHub action.  If we had that we would be able to insert the `pip install gooey` to run downstream.

.pre-commit-config.yaml
```diff
@@ -16,7 +16,6 @@ repos:
       - id: check-yaml
       - id: end-of-file-fixer
       - id: trailing-whitespace
-        exclude: '\.(rst|txt)$'
```
sbillinge, line 19: we will need this for it to pass tests I think.

pyproject.toml
```diff
@@ -1,28 +1,28 @@
...
+keywords = ['content management system']
 readme = "README.rst"
 requires-python = ">=3.11, <3.14"
 classifiers = [
         'Development Status :: 5 - Production/Stable',
         'Environment :: Console',
         'Intended Audience :: Developers',
         'Intended Audience :: Science/Research',
-        'License :: CC0 1.0 Universal (CC0 1.0) Public Domain Dedication',
+        'License :: OSI Approved :: BSD License',
```
sbillinge, line 25: In general, don't EVER change the license (unless there is an error, which there isn't here) Which means we want to bring the old license text here.

tests/conftest.py
```diff
@@ -1,4 +1,4 @@
-"""Copyright (c) 2017, Anthony Scopatz All rights reserved."""
+"""Copyright (c) 2017-2025, Anthony Scopatz All rights reserved."""
```
sbillinge, line 1: Here let's change the copyright to Simon Billinge from 2021-2025.  for 2017-2020 it can be Anthony

sbillinge (commented):
this is great.  I left a few comments.

zmx27:
@sbillinge ready for review

PR #1255, opened 2025-06-23
style: apply docformatter changes

zmx27:


sbillinge:
Thanks @zmx27 .  I will merge this so that we have all the docformatter changes.    I deleted the old migration branch on upstream and will merge this into a new migration branch so please:

1. delete your local migration branch if you didn't already.  If there is a migration branch on your fork, delete that too (there shouldn't be)
2. git fetch upstream
3. git checkout migration
4. create a new branch
5. try removing the `exclude: '\.(...yaml|yml)$'` from the "traliing whitespace" and seeing if you can get those tests to pass again but without that exclued.
6. put that on a new PR and share and we can see what happened.

Pkease build new branches from the base branch called migration.

Good progress.

PR #1256, opened 2025-06-23
ci: remove yaml files from being excluded in trailing whitespaces hook

zmx27:


zmx27:
> try removing the exclude: '\.(...yaml|yml)$' from the "traliing whitespace" and seeing if you can get those tests to pass again but without that exclued.
> put that on a new PR and share and we can see what happened.
@sbillinge Removing those files from the exclude section would trigger autofixes from pre-commit and prevent the tests that were initially passing from continuing to pass. How should I proceed?

zmx27:
@sbillinge ready for review after all tests in pytest now passes

PR #1258, opened 2025-06-24

zmx27:


zmx27:
@sbillinge this is ready for review. The ```tests-on-pr``` GitHub Action is still failing because ```pip install gooey``` is not run, but I have ensured that all the tests in pytest are still passing.

Issue #1210, opened 2024-12-12
Use re-usable workflow for GitHub CI once `gooey` package is available on conda-forge


bobleesj:
### Problem

In `test-on-pr` and matrix CI,  we `pip install gooey` as shown below:

```
      - name: Install regolith and requirements
        run: |
          conda install --file requirements/test.txt
          conda install --file requirements/conda.txt
          pip install gooey
          python -m pip install . --no-deps
```

### Proposed solution

Once `gooey` is available as a conda-package, we can add `gooey` back to `conda.txt` and then use the BG's reusable .yml script.

sbillinge:
closed by #1294

PR #1261, opened 2025-06-27
ci: add run command to tests-on-pr.yml to run pip install gooey

zmx27:
#1210 can be closed when the ```migration``` branch is merged into ```main``` as we are now able to use our reusable script from ```scikit-package``` while pip installing ```gooey``` in this specific repository so tests in the CI can pass.

### DIFF
--- a/.github/workflows/tests-on-pr.yml
+++ b/.github/workflows/tests-on-pr.yml
@@ -11,8 +11,6 @@ jobs:
       project: regolith
       c_extension: false
       headless: false
-      run: |
-        pip install gooey
-        echo "Done"
+      run: pip install gooey
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}


zmx27:
@sbillinge this is ready for review.

@bobleesj the run block works!

.github/workflows/tests-on-pr.yml
```diff
@@ -11,5 +11,8 @@ jobs:
       project: regolith
       c_extension: false
       headless: false
+      run: |
+        pip install gooey
+        echo "Done"
```
bobleesj, line 16: Maybe echo part isn't needed? When it is "Done", it will proceed to another block of code

bobleesj (commented):
Just a very minor comment. I had that "echo" for testing multiple lines of commands. But maybe it isnt needed here

.github/workflows/tests-on-pr.yml
zmx27: Haha oops

zmx27:
@sbillinge issue #1262 is created

PR #1268, opened 2025-10-18
feat: add release schema in regolith.

stevenhua0320:
@sbillinge Ready for review. Here we define the schema for `release_list` that will be built.


stevenhua0320:
Close this as not following the regolith PR standard, I would redo the PR for another one

PR #1269, opened 2025-10-18
feat: add releaselist schemas into schemas.json.

stevenhua0320:
@sbillinge Ready to review the potential schemas for releaselist. The Tests on PR fails now because I have not create the corresponding exemplars in `exemplars.json`. After we reach an agreement on the schemas, I would create the corresponding exemplar.

### DIFF
--- a/src/regolith/exemplars.json
+++ b/src/regolith/exemplars.json
@@ -2169,6 +2169,49 @@
             "year": 2020
         }
     ],
+    "releaseReports": [
+        {
+            "_id": "sbillinge",
+            "date": "17/07/2025",
+            "releases": [
+                {
+                "package_name": "regolith",
+                "description": "A Group Content Management System",
+                "url": "https://github.com/regro/regolith",
+                "release_updates": [
+                    {
+                        "version_number": "0.9.1",
+                        "updated_capacities": ["Rearrange Regolith package to new Billingegroup package standards",
+                            "change name and then fix tests for function that formats awards and honors"],
+                        "release_url": "https://github.com/regro/regolith/releases/tag/0.9.1",
+                        "release_date": "21/07/2024",
+                        "author": ["Simon Billinge", "Rundong Hua"],
+                        "author_id": ["sbillinge", "stevenhua0320"]
+                    },
+                    {
+                        "version_number": "0.8.1",
+                        "updated_capacities": "Fix bug in builder",
+                        "release_date": "11/13/2023",
+                        "author": "Simon Billinge",
+                        "author_id": "sbillinge"
+                    }]
+            },
+                {
+                    "package_name": "diffpy.xpdfsuite",
+                    "description": "The suite of algorithms and software for diffpy programs.",
+                    "release_updates": [
+                        {
+                            "version_number": "0.2.0",
+                            "updated_capacities": "scikit-pakcage everything in the suite to new BillingeGroup standard.",
+                            "release_date": "20/10/2025",
+                            "author": ["Simon Billinge", "Zhiming Xu", "Rundong Hua"],
+                            "author_id": ["sbillinge", "zmx27", "stevenhua0320"]
+                        }
+                    ]
+                }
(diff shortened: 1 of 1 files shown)


src/regolith/schemas.json
```diff
@@ -2924,6 +2924,109 @@
...
+            "type": "integer"
+        },
+        "released_package": {
+            "description": "The list of software releases.",
+            "required": true,
+            "type": "list",
+            "schema": {
+                "type": "dict",
+                "schema": {
+                    "package_name": {
```
sbillinge: isn't the package already known from the top level?

stevenhua0320:
> this schema is not quite right I think. Honestly, it would be easier if you wrote the exemplar before we write the schema, and we can the write the schema to capture the exemplar.

OK, I would first design the exemplar instead and then redo the schemas.

src/regolith/exemplars.json
```diff
@@ -2169,6 +2169,49 @@
             "year": 2020
         }
     ],
+    "releaseReports": [
+        {
+            "_id": "sbillinge",
+            "date": "17/07/2025",
```
sbillinge: dates are in iso standard.  Also, what date is this?, when the package was first created?  in which case `creation_date` or something like that.

src/regolith/exemplars.json
```diff
@@ -2169,6 +2169,49 @@
             "year": 2020
         }
     ],
+    "releaseReports": [
```
sbillinge: I think this collection should be called `software`

src/regolith/exemplars.json
```diff
@@ -2169,6 +2169,49 @@
...
+        {
+            "_id": "sbillinge",
```
sbillinge: I would expect the id to reflect the name of the software

src/regolith/exemplars.json
```diff
@@ -2169,6 +2169,49 @@
...
+            "date": "17/07/2025",
+            "releases": [
+                {
+                "package_name": "regolith",
+                "description": "A Group Content Management System",
```
sbillinge: top level

src/regolith/exemplars.json
```diff
@@ -2169,6 +2169,49 @@
...
+                "url": "https://github.com/regro/regolith",
+                "release_updates": [
+                    {
+                        "version_number": "0.9.1",
+                        "updated_capacities": ["Rearrange Regolith package to new Billingegroup package standards",
+                            "change name and then fix tests for function that formats awards and honors"],
+                        "release_url": "https://github.com/regro/regolith/releases/tag/0.9.1",
+                        "release_date": "21/07/2024",
+                        "author": ["Simon Billinge", "Rundong Hua"],
+                        "author_id": ["sbillinge", "stevenhua0320"]
```
sbillinge: not needed

src/regolith/exemplars.json
```diff
@@ -2169,6 +2169,49 @@
...
+                {
+                "package_name": "regolith",
+                "description": "A Group Content Management System",
+                "url": "https://github.com/regro/regolith",
+                "release_updates": [
+                    {
+                        "version_number": "0.9.1",
+                        "updated_capacities": ["Rearrange Regolith package to new Billingegroup package standards",
+                            "change name and then fix tests for function that formats awards and honors"],
+                        "release_url": "https://github.com/regro/regolith/releases/tag/0.9.1",
```
sbillinge: maybe keep this if we can use it get the release info.  In which case, don't save the other information here, for example the changelog.

Maybe have a `release_type` that can be `major`, `minor`, `patch`.  This will allow us to chosse what levels of release we want to list in some situation.

sbillinge (commented):
please see comments

src/regolith/exemplars.json
sbillinge: Yes, I forgot about that... We Should use that as a start

PR #1270, opened 2025-10-27
style: add software examples in exemplar.json

stevenhua0320:
@sbillinge Ready for the first-round review, after we agree on this I start to add features in `software` in `schemas.json`.

### DIFF
--- a/src/regolith/schemas.json
+++ b/src/regolith/schemas.json
@@ -2961,51 +2961,54 @@
             "required": true,
             "type": "list",
             "schema": {
-                "major": {
-                    "description": "The major release number of the software at the release.",
-                    "required": true,
-                    "type": "integer"
-                },
-                "minor": {
-                    "description": "The minor release number of the software at the release.",
-                    "required": true,
-                    "type": "integer"
-                },
-                "patch": {
-                    "description": "The patch number of the software at the release.",
-                    "required": true,
-                    "type": "integer"
-                },
-                "release_type": {
-                    "description": "The type of release: 'major', 'minor', 'patch', or 'pre-release'.",
-                    "required": true,
-                    "type": "string",
-                    "allowed": ["major", "minor", "patch", "pre-release"]
-                },
-                "pre_release": {
-                    "description": "The number of the release candidate when the release_type is 'pre-release'. Pattern follows major.minor.patch-rc.pre_release.",
-                    "required": false,
-                    "type": "integer"
-                },
-                "release_date": {
-                    "description": "The date of release of the software.",
-                    "required": true,
-                    "type": "date"
-                },
-                "summary": {
-                    "description": "The summary of main updates in the release.",
-                    "required": true,
-                    "type": "string"
-                },
-                "changes": {
-                    "description": "Brief list of changes in the release.",
-                    "required": false,
-                    "type": "list"
-                },
-                "release_id": {
-                    "description": "The id of the release in the GH page of the released software, which is used to build the url pointing to the release given the 'org_name' and 'repo_name'.",
-                    "required": true,
-                    "type": "string"
+                "type": "dict",
+                "schema": {
+                    "major": {
+                        "description": "The major release number of the software at the release.",
+                        "required": true,
+                        "type": "integer"
+                    },
+                    "minor": {
+                        "description": "The minor release number of the software at the release.",
+                        "required": true,
+                        "type": "integer"
+                    },
+                    "patch": {
+                        "description": "The patch number of the software at the release.",
+                        "required": true,
+                        "type": "integer"
+                    },
+                    "release_type": {
+                        "description": "The type of release: 'major', 'minor', 'patch', or 'pre-release'.",
+                        "required": true,
+                        "type": "string",
+                        "eallowed": "RELEASE_TYPES"
+                    },
+                    "pre_release": {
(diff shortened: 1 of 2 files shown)


src/regolith/exemplars.json
```diff
@@ -2175,6 +2175,36 @@
...
     },
+    "software": [
+        {
+            "_id": "diffpy.utils",
+            "groups": ["xrd"],
+            "active": true,
+            "url": "https://github.com/diffpy/diffpy.utils",
+            "grants": ["NSF Funding"],
+            "author": ["Simon Billinge", "Sanjoon Bob Lee", "Zhiming Xu", "Tieqiong Zhang"],
+            "version": [{"0.9.0": ["patch", "08/09/2025"], "0.8.0": ["minor", "06/08/2025"],
```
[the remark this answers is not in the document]
sbillinge: It is probably a good idea to have release_type to make things easier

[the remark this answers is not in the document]
stevenhua0320: Oh, I got your idea

src/regolith/exemplars.json
```diff
@@ -2175,6 +2175,81 @@
...
+                    "minor": 1,
+                    "patch": 0,
+                    "release_date": "2025-10-25",
+                    "summary": "Python 3.14 and something else",
+                    "change": ["deprecated python2 feature.",
+                                "changed to scikit-packaged standard.",
+                                "Add some functionality.",
+                                "Modify existing function to make convenient."
+                    ],
+                    "version_number": "1.0.0",
```
sbillinge: We don't need this because the version number can be built from major. Minor. Patch

src/regolith/exemplars.json
```diff
@@ -2175,6 +2175,81 @@
         "email": "haperson@uni.edu",
         "university_id": "HAP42"
     },
+    "software": [
+        {
+            "_id": "diffpy.utils",
+            "groups": ["xrd"],
+            "active": true,
+            "url": "https://github.com/diffpy/diffpy.utils",
```
sbillinge: Instead of this, I wonder if we want `org_name:` and `repo_name`, then bulk the url. Probably also `platform_name:` which would take GitHub. Then we can build the url ourselves

sbillinge (commented):
Please see comments. Nearly there!

src/regolith/exemplars.json
stevenhua0320: For the platform name, should we have a list datatype here as probably there are some would be released on pypi, conda-forge, etc.

stevenhua0320:
@sbillinge I have made the edits, other than the `release_id` for the type issue that we might have a discussion, it is ready for review.

sbillinge:
Ok, perfect now. Let's write the schema that matches this and then tests should pass.

src/regolith/schemas.json
```diff
@@ -2936,11 +2936,68 @@
...
+                    "required": true,
+                    "format": "date"
+                },
+                "summary": {
+                    "description": "The summary of the major and minor release of software.",
+                    "required": true,
+                    "type": "string"
+                },
+                "change": {
+                    "description": "The detailed change of the software in separate sentences.",
```
sbillinge: Brief list of changes

src/regolith/schemas.json
```diff
@@ -2936,11 +2936,68 @@
...
+                    "description": "The date of release of the software.",
+                    "required": true,
+                    "format": "date"
+                },
+                "summary": {
+                    "description": "The summary of the major and minor release of software.",
+                    "required": true,
+                    "type": "string"
+                },
+                "change": {
```
sbillinge: Changes

src/regolith/schemas.json
```diff
@@ -2962,37 +2962,48 @@
...
-                    "description": "The patch update of the software at the release.",
+                    "description": "The patch number of the software at the release.",
                     "required": true,
                     "type": "int"
                 },
+                "release_type": {
+                    "description": "The type of release: 'major', 'minor', 'patch', or 'pre-release'.",
+                    "required": true,
+                    "type": "string",
+                    "enum": ["major", "minor", "patch", "pre-release"]
```
sbillinge: please check elsewhere in schema, I think we are using a different syntax for this than enum.  We may want to switch to this as it is more standard json schema, but we are using a different schema validation system.  Please look elsewhere in the `schema.json` for how we are doing this.

sbillinge (commented):
one small comment then I can merge.  This is failing tests.  Please could you add the line to the github workflow that pins the version for 3.13 and also make an issue to remove that when we upate to 3.14.

sbillinge:
this is failing tests I am afraid.  Please can you test  locallhy before pushing.

stevenhua0320:
> this may be failing because we are not using json schema but another validation scheme. We handle the `enum` differently. I don't remember the syntax, but look through the schema and you can probably find it.

Yes, I realized that and I initially guessed it is `eallowed` one but I found that it is not correct. Let me keep finding the right one.

sbillinge:
perfect, I merged.   The next thing is to build what we want the releases list (like publication list) to look like. This would be a latex file that goes in the tests/outputs directory.  When we run it through latex we want it to look nice and have hte information we need.   The actual things that will appear in there will come from the exemplar.json, but we should make it look as we want it and then add into the examplar what we need rather than vice versa.

sbillinge:
here is what my publist looks like:
image

and here is what the preslist looks like:
image

PR #1272, opened 2025-11-01
style: add test output style.

stevenhua0320:
@sbillinge Ready to review, one question here is you said all the information in the output should come from the exemplar, but we don't have the short description (eg: for `diffpy.utils` the description is  `General utilities for analyzing diffraction data`). Should we add this into the exemplar and also make an edit on the schemas.json?

The output style looks like this:
Screenshot 2025-11-01 at 2 19 45 PM


sbillinge:
This looks great!  One thing is to have the URL printed explicitly as well as hyperlinked because sometimes it will be printed on paper.  I like the current layout, so maybe just add `url: <url>` indented right below the authors..

Then that is a wrap on this.  I can merge it so we have less to review as we build the builder.  Let's start the builder!

stevenhua0320:
Screenshot 2025-11-01 at 4 50 47 PM
Now the layout of releasereport looks like this.

sbillinge:
Great. Start by making an empty builder and doing the work in the commands that I showed you. This should then start failing tests. Then make the template and the builder. Use the preslist as the template but take inspiration from all over. Push frequently and I will give feedback

PR #1273, opened 2025-11-03
style: add builder for releaselist.

stevenhua0320:
@sbillinge Ready for initial review, I should further add `filter_software` similar style as `filter_presentation` in `tools.py` and add `btype` into `builder.py`. But for now, we should first confirm whether I am on the right track.


news/releasebuilder.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* Add releasebuilder functionality.
```
sbillinge: Releaselistbuilder

src/regolith/tools.py
```diff
@@ -711,6 +711,153 @@ def filter_presentations(
...
+
+    firstclean = list()
+    secondclean = list()
+    thirdclean = list()
+    fourthclean = list()
+    presclean = list()
+
+    # build the filtered collection
+    # only list the talk if the group member is an author
+    for sof in software:
```
sbillinge: `for program in software:` for greater readability

sbillinge (commented):
I am confused why this is passing tests.  Should it not be failing?

sbillinge:
Ok, maybe add a minimal template and see if we can get the test to fail

stevenhua0320:
@sbillinge Got it why it would always pass it. Is that because in the original CI test we tested the original ones that has no coverage on `ReleaseListBuilder`? If this is the case, then we need to write a test function to test the behavior of filter tool. But it might need some time to think about it to follow the original style in `test_tools.py`

stevenhua0320:
@sbillinge I noticed that we have a `test_builders.py` in regolith, should we add something to that so that it would trigger to fail?

stevenhua0320:
@sbillinge I think right now it goes fairly well here. The status right now is it now renders the latex file, but there is a mismatch between the built and the template. So, the next step is to fix those mismatch.

sbillinge:
yeh!

PR #1278, opened 2025-12-23
feat: extract multiple repos for given owner/orgs by running api command

stevenhua0320:
@sbillinge Ready for review, note that for `releases` field I did not test as for my own repo I do not have any release. We might test this and if there is any formatting issue in `software.yml` generated, I could make edits on that. The other fields I have tested locally and it should be right. 

To use this, we first have the Github personal access token export on local. Then use the command `regolith gh-extractor <owner> --all` to extract all repos or `regolith gh-extractor <owner> <package-name>` for individual package. Then it would generate the `software.yml` for the packages. After that we use the original `regolith build releaselist` command to build the releaselist report.

### DIFF
--- a/src/regolith/commands.py
+++ b/src/regolith/commands.py
@@ -7,7 +7,7 @@ import sys
 from copy import copy
 from pprint import pprint
 
-import yaml
+from ruamel.yaml import YAML
 
 from regolith import storage
 from regolith.builder import BUILDERS, builder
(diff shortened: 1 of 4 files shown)


src/regolith/commands.py
```diff
@@ -7,10 +7,13 @@
 from copy import copy
 from pprint import pprint
 
+import yaml
```
sbillinge: we like to use ruamel as a general matter

sbillinge:
thanks @stevenhua0320

PR #1285, opened 2026-07-30
feat: add talks and slides schemas and exemplars

sbillinge:
Add schemas and exemplars for the talks and slides collections to schemas.json and exemplars.json, and the SLIDE_TYPES allowed list to schemas.py.

A deck in slides holds an ordered list of slides on one topic. A talk in talks assembles decks into an ordered list of topics, where an entry is either a topic taken straight from a deck or a supertopic gathering slides from several decks under one section heading. A talk is the reusable content, and a presentation in the presentations collection is one occasion on which it was given, linked by the talk_id field that presentations already carries.

The field types follow the data in an existing talks database of 53 talks and 192 decks holding 883 slides. Several fields there are written as either a string or a list, and the figure widths as either a float or an integer, so those use anyof_type. The slide type uses eallowed rather than allowed so that an unrecognized type warns rather than failing validation.

The new entries are spliced into the two json files rather than written by a round trip, so that no existing collection is reformatted.

### DIFF
--- a/src/regolith/exemplars.json
+++ b/src/regolith/exemplars.json
@@ -2177,6 +2177,37 @@
             "year": 2020
         }
     ],
+    "slides": {
+        "_id": "example-introduction",
+        "name": "introduction",
+        "tags": ["introduction", "overview"],
+        "slides": [
+            {
+                "type": "text",
+                "title": "What is local structure?",
+                "intro": "Three ideas underpin the rest of the talk.",
+                "bullets": ["crystals are described by an average structure", "real materials deviate from that average", "those deviations control the properties"],
+                "outro": "We will return to each of these.",
+                "shrink": false
+            },
+            {
+                "type": "imageleft-2col",
+                "title": "The nanostructure problem",
+                "image": "example_nanoparticle",
+                "caption": "A nanoparticle with a disordered surface.",
+                "fig_width": 0.9,
+                "fig_rotation": 0,
+                "imagecol_width": 0.5,
+                "bullets": ["the surface is not the same as the core"],
+                "shrink": false,
+                "paper": {
+                    "doi": "10.1107/S002188981000988X",
+                    "avatars": ["aexample"],
+                    "render": true
+                }
+            }
+        ]
+    },
     "students": {
         "_id": "Human A. Person",
         "aka": ["H. A. Person"],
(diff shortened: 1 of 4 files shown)


src/regolith/exemplars.json
```diff
@@ -2272,6 +2303,37 @@
             ]
         }
     ],
+    "talks": {
+        "_id": "2406ab_example_talk",
```
sbillinge: let's use the pattern `30-example-talk` which is my new style for this (dashes instead of underscores and prepend a number that is the length of the talk.

src/regolith/exemplars.json
```diff
@@ -2272,6 +2303,37 @@
...
+        "active": true,
+        "talk_description": "An overview of the group's work on nanostructure, for a general audience.",
+        "learning_objectives": ["what a pair distribution function is", "why local structure matters for nanomaterials"],
+        "authorlist": ["A. B. Example", "C. D. Sample"],
+        "addresses": ["Department of Physics, Example University, New York, NY 10027, USA", "Department of Chemistry, Sample University, Boston, MA 02115, USA"],
```
sbillinge: shall we mix in `id`'s here that will be dereferenced to `institutions.txt`

src/regolith/exemplars.json
```diff
@@ -2272,6 +2303,37 @@
             ]
         }
     ],
+    "talks": {
+        "_id": "2406ab_example_talk",
+        "active": true,
+        "talk_description": "An overview of the group's work on nanostructure, for a general audience.",
+        "learning_objectives": ["what a pair distribution function is", "why local structure matters for nanomaterials"],
+        "authorlist": ["A. B. Example", "C. D. Sample"],
```
sbillinge: shall we mix in people `id`'s here that will be dereferenced to people/contacts collections

src/regolith/exemplars.json
```diff
@@ -2272,6 +2303,37 @@
...
+        "addresses": ["Department of Physics, Example University, New York, NY 10027, USA", "Department of Chemistry, Sample University, Boston, MA 02115, USA"],
+        "presenter": "aexample",
```
sbillinge: presenter has to be one of the authors in general

sbillinge (commented):
please see inline comments

PR #1286, opened 2026-07-31
feat: add a presentation builder that renders talks as beamer documents

sbillinge:
Add PresentationBuilder in builders/presentationbuilder.py, register it in the BUILDERS map as presentation, and add the templates it renders. It replaces the standalone script that was kept alongside the talks database.

For every presentation that names a talk with talk_id, the builder gathers the slides the talk asks for and writes one beamer tex file to the build directory. A section of a talk is either a topic taken straight from a deck, named after the deck, or a supertopic gathering slides from several decks under one heading. A slide_list is read as a slice, such as "[:]" or "[2:5]", or as a list of indices, such as "[1,4,7]". The presenter is looked up in people and falls back to contacts, so the author line holds their canonical name.

Three things are done differently from the script it replaces. The topics are gathered into new dicts rather than by updating the deck in place, which the script did and which left the slides collection modified. A slide_list that cannot be applied to its deck, and a topic or talk_id that names something missing, now raise errors that say what to do rather than IndexError or KeyError. The collections are listed rather than left as the generator that all_docs_from_collection returns, because each is scanned once per topic and a generator yields nothing the second time.

The templates are the generic form of the ones kept with the talks database, with the group specific preamble, colours and graphics path left out, since a templates directory beside regolithrc.json already takes precedence over the packaged one. The cork templates need question and answers on a slide, which the slides schema did not yet have, so those two fields are added.

The output directory is rc.builddir, which regolithrc.json can already override, so no path is written into the source.

The exemplars now select the first two slides of the example-methods deck rather than the last two, so that the builder test covers the image template as well as the text, imageleft-2col and imageright-2col ones.

### DIFF
--- a/src/regolith/builders/presentationbuilder.py
+++ b/src/regolith/builders/presentationbuilder.py
@@ -1,0 +1,181 @@
+"""Builder for presentations.
+
+A talk in the talks collection is assembled from decks of slides in the
+slides collection.  A presentation in the presentations collection is
+one occasion on which a talk was given, and points at its talk with
+talk_id. This builder renders a beamer document for every presentation
+that names a talk.
+"""
+
+import ast
+
+from regolith.builders.basebuilder import LatexBuilderBase
+from regolith.tools import all_docs_from_collection, get_person_contact
+
+
+def string_to_slice(slice_str):
+    """Return the slice described by a string such as "[2:5]".
+
+    Parameters
+    ----------
+    slice_str: str
+        The slice written in python syntax, with or without the enclosing
+        brackets.  An omitted start, stop or step is taken as None
+
+    Returns
+    -------
+    selection: slice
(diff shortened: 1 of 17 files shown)


src/regolith/builders/presentationbuilder.py
```diff
@@ -0,0 +1,181 @@
...
+that names a talk.
+"""
+
+import ast
+
+from regolith.builders.basebuilder import LatexBuilderBase
+from regolith.tools import all_docs_from_collection, get_person_contact
+
+
+def string_to_slice(slice_str):
```
sbillinge: please check if this (or something with the same functionality) is in tools.py or elsewhere

sbillinge (commented):
just one inline comment

PR #1310, opened 2026-09-06
feat: add get and find to the database clients

sbillinge:
client_manager.py, fsclient.py, mongoclient.py: builders and helpers can only reach data through all_docs_from_collection, which returns every document of a collection.  There is no way to ask a backend for one document or for a filtered subset, so MongoClient cannot use a server side query and FileSystemClient cannot avoid reading the whole collection.

Add get(collname, _id) and find(collname, filter) to the clients. MongoClient answers get from the server's index on _id and sends the filter to the server; FileSystemClient looks the id up in its dict; and find_one takes the same fast path when the filter is only an _id.

ClientManager fans out over rc.databases, asks each backing client for the documents it holds, and merges the versions with ChainDB in rc.databases order, so get and find resolve a shared document exactly as open_dbs does.  find reads each database once rather than looking each id up again, because a document is matched on its merged value, which no single database knows.  Both copy what they return by default: the documents are the clients' own, and mutating one would change state the client does not know it has to write.

This is the seam the pushdown work needs; it does not speed anything up on its own, since open_dbs still loads every needed collection eagerly.  The source map and lazy loading that use it come next.

### DIFF
--- a/news/client-get-find.rst
+++ b/news/client-get-find.rst
@@ -1,0 +1,28 @@
+**Added:**
+
+* ``get(collname, _id)`` and ``find(collname, filter)`` on the database clients.
+  ``ClientManager`` asks each database for the documents it holds and chains only
+  those, so a mongo backend answers an id lookup from its index and a filtered
+  query on the server instead of returning the whole collection.  The filesystem
+  client gains the same two methods, and ``find_one`` now looks an id up rather
+  than scanning.
+
+**Changed:**
+
+* <news item>
+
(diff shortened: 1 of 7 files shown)


sbillinge (commented):
this looks great.  Is it possible to have the tests more in the group standard using pytest.mark.paramatrize and indicating the cases?  Where this is possible, please could you do it, and where it is not possible, keep it as it is now.  Thanks.

PR #1323, opened 2026-09-06
perf: read a collection with the fast yaml loader unless it will be w…

sbillinge:
…ritten

fsclient.py: every collection was read with ruamel's round trip loader, which records the comments and formatting of the file so that writing it back keeps them.  Almost no command writes, and the round trip loader is around nine times slower than the safe one: profiling l_todo showed ruamel parsing to be all but a fraction of what the command did once starting up had been dealt with.

A collection is now read with the safe loader, and read again with the round trip loader by insert_one, insert_many, update_one and delete_one, which are the only places a collection can become dirty and so the only places a dump can follow.  Nothing that reads pays for fidelity it does not use, and nothing that writes loses it.

_loaded records how each collection was read rather than only that it was, so a collection read fast and then written is read again, while one already read round trip is not.

Reading a 100 kB collection goes from about 0.18 s to about 0.02 s, and regolith helper l_todo on a 5.7 MB database from about 1.40 s to 0.52 s.

### DIFF
--- a/news/fast-yaml-reads.rst
+++ b/news/fast-yaml-reads.rst
@@ -1,0 +1,29 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* A collection is read with ruamel's safe loader, which is around nine times
+  faster than the round trip loader regolith used for everything.  A collection
+  that is about to be written is read again with the round trip loader, so the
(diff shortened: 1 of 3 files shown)


sbillinge:
closing as not already fixed.

PR #1324, opened 2026-09-06
fix: encode dates in the value update_field writes

sbillinge:
mongoclient.py: update_field sent its value to the server as it was given, so a task carrying a datetime.date failed with bson.errors.InvalidDocument.  Mongo cannot encode a date, and regolith stores dates as iso strings, which is what bson_cleanup produces and what update_one has always applied before writing.

update_field now cleans the value the same way.  Only the value: bson_cleanup also replaces the periods in keys, so cleaning the whole $set would rewrite a path like "todos.3" to "todos-3" and set a field of that name rather than the fourth entry of the list.

Reached f_todo, u_todo and a_todo, all of which write a task through update_field, and only against the mongo backend; the filesystem backend stores dates as they are.

tests/test_mongoclient.py: the new cases cover a task carrying a date, a bare date, dates nested in a list, and a value with none, and check the path is not rewritten.  Verified to fail against the old update_field.

### DIFF
--- a/news/fix-update-field-date-encoding.rst
+++ b/news/fix-update-field-date-encoding.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* <news item>
+* No news: fixes a bug introduced in #1320 and never released.
 
 **Changed:**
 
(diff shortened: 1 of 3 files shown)


sbillinge (commented):
I think this should be a no-news. we are just fixing a bug we created.

PR #1377, opened 2026-09-16
fix: give a held goal the period we are in, not the last heading's

sbillinge:
A goal on the wishlist or on deck took its period from whatever ## Goals heading the reader had passed, so one written above the goals had none, failed validation, and stopped the whole document being stored.  parse_document now takes the period to stamp held lines with, and mcsynchelper passes the one today falls in by the group's periods.


news/mc-held-goal-takes-todays-period.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* A goal on the wishlist or on deck is given the period we are in when the document is read, rather than whatever the last ``## Goals`` heading above it said, which was none at all for a line written above the goals and stopped the document being stored.
```
sbillinge: this should probably be a no-news as it is pretty granular

sbillinge (commented):
please see inline

PR #1378, opened 2026-09-16
fix: order periods in l-mcgoals by when they came round, not by their…

sbillinge:
… text

l_mcgoalshelper.latest() picked the default period with max() over the period strings, so with the default semesters '2026summer' beat '2026fall' and the lister showed summer's goals all fall.  Sort with mc.period_key, as the builder has since #1361, reading the group's own periods from mission_control_periods.


news/mc-goals-lister-latest-period.rst
```diff
@@ -0,0 +1,23 @@
...
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
```
sbillinge, line 18: no-news this also please.

